PR #22, opened 2014-06-04
fix: commonjs wrapper

### LINKED ISSUE #19: Does not seem to be browserify compatible
Even if I try to:
```javascript
window.zone = require('zone.js');
```
I get ```zone is not defined``` because the function Zone.patchSetClearFn assumes window.zone exists (i.e. in line 133).

Perhaps Zone.init should just installs on window.zone independently of module.exports, but export the created instance as well. 

Any thoughts?

btford:
Closes #19

### DIFF
--- a/zone.js
+++ b/zone.js
@@ -1,5 +1,7 @@
 'use strict';
 
+(function (window) {
+
 
 function Zone(parentZone, data) {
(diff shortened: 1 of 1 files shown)


kentcdodds:
Would love for this to get merged! Trying to setup using Angular 2 with webpack. Having this fixed would simplify life significantly :-)

PR #27, opened 2014-07-12
chore: run tests on sauce labs

btford:
**WIP** need to fix some tests, and also handle reverse-proxying HTTP requests back to karma maybe :/


btford:
This has long since landed.

PR #31, opened 2014-09-07
Fixing crash on attachedCallback of an Polymer element

coolxeo:
When trying to include polymer paper-button in [Angular 2.0 Projects](https://github.com/angular/projects) it crash when try to register `attachedCallback` callback, `descriptor` is undefined


vicb:
adressed in https://github.com/angular/zone.js/commit/8bce00ee44cd869f3f61d6d7a9cdf5928cfdd1bc

PR #36, opened 2014-12-09
Fix removeEventListener patch

mnmtanish:
Fix a case where the event handler is not removed when the event handler
function is used to handle multiple events on the same element.


vicb:
@btford this lgtm, what do you think ?

btford:
yep, LGTM

vicb:
Maybe one small improvement would be to clear the obj entry when a listener gets removed

vicb:
merged via #96

PR #37, opened 2014-12-13
fix: fork the zone in first example (README)

### LINKED ISSUE #13: confused about the first example
@btford    
I run the first example in the chrome console,but the result is :   

     in the zone: true 
     undefined
     in the zone: true   

I'm just confused about it.

gilbox:
fix #13


vicb:
merged 7b6e8ed7af54adac896accc015aeeed58335ba42
thanks for the PR

PR #43, opened 2015-02-19
fix(patchViaPropertyDescriptor): disable if properties are not configurable

### LINKED ISSUE #42: Do not use property descriptor patching if EventHandler properties are not configurable
This is broken in Safari 8 and WebKit nightly, shouldn't hurt more spec-compliant browsers.

caitp:
Relevant WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=134364

closes #42

### DIFF
--- a/zone.js
+++ b/zone.js
@@ -375,6 +375,14 @@ Zone.patch = function patch () {
 
 //
 Zone.canPatchViaPropertyDescriptor = function () {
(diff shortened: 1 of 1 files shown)


caitp:
Fixes just one of the many, many problems running the angular2 example apps in safari

PR #61, opened 2015-03-25
gitignore node_modules

vicb:
The best practice is probably to have it globally ignored but this doesn't hurt either

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 blink-core/
 blink-idl/
 parsed-idl/
+/node_modules


btford:
I prefer not to add too many rules to a project's `.gitignore`. When I get a PR with node_modules checked in, I typically point the developer to a resource on setting up a global config.

Is there any reason to prefer adding the rule here?

pkozlowski-opensource:
IMO node_modules is project-specifc and not dev-env specific (ex. files produced by a particular editor). As such I would expect it to be part of project's .gitignore

(Sorry that my contribution to the zone-stuff amounts to this, will try to better next time :-))

btford:
@pkozlowski-opensource ha, no worries.

I don't have a strong opinion about this, so I'll just merge it.

btford:
Landed as 973657221881b1acd33d4fe3a0c89c8c70189490.

vicb:
I think the same as Pawel here.

PR #62, opened 2015-03-25
refactor: factorize some code

vicb:


zone.js
```diff
@@ -114,38 +114,25 @@ Zone.patchSetClearFn = function (obj, fnNames) {
...
 
     if (delegate) {
+      var clearName = 'clear' + name;
       var ids = {};
 
-      if (setName === 'setInterval') {
-        zone[setName] = function (fn) {
-          var id;
-          arguments[0] = function () {
-            delete ids[id];
```
vicb: Should we really `delete ids[id];` here ?

By deleting, `clearInterval()` will not call the `dequeueTask()` hook if called after the callback as triggered at least once, not sure if it makes sense ?

btford: Not sure. Let's write a test to cover this case.

vicb: ... and the expected behavior should be `dequeueTask()` always called (whenever clear is called), right ?

btford: yes

btford:
Landed as a3a11e90928c01336ca8035225cc52469e4b3f79.

PR #63, opened 2015-03-25
refactor: changed a named function to be anonymous

vicb:


btford:
Why?

vicb:
Why not ?

Actually consistency. Maybe I'm too much obsess about consistency but I find harder to understand a piece of code when multiple different conventions are used. 

If you consider merging this please wait there are other minor cleanups that I could squash here, if not feel free to close.

pkozlowski-opensource:
So the only one thing I've got "against" anonymous functions is that debugging gets harder. So if we are after consistency I would rather name all the functions.

(once again, sorry for just chiming in with style issues, still trying to wrap my head around the zone thing)

btford:
This fn appears in stacktraces sometimes, so having a name makes it a bit easier when looking at stacktraces.

vicb:
I haven't thought about it makes sense, closing this PR. Thanks for your help.

@pkozlowski-opensource I appreciate your comments, no problem. May be we can discuss about zones some day, I'm working quite a lot around it these days.

PR #65, opened 2015-03-25
test(CI): make Travis use Sauce for testing

vicb:
Tests are executed via Sauce... and failing which at least is consistent with my local setup !


btford:
Landed as 39638cf82d15ef993c249bd3ea43940b593aff93.

PR #66, opened 2015-03-26
test: cleanup

vicb:
Two minor things:
- define `testFn.message` early, otherwise it is undefined at the moment it is checked and then useless,
- use `ifEnvSupports` for Promise tests


btford:
Landed as b7e5b3c4b94a08d5945cc0a7589fa1e15b8613ef.

PR #67, opened 2015-03-26
fix: only fork on zone.fork()

### LINKED ISSUE #38: Do not create a new zone for every callback
Currently, zone.js creates a new zone every time a callback gets called:

```js
z1 = window.zone;

setTimeout(() => {
  z2 = window.zone;
  
  setTimeout(() => {
    z3 = window.zone;
  }, 0);

}, 0);

expect(z1).not.toBe(z2);
expect(z1).not.toBe(z3);
expect(z2).not.toBe(z3);
```

Maybe we should create a new zone only when `fork` is called?

vicb:
fixes #38

previously a zone was created for any callback executed in a zone

@btford could you please check the changes - I am still learning and not 100% confident on the test updates.

Also please not that I have only updating passing tests - WS tests still fail in Chrome/Firefox/Travis and are left untouched by this PR.

### DIFF
--- a/test/long-stack-trace-zone.spec.js
+++ b/test/long-stack-trace-zone.spec.js
@@ -15,8 +15,8 @@ describe('longStackTraceZone', function () {
 
   it('should produce long stack traces', function (done) {
     lstz.run(function () {
-      setTimeout(function () {
-        setTimeout(function () {
+      zone.fork().run(function () {
+        zone.fork().run(function () {
           setTimeout(function () {
             expect(log[0]).toBe('Error: hello');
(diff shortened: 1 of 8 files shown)


vicb:
@btford could we discuss what's left to be done for this PR to be mergeable ?

vicb:
I need to re-work this one.

Forking on each run is useful for keeping stack traces. It should be possible to update the lst zone to update the stack traces on `run` and don't have to fork (need to manually maintain a stack of traces).

btford:
We ca revisit once the code is more modular.

Issue #57, opened 2015-03-12
setInterval RangeError: Maximum call stack


johnlindquist:
Using the preview here: https://github.com/angular/quickstart

Adding `setInterval` anywhere throws an error

```
es6-shim.js:29889 Uncaught RangeError: Maximum call stack size exceeded
es6-shim.js:29889 run
es6-shim.js:29866 zoneBoundFn
es6-shim.js:29940 fnNames.map.forEach.zone.(anonymous function).arguments.(anonymous function)
es6-shim.js:29873 (anonymous function)
es6-shim.js:29889 runes6-shim.js:29866 zoneBoundFn
es6-shim.js:29940 fnNames.map.forEach.zone.(anonymous function).arguments.(anonymous function)
es6-shim.js:29873 (anonymous function)
es6-shim.js:29889 run
es6-shim.js:29866 zoneBoundFn
```

Invis1ble:
The same issue:
```
RangeError: Maximum call stack size exceeded
STACKTRACE:
RangeError: Maximum call stack size exceeded
    at join (native)
    at VmTurnZone.$__export._onError (http://angular/quickstart/angular2/src/core/zone/vm_turn_zone.js:84:44)
    at Zone.errorHandling.onError (http://angular/quickstart/angular2/src/core/zone/vm_turn_zone.js:58:30)
    at Zone.run (http://angular/quickstart/dist/es6-shim.js:29892:14)
    at zoneBoundFn (http://angular/quickstart/dist/es6-shim.js:29866:19)
    at arguments.(anonymous function) (http://angular/quickstart/dist/es6-shim.js:29928:23)
    at Zone.run (http://angular/quickstart/dist/es6-shim.js:29889:19)
    at zoneBoundFn (http://angular/quickstart/dist/es6-shim.js:29866:19)
    at arguments.(anonymous function) (http://angular/quickstart/dist/es6-shim.js:29928:23)
    at Zone.run (http://angular/quickstart/dist/es6-shim.js:29889:19) ```

pkozlowski-opensource:
Same with `setTimeout` - interestingly enough, this only happens when zones are included _before_ ng2 - when you push them after ng2 things seem to work as expected....

pkozlowski-opensource:
Actually, after some more testing, it doesn't seem to be linked to scripts inclusion order.

rkirov:
[...]

PR #70, opened 2015-04-07
fix(zone): keep argument[0] refs around.

rkirov:
Without this fix, zone.js creates infinite loops when running in
non-strict mode.

Closes: #57

### DIFF
--- a/zone.js
+++ b/zone.js
@@ -123,10 +123,10 @@ Zone.patchSetClearFn = function (obj, fnNames) {
       var bindArgs = setName === 'setInterval' ? Zone.bindArguments : Zone.bindArgumentsOnce;
 
       zone[setName] = function (fn) {
-        var id;
+        var id, fnRef = fn;
         arguments[0] = function () {
           delete ids[id];
-          return fn.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


rkirov:
@btford this issue is pretty tricky to test. Do you think separate non-strict run of all tests would be useful? Also looks like there are broken tests at head.

zone.js
```diff
@@ -123,10 +123,10 @@ Zone.patchSetClearFn = function (obj, fnNames) {
       var bindArgs = setName === 'setInterval' ? Zone.bindArguments : Zone.bindArgumentsOnce;
 
       zone[setName] = function (fn) {
-        var id;
+        var id, fnRef = fn;
```
vicb: @rkirov it would be good to add a comment

btford:
> Do you think separate non-strict run of all tests would be useful? 

Yes, I think so.

btford:
I'm going to merge this and cut a release so that others are unblocked, but we should run tests in non-strict as well.

btford:
Landed as 48573ffdd8726f498758df3c2b8d0b03dacef6f5.

PR #71, opened 2015-04-07
[DO NOT MERGE] feat(zone): ability to override $id + full path

### LINKED ISSUE #45: Create unique ID for each zone.
One of the reasons why debugging zones is hard is that it is hard to quickly tell if you have the same zone. I think it would be useful to append a unique `id` to each zone. Looking at the `id` would than give you the path.

`123>4324>angular>root`

It would probably mean that we could have a global counter. and the either expect the caller to supply the zone name in the fork or use the global counter to create o new `id`

vicb:
fixes #45 

@btford this is blocked by #67

### DIFF
--- a/test/zone.spec.js
+++ b/test/zone.spec.js
@@ -10,6 +10,24 @@ describe('Zone', function () {
     expect(zone.$id).toBeDefined();
   });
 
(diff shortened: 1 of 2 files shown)


vicb:
superseded by #120

PR #72, opened 2015-04-07
fix tests

vicb:
Status: WS server is not working.

I think the first commit should be ok (fine locally)


btford:
Needs work to run on travis, but otherwise LGTM

vicb:
fixed tests in d9075909a8ff01c3829c814bd8d28630c4ab7201
we still need to figure out a way to run WS tests on Travis

PR #73, opened 2015-04-08
(updated) fix: commonjs wrapper

### LINKED ISSUE #19: Does not seem to be browserify compatible

unframework:
This is a rebase and tweak of original commonjs wrapper fix: https://github.com/angular/zone.js/pull/22

Closes #19

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -14,7 +14,9 @@ module.exports = function (config) {
 
     reporters: ['progress'],
 
-    preprocessors: {},
+    preprocessors: {
+      'test/commonjs.spec.js': [ 'browserify' ]
+    },
 
     //port: 9876,
     colors: true,
@@ -22,7 +24,7 @@ module.exports = function (config) {
     logLevel: config.LOG_INFO,
 
     browsers: ['Firefox'],
-    frameworks: ['jasmine'],
+    frameworks: ['jasmine', 'browserify'],
 
     captureTimeout: 60000,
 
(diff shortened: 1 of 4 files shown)


btford:
It'd be nice to have a test showing that this works.

unframework:
@btford thanks for the response! I just added a test using karma-browserify; also had to tweak the original implementation because it inadvertently masked original window reference for property patching.

unframework:
... not sure if the test failures were there from before - the code seems to work OK in my local browserify setup in Chrome and FF.

unframework:
I created a temp npm package to be able to reference this as a dependency. Will remove once merged.

btford:
@unframework – FYI you can `npm install unframework/zone.js` and npm will grab your stuff from github. It's better not to pollute the npm namespace, even if only temporarily.

unframework:
Hey thanks for the tip, I forgot about that `package.json` feature. Removed the temp package, using the Git dependency now

btford:
Landed as 5d6bee4b47d82a8cf1cde76ce19b70350a67ff10.

I'll cut a release with this sometime soon.

ocombe:
Ahh nice, waiting for this release as well :)

ocombe:
By the way, you'll have to edit the following file https://github.com/angular/angular/blob/master/modules/angular2/src/core/zone/vm_turn_zone.es6 to make it work with your changes (you probably know that already), I tested it locally and it works with webpack now (I edited this file from angular2 to test)

PR #76, opened 2015-04-15
fix lst zone: honor parent's fork()

### LINKED ISSUE #75: Long stack trace zones & fork
Currently lst fork is as follow:

```
  fork: function (locals) {
    var newZone = this._fork(locals);
    newZone.constructedAtException = Zone.getStacktrace();
    newZone.constructedAtTime = Date.now();
    return newZone;
  },

  _fork: zone.fork
```

Which is wrong because:

```
    zone // in zone A
      .fork({ // B
        '+fork': function() { log.push('fork'); }
      })
      .fork(Zone.longStackTraceZone) // C
      .fork(); // D
```

When creating the zone C, _fork will be set to the current zone's fork, which is zone A. The log is ['fork'] where it should be ['fork', 'fork'] for C, D.

I can not think of an easy fix for now.

My first thought was `var newZone = this.parent.fork(locals);` the issue is that it will recursively call fork() on the parent for forked zones.

vicb:
fixes #75

### DIFF
--- a/test/long-stack-trace-zone.spec.js
+++ b/test/long-stack-trace-zone.spec.js
@@ -50,4 +50,16 @@ describe('longStackTraceZone', function () {
   it('should expose LST via getLogStackTrace', function () {
     expect(lstz.getLongStacktrace()).toBeDefined();
   });
+
+  it('should honor parent\'s fork()', function () {
+    zone
+      .fork({
+        '+fork': function() { log.push('fork'); }
+      })
+      .fork(Zone.longStackTraceZone)
+      .fork();
+
+    expect(log).toEqual(['fork', 'fork']);
+  });
+
 });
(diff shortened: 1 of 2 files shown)


test/long-stack-trace-zone.spec.js
```diff
@@ -50,4 +50,16 @@ describe('longStackTraceZone', function () {
   it('should expose LST via getLogStackTrace', function () {
     expect(lstz.getLongStacktrace()).toBeDefined();
   });
+
+  it('should honor parent\'s fork()', function () {
+    zone
+      .fork({
+        '+fork': function() { log.push('fork'); }
+      })
+      .fork(Zone.longStackTraceZone)
```
vicb: without the fix in this PR, this `fork` would erase the hook previously defined

btford:
LGTM :+1:

vicb:
merged as eef7116d6b6be2b02be2cb85f6eb24533c2d4640

PR #78, opened 2015-04-21
feat(Node): make zone.js work in Node

mlaval:
This PR makes zone.js to work in Node, by setting `zone` and `Zone` as global.
To run tests in Node, do `gulp test`

For sure, all the patches which are DOM related are not applied, and not tested.

The goal behind it is to be able to fully run an Angular2 application in Node.


btford:
@vicb will take care of merging this once we improve the modularity of zone.js

vicb:
see #116

PR #80, opened 2015-04-25
fix: use strcit mode for the zone.js code only

pkozlowski-opensource:
The problem with using strict mode for the entire files
is that it will propagate for the entire bundled files
if zones.js is concatenated with other code written in
the non-strict mode

### DIFF
--- a/zone.js
+++ b/zone.js
@@ -1,6 +1,6 @@
-'use strict';
-
 (function (exports) {
+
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
This should fix https://github.com/angular/angular/issues/1522

btford:
:+1:

PR #82, opened 2015-04-30
refactor: split zone.js into CJS modules, ad zone-microtask.js

vicb:
This PR splits `zone.js` into smaller CJS modules and add `zone-microtask.js` that will be used in Angular2 in conjunction with [es6-promise](https://github.com/jakearchibald/es6-promise/pull/113).

With the help of browserify, we generate:
- `dist/zone.js` - the equivalent as the former zone.js
- `dist/zone-microtask.js` - the new zone with microtask support

Notes: 
- the minimified version (`.min.js`) are generated at the same time.
- tests do pass locally (Chrome & Firefox),
- smoke test of zone-microtask in Angular2 looks good.

TODO (not blocking, will submit follow-up issues/PR if this gets merged):
- [x] re-integrate commonJS support #73. We should be able to come with a proper support now, also relates to #78. The CJS version should not try to patch browser specific stuffs.
- [ ] remove the `es6-promise` source and use the npm package if/when the patch gets integrated
- [x] add tests for `zone-microtask.js`
- [ ] only `zone` and `zone-microtasks` are generated by the build system. Other `.js` files live in the dist folder, we might want to modularize them as well.
- [x] ~~`setImmediate()` should enqueue a microtask in `zone-microtask.js`~~ wrong, that should be a macrotask, nothing to update

Breaking changes:
- `zone.js` and '*-zone.js` moved from '/' to '/dist'


vicb:
Aligned with master (as of c26c7acb2932b7eb1f291108eee2f75810a459ae).

Question: as seen in `lib/zone.js` (the CJS version of zones), we do apply browser patches:

```js
'use strict';

var core = require('./core');
var browserPatch = require('./patch/browser');

global.zone = new core.Zone();

module.exports = {
  Zone: core.Zone,
  zone: global.zone
};

browserPatch.apply();
``` 

This will fail when using node:
```js
require('zone.js');

// -> 
// path/to/zone.js/npm/node_modules/zone.js/lib/patch/property-descriptor.js:26
//  if (!Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'onclick') && ty
```

We could integrate @mlaval PR #78 to make this "work" in node env. However "work" here is mainly not crash. I'm not sure if we should put a lot of effort on this, we'd better come with an impl based on sometging like [strongloop/zone](https://github.com/strongloop/zone) for CJS.

Thoughts ?

/cc @IgorMinar

btford:
@vicb – I'm currently working with @IgorMinar to address some issues. We should perhaps keep this on hold until after Igor's issues/PRs are addressed.

vicb:
I'm currently working on Angular2 `VmTurnZone` which needs this (see https://github.com/angular/angular/pull/1574) - I'll sync with your changes. Do not hesitate to initiate a chat if you fell it could help

mhevery:
@btford this is a priority for Angular2, and is blocking GT.

mhevery:
I assume most of these changes are mechanical. Why do we have a `dist` folder? Are we checking in artifacts? Is that what we want to do?

lib/microtask.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
```
mhevery: Looks reasonable

vicb:
> I assume most of these changes are mechanical. 

most of them are

> Why do we have a dist folder? Are we checking in artifacts? Is that what we want to do?

We now have the sources spread over CJS modules.
Hence I had to create a build step to generate `zone.js` and `zone-microtask.js` from them using browserify. Those 2 files (as well as their minified version) is stored in the dist folder.
Note that ATM there are a few other files living in this folder even though they are not built. This is for consistency & they'll probably need to be converted to CJS at some point (not blocking for now). More details in the PR comment.

vicb:
Aligned with master as of 496b006b5f550d9628a55445915b44fdddc64a06

vicb:
@IgorMinar two pending issues before this can get merged:

- it's still not clear to me what we should do with FF, I vote for disabling it on sauce now (FF38 works locally),
- when I tried to install the package (see package.json below), I get all the files instead of /dist and /lib only

feedback welcome

```
{
"name": "test",
  "dependencies": {
    "zone.js": "vicb/zone.js#0430-modularize"
  }
}
```

IgorMinar:
please make sure you grab d6685a8 and 4edf0d5 from #94

vicb:
@IgorMinar @btford After Igor has found a way to run FF37 + the fix "	test(XmlHttpRequest): fix expect() being called after done()" the build was green.

Then I figured out that some `expect`ations were basically no-op, this should be finxed in "	fix(test): fix an ineffective assertion". The following 2 commits also improve the tests by catching more errors.

However the build is red again now :(

There are only 2 failing on Chrome 42, see [the Travis log](https://travis-ci.org/angular/zone.js/builds/61765398#L292).

The problem is with `HTMLLinkElement` but dissapears with Chrome 44.

The difference is that:
- Chrome 42 is not able to `canPatchViaPropertyDescriptor`
- Chrome 44 `canPatchViaPropertyDescriptor`

Is there something we can do about it, or could it be merged like this (& we want for the Chrome to update ?).

@IgorMinar or @btford can on of you review the commits from "fix(test): fix an ineffective assertion" onwards ? Thanks !

IgorMinar:
If the problem is only in old Chrome then that's fine. We don't care.

vicb:
I think that 42 is the current stable, not yet "old".
I would also appreciate a review of the last commits, I have changed a few behaviors in the last "fix" commits.

Issue #41, opened 2015-02-04
chore(travis): spin up test WebSocket server


btford:
right now, the tests likely fail because of it.

IgorMinar:
As @vicb also found, according http://sauceio.com/index.php/2014/11/sauce-connect-gets-a-speed-boost-websocket-support/ sauce should not have websocket support, but in my manual testing it's not the case. At least with sauce connect v4.3.8.

IgorMinar:
for the record the console error reported by the browser is:

```
WebSocket connection to 'ws://localhost:8001' failed: Unexpected response code: 400
```

IgorMinar:
a solution might be to use browserstack instead.

vicb:
I've contacted SL support for that, seems like it is possible.
Will work with them on getting this running.

PR #84, opened 2015-04-30
test: exclude WS tests for Sauce

vicb:


vicb:
It would be great to finally see Travis green.

It is still not after this PR because of the CJS tests but at least this PR do what's advertised.

btford:
According to their documentation, Sauce Connect should support websockets. Is this no longer the case?

vicb:
Nope it is not that simple - tried that in a closed PR.
I'll try to contact their support next week but before we get an answer, this patch is meaningful. No need to run tests we know are red. It gives the bad habit of stop caring about Travis results.

IgorMinar:
I thought that karma used websockets on SL, but it doesn't. I agree that we should disable the ws test on sauce for now. Maybe we could add write at least one test that would fake ws communication or at least check if the ws methods are being patched.

IgorMinar:
more on this in #41

IgorMinar:
I'm going to merge this in to get the build into a better state.

PR #87, opened 2015-05-01
feat: make root zone id to be 1

IgorMinar:
it feels more natural this way


btford:
:+1:

vicb:
merged into #82

vicb:
merged via #82

PR #94, opened 2015-05-02
don't spawn child zones from the root zone automatically for callabacks + don't leak zones in async jasmine tests

IgorMinar:


IgorMinar:
@mhevery can you please review this as well. especially these changes:

- fix: don't fork new zones for callbacks from the root zone
- fix: always run jasmine's done callbacks for async tests in jasmine's zone

The latter one should be something that anyone that writes jasmine tests for code using zones should be using. That means angular/angular repo and any angular app. This makes me quite sad.

IgorMinar:
for some reason saucelabs fails to provide us with FF37 at the moment. hopefully this is just a temp issue. feel free to rerun the build when you are reviewing this. it should pass all tests.

IgorMinar:
@btford I'm not sure if `test: Element#onclick should run only if env supports it` change is correct. it might be a legit bug in which case we should file an new issue and merge this as is.

vicb:
Isn't the 2nd commit (do not fork the root zone) a BC break ?
If yes we should probably:
- merge all other fixes before & cut a last 0.4 release,
- clearly state that in the comment,
- create a 0.5.0 release with the BC breaks (there is also id numbering pending)

vicb:
Great PR btw !

btford:
RE: https://github.com/angular/zone.js/pull/94#issuecomment-98372472

+1, we should try to cut a v0.4.x release, then get all this in, then release v0.5.

test/jasmine-patch.js
```diff
@@ -0,0 +1,40 @@
...
+      originalFit(description, specFn);  
+    }
+  };
+
+}(window));
+
+// global beforeEach to check if we always start from the root zone
+beforeEach(function() {
+  expect(window.zone.isRootZone()).toBe(true);
+});
```
btford: Missing EOL

test/jasmine-patch.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+(function() {
+  // patch jasmine's it and fit functions so that the `done` callback always resets the zone
+  // to the jasmine zone, which should be the root zone.
```
btford: Link to GH issue/discussion

btford:
LGTM other than the few cosmetic things.

mhevery:
looks reasonable to me.

vicb:
@btford @IgorMinar What about you _(I can not)_ make a release out of master as of now and I integrate all those changes in the split PR (#82) and merge it so that I we can get 0.5 out (would be great if this can happen tomorrow) ?

@IgorMinar If you have no news about what's going on with FF, let's stick to the currently used version for now ? Have you asked sauce labs support ?

IgorMinar:
I haven't had time to work on this. We should definitely upgrade FF since it's unlikely that anyone we care about is using the version we are testing against. Can you try to fiddle with it, maybe use  slightly older version of FF?

I think the rest of the commits are good to go. I can merge those in tomorrow morning.

vicb:
> Can you try to fiddle with it, maybe use slightly older version of FF?

I could

> I think the rest of the commits are good to go. I can merge those in tomorrow morning.

Not sure if it is valuable to merge them on top of master. What I would love to see is a release on the **current master** and then I'll merge my CJS PR and integrate your commits into there and then we cut a 0.5.

I've tried to reach you on hangout w/o success to discuss about this

vicb:
@IgorMinar context: the CJS PR is required for angular2 VmTurnZone which I would love to be able to merge shortly.

test/jasmine-patch.js
IgorMinar: there is a eol

vicb:
merged into #82 last commit excluded, should I cheery pick this one also ?

IgorMinar:
@vicb you do want the sauce connect update, please cp that as well

IgorMinar:
OMG. GREEN BUILD. THAT CAN'T BE TRUE!

@vicb please cp also the ff upgrade and you should be golden

vicb:
GREAT !!!!

FF37 launches ok right now... but I'm still seeing the "expect" issue now :(

zone.js
```diff
@@ -8,7 +8,7 @@ var zone = null;
 function Zone(parentZone, data) {
   var zone = (arguments.length) ? Object.create(parentZone) : this;
 
-  zone.parent = parentZone;
+  zone.parent = parentZone || null;
```
vicb: Doesn't this invalidate a lot of tests ?
ie `expect(window.zone.parent).toBeDefined();` to test if the zone is the root zone ?
I'll try to update to `isRootZone`

vicb: Hum... it's a little more subtle: the root zone will not be forked on bind

vicb:
Playing with `fdescribe`, I was abble to narrow down the expect issue to `XMLHttpRequest`

_As always the error was in the last file in the folder... took me 15+ iterations to found it :(_

vicb:
And the reason is: `onreadystatechange` is called multiple times !

vicb:
OMG... DEEEP GREEEN !!!

zone.js
IgorMinar: right. that's what this PR is doing and that's an intentional change :) see a few lines below.

IgorMinar:
`req.onreadystatechange` is called multiple times per the spec. so our test is not correct.

zone.js
vicb: > right. that's what this PR is doing and that's an intentional change :) see a few lines below.

And I think this is the right thing to do. The wrong part is not to have updated the tests ;)

vicb:
> req.onreadystatechange is called multiple times per the spec. so our test is not correct.

fixed in my PR

vicb:
merged via #82

PR #96, opened 2015-05-03
Fix removeEventListener patch

vicb:
Fix a case where the event handler is not removed when the event handler
function is used to handle multiple events on the same element.

replaces #36:
- add an assertion in tests,
- + `delete _bound[eventName];`

/cc @btford @IgorMinar @mnmtanish

### DIFF
--- a/test/patch/element.spec.js
+++ b/test/patch/element.spec.js
@@ -26,17 +26,20 @@ describe('element', function () {
 
   it('should respect removeEventListener', function () {
     var log = '';
-    var logOnClick = function logOnClick () {
+    var logFunction = function logFunction () {
       log += 'a';
     };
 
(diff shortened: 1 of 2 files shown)


test/patch/element.spec.js
```diff
@@ -26,17 +26,20 @@ describe('element', function () {
...
-    var logOnClick = function logOnClick () {
+    var logFunction = function logFunction () {
       log += 'a';
     };
 
-    button.addEventListener('click', logOnClick);
+    button.addEventListener('click', logFunction);
+    button.addEventListener('focus', logFunction);
     button.click();
     expect(log).toEqual('a');
```
vicb: remove in the original PR, added back

btford:
LGTM.

Issue #98, opened 2015-05-08
Do not commit dist/zone*.js with each update


vicb:
As of today, we need to `gulp build` the sources whenever a file in `/lib` is updated.

This causes a few problem:
- size,
- easy to get out of sync.

What we should do:
- use the `/lib` files for the tests,
- only update the `/dist` when the package is published

btford:
> only update the /dist when the package is published

:+1:

vicb:
> only update the /dist when the package is published

I have tried to follow this policy for the last PRs. One exception is when files gets added to the folder, it then makes sense to update it, do you think it makes sense ?

PR #101, opened 2015-05-09
feat(test): serve lib/ files instead of dist/

vicb:
### first commit:

relates to #98 

Previously tests were using the generated JS files in `/dist` which means that we have to rebuild after any modification of source files (in `/lib`).

With this PR tests use the files in `/lib` and are automatically re-run when one of them is changed. This is a step towards not having to `gulp build` for each single commit but only for releases.

### next commits 

test: strengthen assertions & some cleanup.


IgorMinar:
lgtm

PR #105, opened 2015-05-13
feat(jasmine): export the jasmine patch

vicb:


vicb:
@IgorMinar could you please review ?

vicb:
One added benefit is that the browserify generated script will also work in CJS context (which is required for angular2 tests)

lib/browser/jasmine-patch.js
```diff
@@ -0,0 +1,3 @@
+var jasminePatch = require('../jasmine/patch');
+
+jasminePatch.apply();
```
IgorMinar: why do we need this file? it doesn't look browser specific at all.

why not call apply from within the jasmine-patch?

IgorMinar:
the rest looks good

lib/browser/jasmine-patch.js
vicb: The idea of this 'browser' folder is to generate IIFE in the `dist/` folder.

All other CJS module module in the `lib/` folder are not immediately invoked, ie you can require them and invoke them when you want in a CJS context.

Does this make sense or do you think that importing the module should also apply the patch ?

IgorMinar: it's up to you. if we want the autoinstall behavior in cjs context then I'd
combine the two. if there are usecases when you wouldn't want autoapply
then keep it as is.

vicb:
I've kept it as is.
No strong prefs here but that is consistent with other generated files.

Issue #52, opened 2015-03-01
`registerElement` should support callbacks coming from deep prototype chains


rkirov:
The following line:

https://github.com/angular/zone.js/blob/26a4dc27f781afb6e6dd1258a15cb8c0840c706d/zone.js#L663

Checks `if (opts.prototype[callback])` and then uses `Object.getOwnPropertyDescriptor(opts.prototype, callback)` to retrieve it. If the callback comes from the prototype chain we cannot use `getOwnPropertyDescriptor`.

btford:
+1

tkhemani:
Hi,
I tried to run the "youtube-app" demo, but it's failing at this line. I see that you've mentioned this: 
## need to be patched
- zone.js
	- https://github.com/angular/zone.js/issues/52 (need to patch locally in node_modules/zone.js)
- angular 2
	- https://github.com/angular/angular/issues/776 (need to patch locally)

Can you please share the patch...

kingcody:
I encountered this issue as well (when dealing with polymer elements).

@tkhemani, its not very elegant or heavily tested but here was my [solution](https://github.com/kingcody/zone.js/commit/49fd5bbc3032b2b1d4428938c09aa0a6c081f8b8).

tkhemani:
@kingcody . Thanks, I got the youtube-app working after @rkirov shared the patches. But i'm still not able to get a simple [polumer button](https://github.com/Polymer/paper-button) working inside an angular 2.0 component directive. Can you possibly try it out.

kingcody:
@tkhemani Working in FF, but not in Chrome.

tkhemani:
@kingcody, are you saying u were able to  get a [polymer button](https://github.com/Polymer/paper-button) working inside an angular 2.0 component directive? Can you share the code.

JoshAntBrown:
I'm having this issue too, again like @kingcody when dealing with polymer elements but this time the Google Web Component ones. I did see a pull request at https://github.com/angular/zone.js/pull/31 which  looks like it might have tried to fix the problem but implementing their code into my project stopped the console error but still didn't render the component. I tried the same solution as @kingcody as well but same result as the pull request.

jianliao:
@JoshAntBrown Yes, there is no such issue in FF, but it seems to me the property binding like this 
```
[hidden]="!active" 
``` 
does not work. The reason that customer element does not work in Chrome is probably due to https://github.com/angular/angular/issues/1442

[...]

PR #107, opened 2015-05-15
fix(register-element): add check for descriptor being defined

jeffbcross:
Using other component libraries, like Polymer, alongside Angular2+zones would
cause errors because the descriptor was undefined when the patched registerElement
was being called.

### DIFF
--- a/lib/patch/register-element.js
+++ b/lib/patch/register-element.js
@@ -23,6 +23,8 @@ function apply() {
           descriptor.value = global.zone.bind(descriptor.value || opts.prototype[callback]);
           _redefineProperty(opts.prototype, callback, descriptor);
         }
+      } else if (opts.prototype[callback]) {
+        opts.prototype[callback] = global.zone.bind(opts.prototype[callback]);
       }
     });
     return _registerElement.apply(document, [name, opts]);
--- a/lib/patch/register-element.js
+++ b/lib/patch/register-element.js
@@ -17,7 +17,7 @@ function apply() {
 
   document.registerElement = function (name, opts) {
     callbacks.forEach(function (callback) {
-      if (opts.prototype[callback]) {
+      if (opts.prototype.hasOwnProperty(callback)) {
         var descriptor = Object.getOwnPropertyDescriptor(opts.prototype, callback);
         if (descriptor.value) {
           descriptor.value = global.zone.bind(descriptor.value || opts.prototype[callback]);
--- a/lib/patch/register-element.js
+++ b/lib/patch/register-element.js
@@ -16,17 +16,20 @@ function apply() {
   ];
 
   document.registerElement = function (name, opts) {
-    callbacks.forEach(function (callback) {
-      if (opts.prototype.hasOwnProperty(callback)) {
-        var descriptor = Object.getOwnPropertyDescriptor(opts.prototype, callback);
-        if (descriptor.value) {
-          descriptor.value = global.zone.bind(descriptor.value || opts.prototype[callback]);
-          _redefineProperty(opts.prototype, callback, descriptor);
+    if (opts && opts.prototype) {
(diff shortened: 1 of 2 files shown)


jeffbcross:
I need to add test, and to determine the proper fix for this.

jeffbcross:
I updated the implementation to check `hasOwnProperty` instead of just checking that callback name is present on prototype. Though this will prevent an error, but it would prevent the callback from being properly patched.

@vicb would the right approach here be to just set the zone-bound callback to prototype[callback] if !hasOwnProperty?

vicb:
@jeffbcross might be... I'll take a look sometimes next week after I tackle some other issues here.

jeffbcross:
@vicb k, thanks. I added a commit (the squash commit) that handles this.

I also added a fix for when options argument isn't provided.

lib/patch/register-element.js
```diff
@@ -16,15 +16,20 @@ function apply() {
...
-        var descriptor = Object.getOwnPropertyDescriptor(opts.prototype, callback);
-        if (descriptor.value) {
-          descriptor.value = global.zone.bind(descriptor.value || opts.prototype[callback]);
-          _redefineProperty(opts.prototype, callback, descriptor);
+    if (opts && opts.prototype) {
+      callbacks.forEach(function (callback) {
+        if (opts.prototype.hasOwnProperty(callback)) {
+          var descriptor = Object.getOwnPropertyDescriptor(opts.prototype, callback);
+          if (descriptor.value) {
+            descriptor.value = global.zone.bind(descriptor.value || opts.prototype[callback]);
```
vicb: you should be able to safely remove ` || opts.prototype[callback]`

test/patch/registerElement.spec.js
```diff
@@ -129,4 +129,27 @@ describe('document.registerElement', ifEnvSupports(registerElement, function ()
     var elt = document.createElement('x-props-desc');
   });
 
+
+  it('should check bind callback if not own property', function (done) {
+    testZone.run(function() {
+      var originalProto = {
+        createdCallback: function() {}
+      }
```
vicb: missing ;

vicb: Could you add an assertion in the createdCallback:
```
assert(window.zone).toBeDirectChildOf(testZone);
done();
```

and create the custom element at the end of the test (out of the testZone)

test/patch/registerElement.spec.js
```diff
@@ -129,4 +129,27 @@ describe('document.registerElement', ifEnvSupports(registerElement, function ()
...
+      var secondaryProto = Object.create(originalProto);
+      expect(secondaryProto.createdCallback).toBe(originalProto.createdCallback);
+      expect(function() {
+        document.registerElement('x-inherited-callback', { prototype: secondaryProto });
+        expect(secondaryProto.createdCallback).not.toBe(originalProto.createdCallback);
+        done();
```
vicb: `done` should not be required as of now, the test is sync. 

However I think it would be great to actually create an element an assert the cb zone.

vicb:
@jeffbcross 
- Could you please fix the comments when they are applicable ?
- Please also add `fixes #52` in the first commit comment.

Question: as a custom element can extend and other custom element, should we add some kind of marker to the bound callbacks to avoid patching them multiple time ?

/cc @IgorMinar

lib/patch/register-element.js
vicb: or actually move it to an else branch ?

vicb:
#52 & #31 both report an issue with Polymer, should we add some integration tests ?

jeffbcross:
Comments addressed and merged at https://github.com/angular/zone.js/commit/8bce00ee44cd869f3f61d6d7a9cdf5928cfdd1bc

PR #112, opened 2015-05-15
test(fetch): add a test for fetch

vicb:
Tesing which browsers need a patch

### DIFF
--- a/test/patch/fetch.spec.js
+++ b/test/patch/fetch.spec.js
@@ -1,0 +1,38 @@
+'use strict';
+
+describe('fetch', ifEnvSupports('fetch', function () {
+  var testZone = zone.fork();
+  var oPromise = Promise;
+
+  beforeEach(function() {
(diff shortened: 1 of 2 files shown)


test/patch/fetch.spec.js
```diff
@@ -0,0 +1,38 @@
...
+      throw new Error('use window.Promise');
+    };
+  });
+
+  afterEach(function() {
+    Promise = oPromise;
+  });
+
+
+  fit('should work on success', function(done) {
```
IgorMinar: fit

vicb:
Discussed with @IgorMinar The best strategy seems to be to patch all browser APIs that return `Promise`.

_& upload the idl script to download more stuff, ie https://chromium.googlesource.com/chromium/blink/+/master/Source/modules/serviceworkers/ServiceWorkerRegistration.idl_

vicb:
see #119

PR #114, opened 2015-05-18
feat(geolocation): patch the API

### LINKED ISSUE #113: Geolocation API not patched
See: https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/Using_geolocation

via @jeffbcross

vicb:
fixes #113 

Tests pass locally however the spec is disabled because it requires to manually allow the geolocation (via a popup).

/cc @jeffbcross @btford

### DIFF
--- a/lib/patch/geolocation.js
+++ b/lib/patch/geolocation.js
@@ -1,0 +1,16 @@
+'use strict';
+
+var utils = require('../utils');
+
+function apply() {
+  if (global.navigator && global.navigator.geolocation) {
+    utils.patchPrototype(global.navigator.geolocation, [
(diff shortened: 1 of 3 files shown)


lib/patch/geolocation.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+var utils = require('../utils');
+
+function apply() {
+  if (global.navigator && global.navigator.geolocation) {
+    utils.patchPrototype(global.navigator.geolocation, [
```
vicb: missing `.prototype` ?

IgorMinar:
lgtm

PR #116, opened 2015-05-18
[WIP] feat(Node): make zone.js work in Node

vicb:
This is a port of @mlaval #78 to zone 0.5

relates to https://github.com/angular/angular/issues/1904 & the short term plan described there

/cc @tbosch @IgorMinar

Edit: 
- before merging we should discuss if the short term plan defined in the linked PR makes sense,
- I'd like to find a good solution to distribute both zone.js and zone-microtask.js (have different repos ?)

[Review on Reviewable](https://reviewable.io/reviews/angular/zone.js/116)


vicb:
@btford It is not comprehensive, see the linked PR for more info

matthewjh:
@vicb: does Zone work in node now?

vicb:
Not that I know

matthewjh:
In that case, I can take a look at how node environments can be supported. Seems like it could be interesting.

Is that OK or have you got separate plans to tackle this?

vicb:
That would be great !

boneskull:
also see [continuation-local-storage](https://www.npmjs.com/package/continuation-local-storage)

mhevery:
obsolete. Latest zone works on node.

boneskull:
What versions?

mhevery:
v0.6.4.  (But we don't patch node's API)

boneskull:
right, this is problematic.  it seems like a fool's errand to try to patch every async API method in every possible environment.  hope the proposal works out--from where I sit, the only feasible solution is something at the specification level.

PR #117, opened 2015-05-19
[WIP] refactor(jasmine patch): add support for microtasks

vicb:
This makes sure that any pending microtasks is drained before the next test executes.

Specs are explicitly executed in the `jasmineZone` then there is no need to assert that in the `beforeEach` any more.

_Note: this has been tested in Angular2 context_

@IgorMinar could you please review. We should cut a release to update Angular if you're happy with the changes.


vicb:
Discussed with @IgorMinar will try to come with a less invasive patch (use `setTimeout` only when microtasks are pending at the b/o a test)

vicb:
I'll re-open when I can find a repro case

PR #119, opened 2015-05-20
feat(fetch): patch the fetch API

### LINKED ISSUE #108: fetch api is not being intercepted by zones
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

example usage:
```
fetch('foo.json').then(function(response) { return response.json() });
```

vicb:
fixes #108

### DIFF
--- a/karma-browserify.conf.js
+++ b/karma-browserify.conf.js
@@ -6,7 +6,7 @@ module.exports = function (config) {
     files: [
       'test/util.js',
       'test/commonjs.spec.js',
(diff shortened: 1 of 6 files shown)


vicb:
@IgorMinar this seems to be working...
I haven't patched more APIs because:
- APIs returning `Promise` are most of the time experimental - we might want to wait for more browsers to support `fetch` to test it (currently only Chrome supports it as of the SL versions),
- I prefer that we agree on a first PR before jumping into this.

_Note: #118 has an updated script to fetch more .idl and that will help_

IgorMinar:
looks good! thanks

PR #120, opened 2015-05-20
feat(zone): ability to override $id + full path

### LINKED ISSUE #45: Create unique ID for each zone.

vicb:
fix #45

[Review on Reviewable](https://reviewable.io/reviews/angular/zone.js/120)

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -2,6 +2,7 @@
 
 function Zone(parentZone, data) {
   var zone = (arguments.length) ? Object.create(parentZone) : this;
+  var id;
 
   zone.parent = parentZone || null;
 
@@ -9,8 +10,10 @@ function Zone(parentZone, data) {
 
     var _property = property.substr(1);
 
-    // augment the new zone with a hook decorates the parent's hook
-    if (property[0] === '$') {
+    if (property === '$id') {
+      id = data[property];
+    } else if (property[0] === '$') {
+      // augment the new zone with a hook decorates the parent's hook
       zone[_property] = data[property](parentZone[_property] || function () {});
 
     // augment the new zone with a hook that runs after the parent's hook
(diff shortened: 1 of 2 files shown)


lib/core.js
```diff
@@ -52,6 +57,10 @@ function Zone(parentZone, data) {
 Zone.prototype = {
   constructor: Zone,
 
+  path: function () {
+    return (this.parent ? this.parent.path() : '') + '/' + this.$id;
```
IgorMinar: slash seems like a weird delimiter. how about ` > ` or ` -> ` (note the surrounding spaces added for readability)

test/zone.spec.js
```diff
@@ -11,6 +11,24 @@ describe('Zone', function () {
...
+    zone.fork({'$id': 'test'}).run(function () {
+      expect(zone.$id).toEqual('test');
+    });
+  });
+
+  it('should return the full path', function () {
+    Zone.nextId = 2;
+
+    zone.fork().run(function () {
+      zone.fork({'$id': 'foo'}).run(function () {
```
IgorMinar: why not keep the number as is. I think it's useful to have a sequential number that helps us determine when (relatively speaking) a zone was created.

also exposing $id as a public api seems like an abstraction leak.

instead, I propose that there is a property called "name" (possibly confusing because of window.name), "label", "zoneName" or "zoneLabel" on a zone, which by default is an empty string, but can be overridden just like any `setTimeout` or other hooks.

path could then return a combo of $id and this label property, so you'd get `"root#1 > #2 > angularTurnZone#3 -> #4 > #5"`

lib/core.js
vicb: I kind of like "/" as a path delimiter but let's got " > " as long at it's not "\".

test/zone.spec.js
vicb: I've pick `$id` because:
- we get the id via `$id` why not use the same mechanism to set it ? 
- IMO the property name has to be "special", because you can set any value on a zone and `name` is certainly too generic and it would be easy to shoot yourself in a foot. I think it has to have some kind of escaping/prefix

Also `$id` is already kind of special, it changes for any new zone.

IgorMinar: if you let people override $id, they could set it to 1, which could cause
all kinds of issues.

how about zone.fork([label], [options])?

so you would use it as:

zone.fork('myZone', {...})

I've also had a discussion with Misko about making custom zone installation
a different operation from forking, that would resolve this problem as well.

mhevery:
obsolete

PR #104, opened 2015-05-13
[WIP] chore(Travis): Spin up a WS server

vicb:
I'm working with the SL support to get this working.

PR #123, opened 2015-05-27
Spin up a WS Server, Bonus Safari Edition

DylanLacey:
Add the --vm-version dev-varnish option to Sauce Connect, to spin up an instance capable of doing WebSocket traffic.

Worked around problem Safari has sending localhost traffic through Sauce Connect by adding explicit hostnames.  The `WEBSOCKET_HOST` environment variable will be replaced (with Envify) with the value of `os.hostname` do it should JustWork locally without hosts file wunkery.

Added a hostname entry for `travis.dev` on TravisCI.

This assists with https://github.com/angular/zone.js/pull/104.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,6 @@ script:
   - karma start karma-browserify-sauce.conf.js
   - karma start karma-microtasks-sauce.conf.js
 
+addons:
+  hosts:
+    - travis.dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_script:
   - mkdir -p $LOGS_DIR
   - ./scripts/sauce/sauce_connect_setup.sh
   - ./scripts/sauce/sauce_connect_block.sh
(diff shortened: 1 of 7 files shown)


vicb:
@DylanLacey Thank you so much again for the great support. The remaining issue is a "real" bug caught  by youe updates. Could you please squash the commits and force push ?

@btford @IgorMinar seems like the bot is not checking the CLA, any setup to do on our side ?

.travis.yml
```diff
@@ -14,9 +14,13 @@ before_script:
...
+  - node test/ws-server.js &
 
 script:
   - karma start karma-sauce.conf.js
   - karma start karma-browserify-sauce.conf.js
   - karma start karma-microtasks-sauce.conf.js
 
+addons:
+  hosts:
+    - travis.dev
```
vicb: could you please add a comment here that this relates to WS ?

DylanLacey:
Aye, sure thing.

PR #126, opened 2015-05-27
fix loading in SystemJS (`utils.js.js`)

Mitranim:


Mitranim:
Steps to reproduce:
```sh
jspm install zone=github:angular/zone.js
```

```javascript
import 'zone';
```

Result: `404` when trying to retrieve `zone.js@0.5.0/utils.js.js`,

vicb:
Would be nice to test this.

_Side note: we should expose zone-microtask which is required for ng2_

vicb:
lgtm

PR #129, opened 2015-05-29
test: remove most references to window

vicb:
The goal is to ease the transition to supporting node.js env

@IgorMinar @btford please review


btford:
LGTM

PR #131, opened 2015-05-29
fix(PatchClass): copy static properties

### LINKED ISSUE #127: zones.js mangles XMLHttpRequest constants
the [`patchClass()` function](https://github.com/angular/zone.js/blob/8f262aadcba3a9e3899d23f988f4b97fd602b588/lib/utils.js#L109) doesn't copy over constants from the patched class. This means that a patched `XMLHttpRequest` won't have the constants `XMLHttpRequest.DONE`, etc. This breaks the common pattern of:  

    var req = new XMLHttpRequest();
    req.onreadystatechange = function() {
        if (req.readyState === XMLHttpRequest.DONE) {
            // do something with the result
        }
    };

`patchClass()` should copy over static properties of a class.

vicb:
fixes #127

Without the fix, the added test would fail on Safari where `patchClass` is used to patch MLHttpRequest`

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -146,7 +146,13 @@ function patchClass(className) {
         });
       }
     }(prop));
-  };
+  }
+
+  for (prop in OriginalClass) {
+    if (prop !== 'prototype' && OriginalClass.hasOwnProperty(prop)) {
+      global[className][prop] = OriginalClass[prop];
+    }
+  }
 };
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


btford:
I'm reviewing now.

btford:
LGTM, merge

vicb:
merged via b91f8fec98591d1ab75f948119130c570baf16af

PR #134, opened 2015-06-17
This is needed to make sure that we don't try to bind to non-functions

mhevery:


mhevery:
needs tests

btford:
I added a test – I'll merge this as soon as it goes green.

btford:
Landed as ee4262a7437820480fd6d01131fb5f705a634dda.

PR #136, opened 2015-06-18
add specs for FileReader

robwormald:
from https://github.com/angular/angular/issues/2533

this adds two specs, one using addEventListener that works, the other using the reader.onloadend event that fails with `The zone [1] is not a direct child of the zone [4]` - i cribbed this from the similar XHR spec, so apologies if my logic is wrong here. 

see also plunker demo http://plnkr.co/edit/glcsfUFJo8RlE1sn55Cr?p=preview

i'm gonna have a crack at fixing this next, but wanted to get this in in case its something stupid i'm missing.

[Review on Reviewable](https://reviewable.io/reviews/angular/zone.js/136)


robwormald:
so the simple fix works for chrome and firefox, but borks on Safari and IE.

I referenced the hilarity going on in the websocket patch, and had a crack at fixing that - the latest commit fixes the zone issue (and a bunch of formatting, sorry) but isn't quite right as it loses the constants and readyState seems to drop out as well. I started going around in circles, so perhaps somebody else could have a look.

PR #138, opened 2015-06-18
Fix conflict with Polymer elements

robdodson:
I noticed when including Polymer 1.0 elements in an Angular 2 app, this line throws an error saying it can't find `value` of undefined. I've added a simple workaround which seems to fix the issue and (so far at least) doesn't seem to cause any additional issues with Polymer or Angular.

cc @justinfagnani @addyosmani

[Review on Reviewable](https://reviewable.io/reviews/angular/zone.js/138)

### DIFF
--- a/lib/patch/register-element.js
+++ b/lib/patch/register-element.js
@@ -20,7 +20,7 @@ function apply() {
       callbacks.forEach(function (callback) {
         if (opts.prototype.hasOwnProperty(callback)) {
           var descriptor = Object.getOwnPropertyDescriptor(opts.prototype, callback);
-          if (descriptor.value) {
+          if (descriptor && descriptor.value) {
             descriptor.value = global.zone.bind(descriptor.value);
             _redefineProperty(opts.prototype, callback, descriptor);
           } else {


addyosmani:
LGTM (and tests appear to still be passing too)

btford:
Needs a test, but otherwise LGTM.

Might be worthwhile to add an integration test with Polymer as well.

robdodson:
@btford What do you think would be the best way to test this change, just dropping a polymer element into a harness and verifying it doesn't break?

vicb:
/cc @jeffbcross who last modified this code.

An integration test with a polymer element would be great IMO.

btford:
> just dropping a polymer element into a harness and verifying it doesn't break?

Sounds like a good start. I took a look at doing just that, but I'm not familiar enough with Polymer – just getting it setup with our test harness seems tricky. Does Polymer completely rely on html imports, or is there some programmatic way to import as well?

robdodson:
@btford I just asked the team and they said that currently there is no non-html import way of loading polymer :\

tbosch:
So this would still qualify for a webdriver test...

robdodson:
@tbosch @btford I have to admit, I'm not very familiar with the testing setup for zone.js. If either of you can point me in the right direction i can try to get one written

stevenroose:
Is this change also required to get Polymer 1.0 elements working with AngularDart 2?

btford:
@stevenroose – zone.js is not needed in a Dart environment.

@vicb – do you have time to take a look at this sometime?

vicb:
I can probably take a look later this week

justinfagnani:
Since this is a clear bug, and an extremely simple (13 characters!) fix, can we get this merged and add an issue to make a test that would have broken it?

FWIW the test should only need to have a custom element without some of the lifecycle callbacks so that `descriptor` is undefined at that check.

btford:
Sure. At this point it's going to be more pragmatic to merge before this gets bitrotten and lost to the ages.

Landed as 77b4c0d2960a71207137ac006326edb6160ab13b.

I'm going to open a follow-up issue to add an integration test with Polymer.

robdodson:
yay thanks @btford :smiley:

PR #141, opened 2015-07-07
fix(utils): event listener patches break when passed an object implementing EventListener

matthewjh:
Per https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener:

```
listener
The object that receives a notification when an event of the specified type occurs. 
This must be an object implementing the EventListener interface, or simply a JavaScript function.
```

where EventListener is:

```
void handleEvent(
  in Event event
);
```

At present, zonejs breaks if a patched `addEventListener` is passed an object with a `handleEvent` method, because `zone.bind` is called on the handler, which it of course expects to be a function.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -86,21 +86,37 @@ function patchProperties(obj, properties) {
 
 function patchEventTargetMethods(obj) {
   var addDelegate = obj.addEventListener;
-  obj.addEventListener = function (eventName, fn) {
-    fn._bound = fn._bound || {};
-    arguments[1] = fn._bound[eventName] = zone.bind(fn);
+  obj.addEventListener = function (eventName, handler) {
+    var fn;
+    
+    if (handler.handleEvent) {
+      // Have to pass in 'handler' reference as an argument here, otherwise it gets clobbered in
+      // IE9 by the arguments[1] assignment at end of this function.
+      fn = (function(handler) {
+        return function() {
+          handler.handleEvent.apply(handler, arguments);
+        };
+      })(handler);
+    } else {
+      fn = handler;
+    }
+
+    handler._fn = fn;
+    handler._bound = handler._bound || {};
+    arguments[1] = handler._bound[eventName] = zone.bind(fn);
     return addDelegate.apply(this, arguments);
   };
 
   var removeDelegate = obj.removeEventListener;
-  obj.removeEventListener = function (eventName, fn) {
-    if(arguments[1]._bound && arguments[1]._bound[eventName]) {
-      var _bound = arguments[1]._bound;
+  obj.removeEventListener = function (eventName, handler) {
+    if(handler._bound && handler._bound[eventName]) {
+      var _bound = handler._bound;
+      
       arguments[1] = _bound[eventName];
       delete _bound[eventName];
     }
     var result = removeDelegate.apply(this, arguments);
-    global.zone.dequeueTask(fn);
+    global.zone.dequeueTask(handler._fn);
     return result;
   };
 };
--- a/test/patch/element.spec.js
+++ b/test/patch/element.spec.js
@@ -14,17 +14,43 @@ describe('element', function () {
     document.body.removeChild(button);
   });
 
-  it('should work with addEventListener', function () {
+  it('should work with addEventListener when called with a function listener', function () {
+    var clickEvent = document.createEvent('Event');
+    clickEvent.initEvent('click', true, true);
+
     testZone.run(function() {
-      button.addEventListener('click', function () {
+      button.addEventListener('click', function (event) {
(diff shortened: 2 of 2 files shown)


matthewjh:
@vicb, @btford: can someone take a look at this please?

vicb:
@matthewjh could you please add tests for adding / removing listeners ?

matthewjh:
@vicb I had trouble finding where to put the tests for this but I think I've got it now (`zone.js/test/patch/element.spec.js`). Will push some tests today. :)

matthewjh:
@vicb Done tests. Can you take another look please?

Cheers, Matt.

lib/utils.js
```diff
@@ -86,21 +86,34 @@ function patchProperties(obj, properties) {
...
   var addDelegate = obj.addEventListener;
-  obj.addEventListener = function (eventName, fn) {
-    fn._bound = fn._bound || {};
-    arguments[1] = fn._bound[eventName] = zone.bind(fn);
+  obj.addEventListener = function (eventName, handler) {
+    var fn;
+    
+    if (handler.handleEvent) {
+      // Have to pass in 'handler' reference as an argument here, otherwise it gets clobbered in
+      // IE9 by the arguments[1] assignment on line 106.
```
vicb: please replace 106 by a reference (ie a comment). l 106 is already outdated.

lib/utils.js
```diff
@@ -86,21 +86,34 @@ function patchProperties(obj, properties) {
...
+      fn = (function(handler) {
+        handler.handleEvent.apply(handler, arguments);
+      }).bind(this, handler);
```
vicb: do we really need `.bind()` here ?

test/patch/element.spec.js
```diff
@@ -24,7 +24,25 @@ describe('element', function () {
     button.click();
   });
 
-  it('should respect removeEventListener', function () {
+  it('should work with addEventListener when called with an EventListener-implementing listener', function () {
+    var eventListener = {
+      x: 5,
+      handleEvent: function() {
```
vicb: test the arg (ie click event) ?

vicb:
lgtm with a few comments

lib/utils.js
matthewjh: Yes, unfortunately. It seems that closing over an argument and then subsequently reassigning `arguments[i]` in that scope breaks in IE9. That's why I had to use `bind`.

vicb: Have you tried something like:

```
fn = (function (handler) { 
  return function() {
    handler.handleEvent.apply(handler, arguments);
  } 
})(handler);
```

well, not sure if it's nice (provided it works)

matthewjh: I'll give that a go. One thing I just caught by adding a test for the `event` argument is that my original implementation means that the wrong arguments will be passed to `handlerEvent` (first arg will be `handler` reference). I'll push that now and see if IE9 tests pass.

matthewjh:
:+1: Thanks for your comments, Victor. I've addressed them all.

@vicb

lib/utils.js
```diff
@@ -86,21 +86,36 @@ function patchProperties(obj, properties) {
...
+    handler._fn = fn;
+    handler._bound = handler._bound || {};
+    arguments[1] = handler._bound[eventName] = zone.bind(fn);
     return addDelegate.apply(this, arguments);
   };
 
   var removeDelegate = obj.removeEventListener;
-  obj.removeEventListener = function (eventName, fn) {
+  obj.removeEventListener = function (eventName, handler) {
     if(arguments[1]._bound && arguments[1]._bound[eventName]) {
```
vicb: what about changing the read of `arguments[1]` to `handler`, IMO it would make the code clearer.

matthewjh: Agreed. line 114 will have to stay the same, though, unless we want to 'manually' pass each argument to the delegate.

matthewjh: Or we can do exactly that (pass the arguments to the delegate manually) in both the `removeEventListener` and `addEventListener` patches and remove references to `arguments`, which would also, I believe, eliminate the weird IE9 issue I mentioned.

What do you think? Only benefit of mutating/accessing `arguments` directly is that we don't have to consider how many arguments the delegate should receive. If an optional third argument to add/removeEventListener was added in some exotic browser, the patch would still work.

vicb: only modifying the 3 first (read) accesses should be fine. I'll merge once Travis is green.

Thank you very much for your work on this PR.

matthewjh:
Done. No problem.

@vicb

vicb:
landed as af88ff8a01c1d5b445eeb212569dde5a88c8e509
Thanks !

Issue #142, opened 2015-07-10
addEventListener hook?


matthewjh:
Is the `addEventListener` hook implemented? I couldn't get it to work at all.

vicb:
What are you trying to do ?

What browser, what version of zone.js, do you have a code sample ?

matthewjh:
```javascript
	    var myZone = zone.fork({
	        addEventListener: function() {
	            console.log('addEventListener hook: ', arguments);
	        }
	    });
	    myZone.run(function() {
	        window.addEventListener('click', function() {
	            console.log('click!');
	        });
	    });
```

When I run that snippet with the master dist copy of zone.js loaded in, `"click!"` is logged when you click in the window but `"addEventListener hook"` isn't. I searched the codebase for `zone.addEventListener` and it turned up nothing, hence my asking whether this feature had been implemented.

If it hasn't been implemented, I'm interested in submitting a PR for it, but I wanted to check first.

btford:
I think you want:

```
        var myZone = zone.fork({
            afterTask: function() {
                console.log('addEventListener hook: ', arguments);
            }
        });
        myZone.run(function() {
            window.addEventListener('click', function() {
                console.log('click!');
            });
        });
```

Zone is a generic mechanism that fires on any async operation. Are you just trying to capture clicks? What's your specific use case?

matthewjh:
Here, I have no specific use case. I'm just following the readme which explicitly says there's an addEventListener hook, which it says "allows you to intercept calls to EventTarget.addEventListener"

matthewjh:
My use case is that of someone following the readme. :p

vicb:
If you don't have any specific use case, may be the best would be to drop this feature (from the doc).

My concern is that `patchEventTargetMethods()` [is used in different ways](https://github.com/angular/zone.js/blob/master/lib/patch/event-target.js) according to what the browser supports.

@btford what do you think ?

matthewjh:
A very general use case would be if you want to do some action before or after a task, or otherwise wrap the callback function, but only for event listeners. What if you wanted to report all effecting events in an app to an analytics service? What if you want custom error handling for event listeners? IMO it could be pretty useful to have. 

More generally, it would be cool if you could supply custom patches for `add/removeEventListener`, `setTimeout`, `alert`, etc., on a per-zone basis, which is what the readme seems to hint at with interceptors.

vicb:
> What if ...

[...]

PR #144, opened 2015-07-13
feat(utils): add/removeEventListener interceptor hooks

matthewjh:
Related to #142.

README.md:
### `zone.addEventListener`

This hook allows you to intercept calls to `EventTarget.addEventListener`.

```javascript
var myZone = zone.fork({
  addEventListener: function(delegate) {
    return function addEventListener(name, listener) {
      if (name === 'click') clickListeners++;
      delegate.apply(this, arguments);
    }
  }
});
```

### `zone.removeEventListener`

This hook allows you to intercept calls to `EventTarget.removeEventListener`.

```javascript
var myZone = zone.fork({
  removeEventListener: function(delegate) {
    return function removeEventListener(name, listener) {
      if (name === 'click') clickListeners--;
      delegate.apply(this, arguments);
    }
  }
});
```

Feedback wanted.

I wonder whether this will pass the IE9 tests...

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ Transforms a function to run within the given zone.
 ### `zone.fork`
 
 ```javascript
-zone.fork({
+var myZone = zone.fork({
   onZoneCreated: function () {},
   beforeTask: function () {},
   afterTask: function () {},
@@ -190,7 +190,8 @@ zone.fork({
   setInterval: function () {},
   alert: function () {},
   prompt: function () {},
-  addEventListener: function () {}
+  addEventListener: function () {},
+  removeEventListener: function () {}
 });
 myZone.run(function () {
   // woo!
@@ -236,6 +237,31 @@ While in this zone, calls to `window.setTimeout` will redirect to `zone.setTimeo
 
 This hook allows you to intercept calls to `EventTarget.addEventListener`.
 
+```javascript
+var myZone = zone.fork({
+  addEventListener: function(delegate) {
+    return function addEventListener(name, listener) {
+      if (name === 'click') clickListeners++;
+      delegate.apply(this, arguments);
+    }
+  }
+});
+```
+
+### `zone.removeEventListener`
+
+This hook allows you to intercept calls to `EventTarget.removeEventListener`.
+
+```javascript
+var myZone = zone.fork({
+  removeEventListener: function(delegate) {
+    return function removeEventListener(name, listener) {
+      if (name === 'click') clickListeners--;
+      delegate.apply(this, arguments);
+    }
+  }
+});
+```
 
 ## Status
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -84,41 +84,76 @@ function patchProperties(obj, properties) {
     });
 };
 
-function patchEventTargetMethods(obj) {
+function patchAddEventListener(obj) {
   var addDelegate = obj.addEventListener;
-  obj.addEventListener = function (eventName, handler) {
+
+  function addEventListenerBase (eventName, listener) {
     var fn;
     
-    if (handler.handleEvent) {
-      // Have to pass in 'handler' reference as an argument here, otherwise it gets clobbered in
+    if (listener.handleEvent) {
+      // Have to pass in 'listener' reference as an argument here, otherwise it gets clobbered in
       // IE9 by the arguments[1] assignment at end of this function.
-      fn = (function(handler) {
+      fn = (function(listener) {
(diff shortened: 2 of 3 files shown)


matthewjh:
@vicb, @btford: I think most of the work here is done. Can you kindly take a look and let me know what you think.

Cheers.

matthewjh:
@vicb can you still look at this?

vicb:
@matthewjh that is no more a WIP right ? I'll take a look today or tomorrow

matthewjh:
@vicb Yep, exactly. I've just removed WIP from the title.

And thanks.

vicb:
@matthewjh I don't this code is correct - at least it is not consistent with other hooks.

I guess you want to add `addEventListener()` and `removeEventListener()` in [core.js after the other hooks](https://github.com/matthewjh/zone.js/blob/02772ac4f23aa386219d4147c426137c741cdff8/lib/core.js#L123)

Then the syntax of your examples would become:

```
someZone.fork({
  `$addEventListener`: function (parentAddEventListener) {
    return function () {
      //...
    };
  }
})
```

Please update your PR and rebase on the latest master. Thanks.

README.md
```diff
@@ -236,6 +237,31 @@ While in this zone, calls to `window.setTimeout` will redirect to `zone.setTimeo
 
 This hook allows you to intercept calls to `EventTarget.addEventListener`.
 
+```javascript
+var myZone = zone.fork({
+  addEventListener: function(delegate) {
```
vicb: should be `$addEventListener: ...`

README.md
```diff
@@ -236,6 +237,31 @@ While in this zone, calls to `window.setTimeout` will redirect to `zone.setTimeo
...
+});
+```
+
+### `zone.removeEventListener`
+
+This hook allows you to intercept calls to `EventTarget.removeEventListener`.
+
+```javascript
+var myZone = zone.fork({
+  removeEventListener: function(delegate) {
```
vicb: same thing here `$...`

lib/utils.js
```diff
@@ -84,41 +84,76 @@ function patchProperties(obj, properties) {
...
+}
+
+function patchEventTargetMethods(obj) {
+  function defaultInterceptor(delegate) {
+    return function() {
+      return delegate.apply(this, arguments);
+    }
+  }
+
+  global.zone.addEventListener = defaultInterceptor;
```
vicb: we don't want to do this when `patchEventTargetMethods` is called but rather have the hook defined in `core.js`

matthewjh:
@vicb: Thanks for taking a look!

My logic for doing it this way was to be consistent with the `zone.setTimeout`/`setInterval`/`alert`/`prompt` hooks. See https://github.com/angular/zone.js/blob/master/lib/patch/functions.js#L75:

````javascript
function patchSetFunction(obj, fnNames) {
  fnNames.forEach(function (name) {
    var delegate = obj[name];

    if (delegate) {
      global.zone[name] = function (fn) {
        var fnRef = fn;
        arguments[0] = function () {
          return fnRef.apply(this, arguments);
        };
        var args = utils.bindArgumentsOnce(arguments);
        return delegate.apply(obj, args);
      };

      obj[name] = function () {
        return zone[name].apply(this, arguments);
      };
    }
  });
};

function patchFunction(obj, fnNames) {
  fnNames.forEach(function (name) {
    var delegate = obj[name];
    global.zone[name] = function () {
      return delegate.apply(obj, arguments);
    };

    obj[name] = function () {
      return global.zone[name].apply(this, arguments);
    };
  });
};
````

https://github.com/angular/zone.js/blob/master/test/zone.spec.js#L98

````javascript
    it('should allow you to override alert', function () {
      var spy = jasmine.createSpy();
      var myZone = zone.fork({
        alert: spy
      });

      myZone.run(function () {
        alert('foo');
      });

      expect(spy).toHaveBeenCalled();
    });
  });
````

I'm happy to modify the code to mirror `beforeTask`, `onError`, etc. I would say that `beforeTask`, `afterTask`, etc. are a different group of hooks to `alert`, `setTimeout`, `addEventListener`, the former being akin to zone-specific 'events' while the latter to custom patches / interceptors. Just thought I'd mention this in case there was a reason why the `alert` etc. hooks are slightly different in the first place.

What are your thoughts?

vicb:
Isn't `patchSetFunction()` special it is used to patch several functions (`setTimeout` & `setInterval`).

There should be no reason to have this level of complexity for `addEventListener` ?

PR #145, opened 2015-07-15
Add web-worker zone support

jteplitz602:
Added basic web-worker support. Looking into how to properly test.


btford:
@vicb – do you have any ideas here?

vicb:
What is the difference of the ww version ? patching less APIs (only the one available) ?

If that's the case I'm not sure if extensive tests would help (if we miss some APIs in the src, we'll probably miss them in the test too) and we should be good to go.

Idea: If the only difference is patching less APIs, could we detect them and patch only when they are available (or globally detect the WW context)

jteplitz602:
That's a good idea. @btford what do you think?

jteplitz602:
@btford, @vic: Submitted a new PR #151 that does that instead.

Issue #143, opened 2015-07-11
bug: `requestAnimationFrame` zone grows forever


justindujardin:
When using a basic requestAnimationFrame render loop, the zone grows forever. 
Related to https://github.com/angular/angular/issues/2912

```javascript
zone.fork().run(function () {
    function frameCallback() {
        // frame++
        requestAnimationFrame(frameCallback);
    }
    requestAnimationFrame(frameCallback);
});
```

Example: https://jsfiddle.net/9dqzky6g/3/

justindujardin:
It looks like subsequent calls to `requestAnimationFrame` create child Zones, so unless your `requestAnimationFrame` loop stops, you will add a child zone for each invocation.

zoneBoundFn eats up all of the profile idle time in a few thousand frames.

justindujardin:
Anyone care to comment on this?  It is making it hard for me to run my app for more than a few minutes with Angular2.  Everything is wonderfully fast, until the zones start piling up.

I'm willing to look at fixing the issue if I could get some direction on where to go with it.

justindujardin:
Updated example with patch applied: https://jsfiddle.net/cg9fa5nf/1/

vicb:
closed via #147

btford:
:clap: Thanks @vicb

PR #147, opened 2015-07-17
fix(patch): do not leak zones when wrapping requestAnimationFrame

justindujardin:
# Problem
requestAnimationFrame is often called from within the callback function that is bound to its zone.  This results in every subsequent call to the same function creating and binding to a new zone.

# Proposed Solution
Maintain a simple reference count of function signatures given to requestAnimationFrame, and when the reference count is greater than 0 bind to to the parent zone rather than creating a new one.

Feedback welcome.

Fixes: #143, https://github.com/angular/angular/issues/2912

### DIFF
--- a/lib/patch/functions.js
+++ b/lib/patch/functions.js
@@ -48,6 +48,33 @@ function patchSetClearFunction(obj, fnNames) {
   });
 };
 
+
+/**
+ * requestAnimationFrame is typically recursively called from within the callback function
+ * that it executes.  To handle this case, only fork a zone if this is executed
+ * within the root zone.
+ */
+function patchRequestAnimationFrame(obj, fnNames) {
+  fnNames.forEach(function (name) {
+    var delegate = obj[name];
+    if (delegate) {
+      global.zone[name] = function (fn) {
+        var callZone = global.zone.isRootZone() ? global.zone.fork() : global.zone;
+        if (fn) {
+          arguments[0] = function () {
+            return callZone.run(fn, arguments);
+          };
+        }
+        return delegate.apply(obj, arguments);
+      };
+
+      obj[name] = function () {
+        return global.zone[name].apply(this, arguments);
+      };
+    }
+  });
+};
+
 function patchSetFunction(obj, fnNames) {
   fnNames.forEach(function (name) {
     var delegate = obj[name];
@@ -86,5 +113,6 @@ function patchFunction(obj, fnNames) {
 module.exports = {
   patchSetClearFunction: patchSetClearFunction,
   patchSetFunction: patchSetFunction,
+  patchRequestAnimationFrame: patchRequestAnimationFrame,
   patchFunction: patchFunction
 };
--- a/test/patch/requestAnimationFrame.spec.js
+++ b/test/patch/requestAnimationFrame.spec.js
@@ -3,61 +3,51 @@
 describe('requestAnimationFrame', function () {
   var testZone = zone.fork();
 
-  describe('requestAnimationFrame', function () {
-    it('should run the passed callback in a zone', function (done) {
-      testZone.run(function() {
-        if (typeof requestAnimationFrame === 'undefined') {
-          console.log('WARNING: skipping requestAnimationFrame test (missing this API)');
-          return done();
-        }
+  var functions = [
+    'requestAnimationFrame',
+    'webkitRequestAnimationFrame',
+    'mozRequestAnimationFrame'
+  ];
 
-        // Some browsers (especially Safari) do not fire requestAnimationFrame
-        // if they are offscreen. We can disable this test for those browsers and
-        // assume the patch works if setTimeout works, since they are mechanically
-        // the same
-        requestAnimationFrame(function () {
-          expect(zone).toBeDirectChildOf(testZone);
+  functions.forEach(function (fnName) {
+    var rAF = window[fnName];
+    describe(fnName, function () {
+      if (typeof rAF === 'undefined') {
+        console.log('WARNING: skipping ' + fnName + ' test (missing this API)');
+        return;
+      }
+
+      it('should be tolerant of invalid arguments', function (done) {
+        testZone.run(function () {
+          // rAF throws an error on invalid arguments, so expect that.
+          expect(function () {
+            rAF(null);
+          }).toThrow();
           done();
         });
       });
-    });
-  });
-
-  describe('mozRequestAnimationFrame', function () {
-    it('should run the passed callback in a zone', function (done) {
-      testZone.run(function() {
-        if (typeof mozRequestAnimationFrame === 'undefined') {
-          console.log('WARNING: skipping mozRequestAnimationFrame test (missing this API)');
-          return done();
-        }
-
-        // Some browsers (especially Safari) do not fire mozRequestAnimationFrame
-        // if they are offscreen. We can disable this test for those browsers and
-        // assume the patch works if setTimeout works, since they are mechanically
-        // the same
-        mozRequestAnimationFrame(function () {
-          expect(zone).toBeDirectChildOf(testZone);
-          done();
+      it('should execute callback function in a zone', function (done) {
+        testZone.run(function () {
+          rAF(function () {
+            expect(zone === testZone).toBe(true);
+            done();
+          });
         });
       });
-    });
-  });
 
-  describe('webkitRequestAnimationFrame', function () {
-    it('should run the passed callback in a zone', function (done) {
-      testZone.run(function() {
-        if (typeof webkitRequestAnimationFrame === 'undefined') {
-          console.log('WARNING: skipping webkitRequestAnimationFrame test (missing this API)');
-          return done();
-        }
+      it('should bind to same zone when called recursively', function (done) {
+        testZone.run(function () {
+          var frames = 0;
 
-        // Some browsers (especially Safari) do not fire webkitRequestAnimationFrame
-        // if they are offscreen. We can disable this test for those browsers and
-        // assume the patch works if setTimeout works, since they are mechanically
-        // the same
-        webkitRequestAnimationFrame(function () {
-          expect(zone).toBeDirectChildOf(testZone);
-          done();
+          function frameCallback() {
+            expect(zone === testZone).toBe(true);
+            if (frames++ > 15) {
+              return done();
+            }
+            rAF(frameCallback);
+          }
+
+          rAF(frameCallback);
         });
       });
     });
--- a/README.md
+++ b/README.md
@@ -232,6 +232,12 @@ For instance `clearTimeout` and `removeEventListener`.
 These hooks allow you to change the behavior of `window.setTimeout`, `window.setInterval`, etc.
 While in this zone, calls to `window.setTimeout` will redirect to `zone.setTimeout`.
 
+### `zone.requestAnimationFrame`, `zone.webkitRequestAnimationFrame`, `zone.mozRequestAnimationFrame`
+
+By default `requestAnimationFrame()` does not fork the zone, but you may customize this behavior 
+by using this hook. 
+
+
 ### `zone.addEventListener`
 
 This hook allows you to intercept calls to `EventTarget.addEventListener`.
--- a/lib/patch/browser.js
+++ b/lib/patch/browser.js
@@ -17,7 +17,7 @@ function apply() {
     'immediate'
   ]);
 
-  fnPatch.patchSetFunction(global, [
+  fnPatch.patchRequestAnimationFrame(global, [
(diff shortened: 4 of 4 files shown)


vicb:
@justindujardin Thanks for the PR, I'll try to take a look but I have limited availability until the start mid-August (may be someone else can take a look before).

Please remove the commit that updates the dist folder (when only update generated files when we release a new version). Thanks.

justindujardin:
@vicb Sure, snipped that commit out.

The problem was originally noted using angular2 dev preview.  I'll produce a custom build of angular2 that uses my fork of zone.js, but this issue will pretty severely degrade the performance of any app that uses a rAF time loop. If you have guidance on who I could reach out to about this, I'm happy to do so.   

For now, my issue is resolved by a custom angular2 build.

jomeier:
@vicb @justindujardin Whats the status of this one?

btford:
I'm going to look at this with @caitp next week. Looks like this is about ready to go!

justindujardin:
@btford That's great.  Let me know if anything needs to change or is unclear.

btford:
@vicb can you take a look at this please?

vicb:
@justindujardin could you please squash the commits to ease the review ? (The commit label should rather be feat). I'll review by next Tuesday (not sure to have time tonight). Thanks.

justindujardin:
@vicb no problem.

lib/patch/functions.js
```diff
@@ -48,6 +48,58 @@ function patchSetClearFunction(obj, fnNames) {
...
+ *
+ *  1) Calculate the function signature of the given callback, and keep a reference count
+ *     on it.
+ *  2) If the patched function is called from within its own callback function, bind to the
+ *     parent zone, and then increment the reference count.
+ *  3) After the callback function has executed, decrement the reference count.
+ *  4) If the reference count is <= 0, clear the cached function signature and dequeue
+ *     the current task.
+ */
+function patchRepeatingCallbackFunction(obj, fnNames) {
```
vicb: `patchRequestAnimationFrame` ? _(Not sure)_

lib/patch/functions.js
```diff
@@ -48,6 +48,58 @@ function patchSetClearFunction(obj, fnNames) {
...
+  fnNames.forEach(function (name) {
+    var delegate = obj[name];
+    if (delegate) {
+      var ids = {};
+      global.zone[name] = function (fn) {
+        if(fn){
```
vicb: minor: `if (fn) {` - white spaces

lib/patch/functions.js
```diff
@@ -48,6 +48,58 @@ function patchSetClearFunction(obj, fnNames) {
...
+          // Reference count functions by their toString() value.
+          var fnRef = fn, id = fn.toString();
```
vicb: consistency:
```
var fnRef = fn;
var id = fn.toString();
```

lib/patch/functions.js
```diff
@@ -48,6 +48,58 @@ function patchSetClearFunction(obj, fnNames) {
...
+          arguments[0] = function () {
+            fnRef.apply(this, arguments);
+            ids[id]--;
+            if (ids[id] <= 0) {
+              delete ids[id];
+              zone.dequeueTask();
```
vicb: zone.dequeueTask(fnRef);

vicb:
@justindujardin I have added (mostly minor) comments in the PR.

Two questions:
- What about adding a property to the callback instead of maintaining an array of ids ?
- Couldn't we only say `patchPrototype(global, ['requestAnimationFrame', 'mozRequestAnimationFrame', 'webkitRequestAnimationFrame']` - the callback would be executed in the zone when the RAF is called.

Maybe the first thing would be to consider (or reject) the last bullet above and if it can not work then re-work the implementation in this PR according to the other comment.

What do you think (you probably have spend much more time than me on this).

lib/patch/functions.js
justindujardin: I'm still a bit unsure about how the `dequeueTask` method is supposed to work, because it is unimplemented in the base zone implementation.  Thanks for the correction.

justindujardin:
@vicb I've applied most of your suggestions, but I don't see how the use of `patchPrototype` would solve this problem. The issue with rAF is that it's called recursively from within a callback that is bound to its own zone. Using `patchPrototype` creates a new child zone for each invocation, which causes the tests to fail. Maybe I'm missing something in what you have suggested.

Your suggestion to track the reference count on the function itself resolves a concern that @btford had about how I was calculating function signatures using `toString()`. :+1:

vicb:
> I don't see how the use of patchPrototype would solve this problem ...

you are absolutely right, that was a bad idea.

I'll add an other round of inline comments

lib/patch/functions.js
```diff
@@ -48,6 +48,49 @@ function patchSetClearFunction(obj, fnNames) {
...
+ *     parent zone and increment the reference count.
+ *  2) After the callback function has executed, decrement the reference count.
+ *  3) If the reference count is <= 0, dequeue the current task.
+ */
+function patchRequestAnimationFrame(obj, fnNames) {
+  fnNames.forEach(function (name) {
+    var delegate = obj[name];
+    if (delegate) {
+      global.zone[name] = function (fn) {
+        if (fn) {
```
vicb:
```js
        if (fn) {
          // vicb: this would be called first, moved on top
          if (typeof fn._zonejs_refs === 'undefined') {
            fn._zonejs_refs = 1;
            fn._zonejs_zone = global.zone;       
          } else {
            fn._zonejs_refs++;
          }

          arguments[0] = function () {
            // vicb: moved up (in case the fn throws)
            if (--fn._zonejs_refs === 0) {
              delete fn._zonejs_refs;
            }

            fn.apply(this, arguments);
          };

          // vicb: by using bindOnce dequeue should not be required any more
          arguments[0] = fn._zonejs_zone.bindOnce(arguments[0]);
        }
```

Notes:
- changed the property name to make collision less likely,
- moved up some code so that it match execution order,
- added some comments to explain the changes (should be removed).

Feel free to use it or now

justindujardin: I adopted your changes where possible, but a few subtleties are lost in translation:
 - The decrement cannot happen before the original callback function is executed.  If it does the count will never be > 0 and any rAF calls inside it will not bind to the parent.  
 - In order to ensure the reference is decremented I would need to wrap the `fn.apply` in a try/catch block.  Are there concerns wrapping each callback?

**important**
I also failed to consider the asynchronous recursive function call cases, i.e. what if you call setTimeout -> requestAnimationFrame from within a rAF callback.  The setTimeout call will maintain the zone, because it is patched, and the new call to rAF will have no reference count, and will incorrectly bind a new zone. 

I constructed a simple test case and it demonstrates that the zones grow ad-infinitum in when used as described. 

I have no time to work on this issue now, but I will take a look at it after work is done for the day.  Here's the spec change if you are interested in looking into it.

```javascript
  it('should bind to parent zone when called recursively in an async fashion', function (done) {
    testZone.run(function () {
      var frames = 0;

      function frameCallback() {
        expect(zone).toBeDirectChildOf(testZone);
        if (frames++ > 15) {
          return done();
        }
        setTimeout(function(){
          rAF(frameCallback);
        },1);
      }

      rAF(frameCallback);
    });
  });

```

vicb:
I have been thinking about this a little bit more:

We do fork on bind to be able to track stack traces (going up the zone hierarchy). What this PR does it to prevent forking in order to minimize stack usage.

A simpler way to do this would be not to use `zone.bind()` at all and we could use `zone.run()` directly. That being said, we have to keep in mind that the root zone has a special behavior and we probably want to fork when called from there.

What about something like:

```js
function patchRequestAnimationFrame(obj, fnNames) {
  fnNames.forEach(function (name) {
    var delegate = obj[name];
    if (delegate) {
      global.zone[name] = function (fn) {
        var callZone = global.zone.isRoot() ? global.zone : global.zone.fork();
        if (fn) {
          arguments[0] = function () {
            return callZone.run(fn, arguments);
          };
        }
        return delegate.apply(obj, arguments);
      };

      obj[name] = function () {
        return global.zone[name].apply(this, arguments);
      };
    }
  });
};

```

And then we could something in the doc to describe the special behavior of `requestAnimationFrame()` and how to override it thanks to the `requestAnimationFrame` hook ?

@justindujardin @btford what do you think ?

justindujardin:
@vicb I like your solution, it is much more direct.  I've updated the patch function to work like you suggest.  :+1: 

The async rAF loop is still a problem, unfortunately.  If you start a rAF loop and use setTimeout, or some other forking patched function, then call rAF from within that callback, it will cause setTimeout to become the culprit in the ever-growing stack problem that kills performance.  I'm not sure what to do about this case.  

I have a test that exhibits the problem behavior commented out at the bottom of the spec file.  Thoughts?

lib/patch/functions.js
```diff
@@ -48,6 +48,36 @@ function patchSetClearFunction(obj, fnNames) {
   });
 };
 
+
+/**
+ * Patch requestAnimationFrame
```
vicb: not sure if this is very useful, the description alone should be enough

test/patch/requestAnimationFrame.spec.js
```diff
@@ -3,63 +3,82 @@
...
+      //    function frameCallback() {
+      //      // For this check, since we're injecting ANOTHER zone by using
+      //      // setTimeout, we need to expect that the parent zone will be
+      //      // our test zone.
+      //      var expectedZone = frames ? zone.parent : zone;
+      //      expect(expectedZone === testZone).toBe(true);
+      //      if (frames++ > 15) {
+      //        return done();
+      //      }
+      //      setTimeout(function () {
```
vicb: Do you see any reason why someone would want to do this ?

I don't and I would delete this test - we can re-open if there is a real use case later on.

Thoughts ?

justindujardin: Sadly, I can easily see someone doing this.  Imagine you want to use requestAnimationFrame but only periodically, so you call setTimeout in your handler and then request the next frame when the handler is triggered.  Sure this would be misguided, but hopefully zone.js wouldn't fail to work seamlessly because you've constructed some questionable code.

I resolved the case by using a hybrid of the two approaches that you have suggested.   I store a reference to single forked zone per callback function.  This satisfies all the tests.

If this isn't satisfactory I can revisit it after work.

vicb:
Looks like you have reverted the last changes ???

I was ok with them. The only little missing thing was an update to the README about `zone.rAF, zone.wkRAF, ...` hook, quickly describing the behavior and hook. Something by `By default requestAnimationFrame() does not fork the zone but you can override the behacior using this hook`

Thank you very much for your work on this !

test/patch/requestAnimationFrame.spec.js
vicb: > Sadly, I can easily see someone doing this. Imagine you want to use requestAnimationFrame but only periodically, so you call setTimeout in your handler and then request the next frame when the handler is triggered. Sure this would be misguided, but hopefully zone.js wouldn't fail to work seamlessly because you've constructed some questionable code.

Unless you have a concrete use case in mind, I would revert to the simpler solution. IMO when rAF gets used, you are interested in getting called for each single frame.

justindujardin: Hmm, I don't see the original solution being much simpler, and it definitely fails in this case.

The reason I'm concerned with this issue is that Zone.js is wrapping all code executed in an angular2 app, and given the popularity of angular, I would assume you'll get every kind of bone-headed app setup that is conceivable.  

It's not my project though, so I'll revert this back to what it was originally and remove the test this evening.

vicb: I can also craft a test that fails with your solution when raf is called from 2 different zones. 

So let's come with something that works for most common use cases and adapt later if needed.

justindujardin:
@vicb I'm not sure I understood fully the comments you wanted in the README, please confirm them.  I've reverted to the simpler implementation and removed the async test.

The Travis build seems to be kind of fragile.  It fails intermittently on my commits, with timeouts in unrelated functions on varying operating systems.  I wonder if some of the tests run long enough that slight variations in timings cause them to go over the timeout value?  Have you see this behavior before?

vicb:
Great I will merge later today. 

We have seen some timing related Travis failures in the past. It most probably has nothing to do with the changes here. However can open a ticket and link the falling Travis builds,  that would help.

vicb:
landed as 3d6dc0850b20fd86eef5879f8283de5ee668a668

That will be very useful, thanks @justindujardin for all your follow up

justindujardin:
Thanks for your review, @vicb.  I opened another issue with about the timing sensitive tests.

PR #149, opened 2015-07-27
Foreign window handlers must be ignored.

opichals:
For example the Firefox's WebDriver extension seems to pass in a
handler that resides outside of the zone.js window. Appears to be
this one:
https://github.com/SeleniumHQ/selenium/blob/6dae6c1c19246e5955b04a463b8e99952d6e691b/javascript/firefox-driver/js/utils.js#L1068

Such handlers cannot be modified nor inspected and therefore must be
ignored otherwise selenium reports `Avoid permission denied on
handler 'handleEvent'`

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -88,35 +88,47 @@ function patchEventTargetMethods(obj) {
   var addDelegate = obj.addEventListener;
   obj.addEventListener = function (eventName, handler) {
     var fn;
-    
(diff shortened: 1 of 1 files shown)


mhevery:
Can you please rebase and submit a new PR. Sorry for the long delay

opichals:
I assume this was solved by https://github.com/angular/zone.js/pull/309

PR #154, opened 2015-08-13
fix(utils): addEventListener patch callback binding

opichals:
The callback is to be bound to the current zone for every event
occurence using bindOnce (dequeued immediatelly after finishing).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -108,7 +108,11 @@ function patchEventTargetMethods(obj) {
 
       handler._fn = fn;
       handler._bound = handler._bound || {};
-      arguments[1] = handler._bound[eventName] = zone.bind(fn);
+
+      var listenerZone = global.zone;
+      arguments[1] = handler._bound[eventName] = function() {
+          return listenerZone.bindOnce(fn).apply(this, arguments);
+      }
     }
     return addDelegate.apply(this, arguments);
   };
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -107,7 +107,11 @@ function patchEventTargetMethods(obj) {
 
     handler._fn = fn;
     handler._bound = handler._bound || {};
-    arguments[1] = handler._bound[eventName] = zone.bind(fn);
+
+    var listenerZone = global.zone;
+    arguments[1] = handler._bound[eventName] = function() {
```
matthewjh: Any reason you can't just do `arguments[1] = handler._bound[eventName] = zone.bindOnce(fn);` here?

opichals:
>> +    var listenerZone = global.zone;
>>  +    arguments[1] = handler._bound[eventName] = function() {
>
> Any reason you can't just do arguments[1] = handler._bound[eventName] = zone.bindOnce(fn); here?

The handler callback needs to do the enqueueTask/dequeueTask on every event trigger.

It may be that the right thing might even be something like
```
      var listenerZone = global.zone;
      arguments[1] = handler._bound[eventName] = function() {
          return listenerZone.run(fn, this, arguments);
      }
```
but that would not create a zone fork() which is asserted on various places in tests to happen

mhevery:
obsolete, should be covered in the latest zone rewrite.

Issue #150, opened 2015-08-06
"Expecting function" error under Microsoft Edge with zone.js 0.5.2 & Angular 2 alpha 33


dsebastien:
Currently, the 'patchEventTargetMethods' method in utils.js seems to cause an error under Microsoft Edge. The bind operation fails because the function object that is passed is not of type 'function' as expected but is an object instance of type 'FunctionWrapper'.

Here's the output I get in the console:
```
SCRIPT5022: Expecting function got: [object FunctionWrapper]
core.js (44,7)
```

Putting a breakpoint in the bind function, I see that typeof fn returns "object".

I'm not sure whether the issue is actually caused by zone.js, angular or another dependency and I don't really know how to go about and debug it..

Here's a basic plunker that has the issue: http://embed.plnkr.co/GohfcaiNrwVRn8Hq42tM/preview

matthewjh:
I've seen this on IE11 as well. My conclusion was that it only occurs when the developer tools are open and is (as far as I could tell) innocuous. 

Also hard to see what could b e done to fix this. I for one could find zero documentation (or even a mention) of `FunctionWrapper` on the internet.

matthewjh:
Furthermore, I would say that Angular 2 is completely unrelated as I was not using Angular 2 when I encountered this.

btford:
Thanks for the report, @dsebastien and @matthewjh. I chatted with @vicb, and he said he's is going to take a look at this next week.

dsebastien:
FYI FunctionWrapper seems to be part of Microsoft's JScript, their implementation of EcmaScript included in the .Net framework & used in IE. I assume that they're reusing it for MS Edge.

Here's the only doc about it: https://msdn.microsoft.com/en-US/en_us/library/windows/desktop/microsoft.jscript.functionwrapper(v=vs.110).aspx

dsebastien:
Maybe @csuwildcat could be interested, I saw that he wanted to know about library compat issues with MS Edge (cfr https://github.com/angular/angular/issues/3155)

bestander:
subscribed to the issue, yep all Angular 2 apps can't run in Edge.

vicb:
@mlaval you probably have a much deeper experience debugging win issues than I have. Could you please help here ?

It is obviously linked to https://github.com/angular/zone.js/commit/ee4262a7437820480fd6d01131fb5f705a634dda but reverting the commit (ie the function check) is not enough to get the plunker to work.

[...]

PR #155, opened 2015-08-16
fix(addEventListener patch): ignore FunctionWrapper for IE11 & Edge d…

mlaval:
…ev tools

Fixes the issue described in #150 in the IE and Edge dev tools.
These tools add add event listeners with a special object as a handler, a `FunctionWrapper`. Zone shouldn't interfere with those.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -92,22 +92,24 @@ function patchEventTargetMethods(obj) {
   var addDelegate = obj.addEventListener;
   obj.addEventListener = function (eventName, handler) {
     var fn;
-    
-    if (handler.handleEvent) {
-      // Have to pass in 'handler' reference as an argument here, otherwise it gets clobbered in
-      // IE9 by the arguments[1] assignment at end of this function.
-      fn = (function(handler) {
-        return function() {
-          handler.handleEvent.apply(handler, arguments);
-        };
-      })(handler);
-    } else {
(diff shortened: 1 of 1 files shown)


vicb:
LGTM, thanks @mlaval !

vicb:
merged as 3b0ca3f516ba6f85c3d8ad8dfb91ed03cabfb643

wizarrc:
@mlaval  @vicb please reopen.

https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L368

I don't know if anyone has checked lately, but the latest version of Edge, `data.handler.toString()` does not return `[object FunctionWrapper]` but `function __BROWSERTOOLS_CONSOLE_SAFEFUNC() { [native code] }` instead. I made changes to zone.js/utils.ts to see if the browser events would go away in my performance tab, and sure enough they did.

Can anyone confirm this issue is back in EdgeHTML 15? There should be a better way than just check a string, because they have seemed to changed their internal function. Anyways, below is my local changes and works on my machine.

```
var testString = data.handler.toString();
validZoneHandler = data.handler && (testString === '[object FunctionWrapper]' || testString == 'function __BROWSERTOOLS_CONSOLE_SAFEFUNC() { [native code] }');
```

vicb:
@wizarrc could you please open a new issue. Thanks.

PR #159, opened 2015-08-20
chore(test): make Microtasks test safer in slow env

mlaval:
In the failure, 'mat2' is logged before 'mat1.mit', which can happen in a slow environment, like an Android emulator in SauceLabs.
Trying to make the test safer.

### DIFF
--- a/test/microtasks.spec.js
+++ b/test/microtasks.spec.js
@@ -37,7 +37,7 @@ describe('Microtasks', function () {
 
     setTimeout(function() {
       log.push('mat2');
-    }, 20);
+    }, 30);
 
     setTimeout(function() {
       expect(log).toEqual([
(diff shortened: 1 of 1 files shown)


vicb:
Thanks Marc !

I'm going to run Travis a few times before actually merging.

vicb:
Failed on first try :(

mlaval:
I was able to reproduce the issue consistently on one of the Android emulators that I have.
I realized that I didn't to increase the right timer in this PR ...

With the current version, I don't see issue anymore in that emulator,
neither on SauceLabs after restarting the build a few times.

vicb:
landed as f25ca2fc0604e171e9436b2fd1bb5f0fe8e5815d

PR #161, opened 2015-08-27
feat(patch): support requestAnimationFrame time loops

vicb:
- Add custom patching function for repeating callback functions to
support requestAnimationFrame wrapping.
- Bind requestAnimationFrame callback functions to the current zone
unless it is the root zone.
- DRY up requestAnimationFrame spec file (to remove duplication across
browsers)
- Add test that verifies that when registering frame callbacks from
within a frame callback, the zone in which they execute remains
constant.
- Add test to verify the wrapper is tolerant of invalid arguments.
- Add note to readme about behavior of raf


vicb:
Checking Travis

PR #168, opened 2015-09-01
chore(npm): update es6-promise to 3.0.2, moved to dependencies

### LINKED ISSUE #166: Include es6-promise into dependencies for CommonJS use case
Hi, I'm using angular2 with browserify and noticed that I had to `require('zone.js/lib/browser/zone-microtask')` instead of `require('zone.js')` after a few hours of struggle. But it throws the following error unless I don't `npm install -S zone.js` at the top level of my app.

```
Error: Cannot find module 'es6-promise' from '/path/to/my/app/node_modules/zone.js/lib/browser'
```

What do you think about putting `es6-promise` in `dependencies` instead of `devDependencies`? It is required by `zone-microtask`, which could be considered as a part of the library body, even though it's not in `package.json#main`.

Cheers

vicb:
fixes #166

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,9 +24,11 @@
   "bugs": {
     "url": "https://github.com/angular/zone.js/issues"
   },
+  "dependencies": {
+    "es6-promise": "^3.0.2"
+  },
   "devDependencies": {
     "browserify": "^9.0.8",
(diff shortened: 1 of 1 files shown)


vicb:
merged via #179

PR #171, opened 2015-09-03
Add/remove event listener hooks

matthewjh:
#142 

@vicb Please take a look, assuming travis tests pass.

@scamden Please also take a look. The public API for this is documented in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ Transforms a function to run within the given zone.
 ### `zone.fork`
 
 ```javascript
-zone.fork({
+var myZone = zone.fork({
   onZoneCreated: function () {},
   beforeTask: function () {},
   afterTask: function () {},
@@ -190,7 +190,6 @@ zone.fork({
   setInterval: function () {},
   alert: function () {},
   prompt: function () {},
-  addEventListener: function () {}
 });
 myZone.run(function () {
   // woo!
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -242,7 +241,51 @@ growing the stack size on each recursive call.
...
+This hook allows you to intercept calls to `EventTarget#addEventListener`.
+
+````javascript
+var clickListenerCount = 0;
+
+zone.fork(
+  $addEventListener: function(parentAddEventListener) {
+    return function (type) {
+      if (type === 'click') clickListenerCount++;
+      return parentAddEventListener.call(this, arguments);
```
scamden: should this be apply?

README.md
```diff
@@ -242,7 +241,51 @@ growing the stack size on each recursive call.
...
+This hook allows you to intercept calls to `EventTarget#removeEventListener`.
+
+````javascript
+var clickListenerCount = 0;
+
+zone.fork(
+  $removeEventListener: function(parentRemoveEventListener) {
+    return function (type) {
+      if (type === 'click') clickListenerCount--;
+      return parentRemoveEventListener.call(this, arguments);
```
scamden: should this be apply as well?

README.md
matthewjh: It should. Thanks!

scamden: :)

scamden:
this looks like it would solve my use case! i would still need to patch the handler, which feels slightly hairy but it might be over-engineering to provide an abstraction around overriding the callback

matthewjh:
@scamden I take your point and can sympathise. This however seems like the simplest solution for now -- at least it provides max flexibility.

Another problem with overriding the handler/listener is that the `listener` argument can be an [EventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventListener) object rather than a function. I think we can abstract that out easily so that the hook can always assume that the listener will be a function, otherwise the consumer will have to put in their own nasty logic around a `listener.apply` call.

lib/core.js
```diff
@@ -119,7 +119,13 @@ Zone.prototype = {
   onZoneCreated: function () {},
   afterTask: function () {},
   enqueueTask: function () {},
-  dequeueTask: function () {}
+  dequeueTask: function () {},
+  addEventListener: function () {
+    return this['_addEventListener'].apply(this, arguments);
```
vicb: `this._addEventListener` ?

scamden:
ya i figured it wouldn't make sense to add, it's a bit overkill, just wanted to note it. this seems great!

lib/core.js
matthewjh: See https://github.com/angular/zone.js/pull/171/files#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR92.

The `this` binding from the `addEventListener` call is used for the `apply` on `global.zone.addEventListener`. I don't see how else you could implement this as `patchEventTargetMethods` can be called on different objects when `EventTarget` isn't implemented.

vicb: I only mean `this._add` instead of `this['_add']`

vicb:
LGTM with one minor nit.

lib/core.js
matthewjh: Oh yeah -- not sure why I did that. Will correct.

matthewjh:
@vicb Nit fixed, cheers.

vicb:
landed as 18974403845a68dd450180b80c18fe20cb7a7959
Thanks !

scamden:
thank you guys!

Issue #172, opened 2015-09-04
Use symbols if env supports them when attaching properties to foreign objects


matthewjh:
e.g. in https://github.com/angular/zone.js/blob/master/lib/utils.js#L91 properties such as `_fn`, `_addEventListener` are attached to foreign objects which risks interactions with any other libraries that may also be using these properties.

I think we could use `Symbol`s if the environment supports them, with strings as a fallback, to attach/read these properties.

@vicb

btford:
:+1: 

PR welcome

PR #173, opened 2015-09-04
Use symbols if possible for private properties on exposed objects

matthewjh:
#172 

Seeing whether this passes on Travis…

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var keys = require('./keys');
+
 function Zone(parentZone, data) {
   var zone = (arguments.length) ? Object.create(parentZone) : this;
 
@@ -121,10 +123,10 @@ Zone.prototype = {
   enqueueTask: function () {},
   dequeueTask: function () {},
   addEventListener: function () {
-    return this._addEventListener.apply(this, arguments);
+    return this[keys.common.addEventListener].apply(this, arguments);
   },
   removeEventListener: function () {
-    return this._removeEventListener.apply(this, arguments);
+    return this[keys.common.removeEventListener].apply(this, arguments);
   }
 };
 
--- a/lib/keys.js
+++ b/lib/keys.js
@@ -1,0 +1,23 @@
+/**
+ * Creates keys for `private` properties on exposed objects to minimize interactions with other codebases.
+ * The key will be a Symbol if the host supports it; otherwise a prefixed string.
+ */
+if (typeof Symbol !== 'undefined') {
+  function create(name) {
+    return Symbol(name);
+  } 
+} else {
+  function create(name) {
+    return '_zone$' + name;
+  }
+}
+
+var commonKeys = {
+  addEventListener: create('addEventListener'),
+  removeEventListener: create('removeEventListener')
+};
+
+module.exports = {
+  create: create,
+  common: commonKeys
+};
--- a/lib/patch/define-property.js
+++ b/lib/patch/define-property.js
@@ -1,11 +1,14 @@
 'use strict';
 
+var keys = require('../keys');
+
 // might need similar for object.freeze
 // i regret nothing
 
 var _defineProperty = Object.defineProperty;
 var _getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
 var _create = Object.create;
+var unconfigurablesKey = keys.create('unconfigurables');
 
 function apply() {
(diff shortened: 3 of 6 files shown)


matthewjh:
Tests look good.

@btford: can you take a look please?

lib/utils.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
+function getKey(name) {
```
vicb:
```
var getKey = typeOf Symbol !== 'undefined' ? function(name) {...} : function(name) {...};
```

lib/core.js
```diff
@@ -121,10 +123,10 @@ Zone.prototype = {
   enqueueTask: function () {},
   dequeueTask: function () {},
   addEventListener: function () {
-    return this._addEventListener.apply(this, arguments);
+    return this[utils.originalAddEventListenerKey].apply(this, arguments);
```
vicb: `keys.addEventListener` ?

lib/patch/define-property.js
```diff
@@ -1,11 +1,14 @@
...
 
+var utils = require('../utils');
+
 // might need similar for object.freeze
 // i regret nothing
 
 var _defineProperty = Object.defineProperty;
 var _getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
 var _create = Object.create;
+var unconfigurablesKey = utils.getKey('unconfigurables');
```
vicb: keys.create('...'); ?

matthewjh: Good point, I'll factor this out into `keys.js`. utils already has a fair bit of stuff in it.

matthewjh:
Thanks @vicb. Just pushed some changes.

lib/keys.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Creates keys for `private` properties on exposed objects to minimize interactions with other codebases.
+ * The key will be a Symbol if the host supports it; otherwise a prefixed string.
+ */
+var create =
```
vicb: may be use `if/else` here ?

lib/keys.js
```diff
@@ -0,0 +1,23 @@
...
+  function(name) {
+    return Symbol(name);
+  } 
+  :
+  function(name) {
+    return '_zone$' + name;
+  };
+
+var commonKeys = {
+  originalAddEventListener: create('originalAddEventListener'),
```
vicb: What about adding `addEventListener` & `removeEventListener` to exports ? `common.original` sounds unnecessarily verbose to me

vicb:
LGTM with a few comments. @btford ?

lib/keys.js
matthewjh: 1) WRT `exports.common`, I prefer having the common keys on that object rather than being exported directly, so as to differentiate the common/shared keys from any (more general/abstract) key related functions. What if we need a common key called `create` in the future?

2) WRT the 'original' prefix, I carried that over from the _ prop name in `patchEventTargetMethods`. It refers to the 'original' method implementation rather than the patch, so personally I like having that explicit in the name. I don't mind either way though so it's up to you.

lib/keys.js
matthewjh: I've changed to an `if/else` statement containing FDs for `create`. Not sure it reads better to be honest, but it's either that, ternary, or wrapping it in an IIFE (nasty IMO).

lib/keys.js
vicb: 1) make key names UPPER_CASE ?

2) IMO `original` can be dropped, it has no added value.

matthewjh: Okay, cool. I removed the `original` prefix.

For 1), I really prefer having the common keys on a separate object. The common keys are a logical group distinct from other keys exports so IMO there's a clear benefit. I could make their export names uppercase, but that doesn't seem any more intuitive and would imply that keys create with `keys.create` elsewhere should have uppercased names for consistency. Unless there's a good reason why we shouldn't export them on their own object, I'd be happier to leave it as is.

vicb:
pls add `fix #172` in the commit comment

matthewjh:
Sure thing. Done.

(Is that so GH auto-closes the issue?)

vicb:
> Is that so GH auto-closes the issue?

Yep, that's it.

vicb:
landed as f8a2be5ab889b3026774ea36cea4e385399a52df

PR #174, opened 2015-09-05
Zone-binds FileReader#onEventName listeners

### LINKED ISSUE #137: FileReader events aren't zoned
Ref: https://github.com/angular/angular/issues/2533

Failing spec PR: https://github.com/angular/zone.js/pull/136

It appears that the doing using reader.addEventListener('loadend', function(... works, but doing the equivalent with reader.onloadend etc isn't captured properly.

matthewjh:
I took a stab at making the changes to close #137. 

No issues as far as I can tell, as `utils.patchClass` does all the leg work. For some reason though, Android 4.3's native browser's `FileReader` doesn't have `addEventListener` or `removeEventListener,` so there's a work around for that.

### DIFF
--- a/lib/patch/file-reader.js
+++ b/lib/patch/file-reader.js
@@ -1,0 +1,11 @@
+'use strict';
+
+var utils = require('../utils');
+
+function apply() {
+  utils.patchClass('FileReader');
+}
+
+module.exports = {
+  apply: apply
+};
--- a/test/patch/FileReader.spec.js
+++ b/test/patch/FileReader.spec.js
@@ -1,0 +1,95 @@
+'use strict';
+
+describe('FileReader', ifEnvSupports('FileReader', function () {
+  var fileReader;
+  var blob;
+  var data = 'Hello, World!';
+  var testZone = zone.fork();
+
+  // Android 4.3's native browser doesn't implement add/RemoveEventListener for FileReader 
+  function supportsEventTargetFns () {
+    return FileReader.prototype.addEventListener && FileReader.prototype.removeEventListener;
+  }
+  supportsEventTargetFns.message = 'FileReader#addEventListener and FileReader#removeEventListener';
+
+  beforeEach(function () {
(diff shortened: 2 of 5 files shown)


matthewjh:
@vicb can you take a look at this?

test/patch/FileReader.spec.js
```diff
@@ -0,0 +1,95 @@
...
+        done();
+      };
+    });
+
+    fileReader.readAsText(blob);
+  });
+
+  it('should have correct readyState', function (done) {
+    fileReader.onloadend = function () {
+      expect(fileReader.readyState).toBe(2);
```
vicb: pls change 2 for `FileReader.DONE` (everywhere this is applicable)

matthewjh: Done

lib/patch/file-reader.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+var utils = require('../utils');
+
+function apply() {
+  utils.patchClass('FileReader');
```
vicb: Quick review, not much time this week:
- What if the patch has been applied in `lib/patch/event-target.js` already ?
- `patchClass('FileReader')` will `new FileReader()`, is that allowed ?

matthewjh: * What if the patch has been applied in lib/patch/event-target.js already ?

The event target patch patches `addEventListener` and `removeEventListener`, which `patchClass` will wrap in addition to taking care of `onXXXX` listeners / other function setters that need to be bound. The two patches achieve separate things and the order in which they're applied shouldn't matter.

* `patchClass('FileReader')` will `new FileReader()`, is that allowed ?

Yes. There's a guard in `patchClass` for environments where the class is falsey. I'm not aware of any other scenarios in which you can't new a `FileReader`.

vicb:
LGTM, thanks !

vicb:
closed via #174

PR #177, opened 2015-09-11
EventTarget patch fixes

matthewjh:
Two bugs fixed here, each in a separate commit:

1. Calling `eventTarget.addEventListener(name, fn)` multiple times with the same `fn` was incorrectly adding a new listener each time. All but the last would be unremovable. Per [the spec](http://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-EventTarget-addEventListener), duplicate listeners with the same args should be discarded.

2. When adding the same listener twice, one with `useCapture=true` and the other with `=false`, per [the spec](http://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-EventTarget-addEventListener) each has to be removed separately with its corresponding `useCapture` argument. The zone patch meant that you could remove one but not the other, leaving an event listener than could not be deregistered.

@vicb @btford

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -115,7 +115,8 @@ function patchEventTargetMethods(obj) {
 
       handler[originalFnKey] = fn;
       handler[boundFnsKey] = handler[boundFnsKey] || {};
-      arguments[1] = handler[boundFnsKey][eventType] = zone.bind(fn);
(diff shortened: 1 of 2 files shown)


vicb:
LGTM

vicb:
landed as 1b804cfeb1b8ca9f9ce184a7b9508f3d13480b32 + previous commit

PR #181, opened 2015-09-11
fix(Jasmine): add support for jasmine 2 done.fail()

### LINKED ISSUE #180: The `done` fn created by jasmine-patch.js does not support `done.fail`
Original issue https://github.com/angular/angular/issues/4136

The [`done.fail` option in Jasmine 2.0](http://jasmine.github.io/2.3/introduction.html#section-53) makes it easy to report the error from a failed promise but `done.fail` isn't implemented by the zone-aware version from *jasmine-patch.js*.

Therefore, the following test reports an async timeout instead of the actual error:
```
  // this test fails the wrong way
  it('async test with rejected promise reports error as expected via "done.fail"', done => {
    Promise.reject('failed on purpose')
      .then(result => {
        fail('should not get here');
      })
      .then(done,
        reason => {
          done.fail(reason); // written this way for easy breakpointing
        });
  });
```
Not a huge priority but worth fixing and easy to fix.

/cc  @wardbell

vicb:
fixes #180 

@IgorMinar Could you please review ?

/cc @wardbell

### DIFF
--- a/lib/jasmine/patch.js
+++ b/lib/jasmine/patch.js
@@ -15,12 +15,20 @@ function apply() {
   var originalIt = global.it;
   var originalFit = global.fit;
 
+  // Patch jasmine `done()`  and `done.fail()` so that they execute in the root zone
+  var patchJasmineDone = function(jasmineDone) {
+    var done = jasmineZone.bind(jasmineDone);
+    if (typeof jasmineDone.fail === 'function') {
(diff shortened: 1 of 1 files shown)


IgorMinar:
lgtm, go ahead and merge

vicb:
landed as https://github.com/angular/zone.js/pull/181

PR #182, opened 2015-09-12
node tests + process/fs/http/events patch

matthewjh:
@vicb 

So I've got some tests running on Node via `jasmine-node`. `npm test` runs both `npm run test.browser` and `npm run test.node`.

There's also some restructuring of the `test` folder to accommodate both browser and node tests:

`test/setup`:
- `browser-microstask.js` (was test/setup-microtask.js)
- `browser.js` (was test/setup.js)
- `node.spec.js` (same as above but for the node tests). Unfortunately ATM it needs to be suffixed with 'spec.js' for `jasmine-node` to run it.

`test/patch`:
- files directly in here should be testing patches that apply to both platforms
- `node/` - tests for node-specific patches
- `browser/` - tests for browser-specific patches

The structure of this folder is intended to mirror that of `lib/patch`.

Also need to take a look at `process.binding` and `async_wrap`.


lib/zone.js
```diff
@@ -1,14 +1,17 @@
...
 
-global.zone = new core.Zone();
+if (!global.zone) {
+  global.zone = new core.Zone();
+  core.Zone.longStackTraceZone = longStackTraceZone;
+
+  nodePatch.apply();
+}
 
 module.exports = {
```
matthewjh: It would be nice if the exports here were `rootZone`, `bindPromiseFn`, and `longStackTraceZone`. Then there'd be no need to put them on the `Zone` ctor on node. But I imagine having the public APIs roughly equal for node and browser is important?

lib/node/zone.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+var core = require('../core');
+var nodePatch = require('../patch/node');
+var longStackTraceZone = require('../zones/long-stack-trace');
+
+global.zone = new core.Zone();
+core.Zone.longStackTraceZone = longStackTraceZone;
+
+module.exports = {
```
matthewjh: It would be nice if the exports here were `rootZone`, `bindPromiseFn`, and `longStackTraceZone`. Then there'd be no need to put them on the Zone ctor on node. But I imagine having the public APIs roughly equal for node and browser is important?

matthewjh:
@vicb, can you take a look at this when you get a chance? I don't want to let it diverge from master too much, due to the file moves.

Also, I'm working on the `EventEmitter` patch now, which should probably reside in a separate PR for legibility.

mhevery:
@juliemr I think this may be of interest as well.

@matthewjh Any chance this could be rebased on latest master?

mhevery:
abandoned

PR #184, opened 2015-09-23
fix(utils): fixes event target patch in web workers

matthewjh:
Calling `addEventListener` in a Web Worker was breaking.

No tests, but we should probably run all the zone tests (if possible) inside a web worker as part of the test suite to ensure compatibility.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -118,7 +118,9 @@ function patchEventTargetMethods(obj) {
       arguments[1] = handler[boundFnsKey][eventType];
     }
 
-    return global.zone.addEventListener.apply(this, arguments);
+    var target = isWebWorker() && !this ? self : this;
+
+    return global.zone.addEventListener.apply(target, arguments);
   };
 
   // This is required for the removeEventListener hook on the root zone.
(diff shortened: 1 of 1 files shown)


vicb:
Thanks for the fix ! 
I'll test this in the context of Angular before merging.

matthewjh:
Maybe. 

If we change the target to `this || self`, then obviously we'll be invoking it with a target of `self` when `this` is falsy. I have no idea how all the browsers normally behave when you invoke these functions with a falsy target (is it specced?). The current solution is somewhat 'safer' in that the `self` path is only taken on web workers.

What do you think?

vicb:
When would `this` be falsy ? (outside of a web worker)

matthewjh:
@vicb 

In 'working' code? I have no idea. But if someone does, in development, the following:
````
'use strict';

var fn = eventTarget.addEventListener;
fn('click', function () {});  
````
Then shouldn't the behaviour be the same (or as close as possible) whether zone is loaded or not?

**Comment by vicb** [comment]
Not sure I get your point here, wouldn't you write `fn.call(eventTarget, 'click', function);` above ?

**Comment by matthewjh** [comment]
You would, yes. But my point was, that if a developer when writing their app *mistakenly* writes:

```
'use strict';

var fn = eventTarget.addEventListener;
fn('click', function () {});  
```

Then the code should error out (or nor as the case may be) whether zone is loaded or not. Basically, the behaviour should be consistent for zone and non-zone code, even if that behaviour is an error. In this case, one would presume that `this` would be falsy inside the `addEventListener` implementation, and the code should run as such.

vicb:
That makes sense.

It seems like Travis testing Angular with this is happy (currently re-running a failed test but looks like a glitch). Once it's green, I'll merge the PR. Thanks !

vicb:
landed as ad5c0c8df779496ce21294cfef590c30bfba0a4a

PR #191, opened 2015-10-06
fix(addEventListener): when called from the global scope

### LINKED ISSUE #190: `addEventListener` in strict mode
When `addEventListener` is called in strict mode, `var target = isWebWorker() && !this ? self : this` would evaluate to `undefined` in `patchEventTargetMethods()` resulting in an error.

Notes:
- in non strict mode, `this` would be set to `window` and there would be no error,
- `window.addEventListener()` would work in both strict & non-strict mode.

/ref https://github.com/angular/angular/issues/4531

vicb:
fixes #190 

@IgorMinar / @btford could you please review.

I'll check within the context of Angular & report here

/ref https://github.com/angular/angular/issues/4531


vicb:
Also see the discussion in https://github.com/angular/zone.js/pull/184

/cc @matthewjh

vicb:
Status:
- There is one issue with this PR on Android,
- I can get my my Angular env to work (most probably related to the node/npm upgrades).

I'll try to solve those tomorrow.

dist/zone-microtask.js
```diff
@@ -1072,11 +1075,15 @@ function patchEventTargetMethods(obj) {
     var eventType = eventName + (useCapturing ? '$capturing' : '$bubbling');
     if (handler[boundFnsKey] && handler[boundFnsKey][eventType]) {
       var _bound = handler[boundFnsKey];
-      
       arguments[1] = _bound[eventType];
       delete _bound[eventType];
     }
-    var target = isWebWorker() && !this ? self : this;
+
+    // - Inside a Web Worker, `this` is undefined, the context is `global` (= `self`)
```
IgorMinar: remove the reference to "self", it only confuses the reader

IgorMinar:
otherwise lgtm

Issue #167, opened 2015-09-01
Safari crashing on reloading remote iOS Safari with Zone


mlynch:
I'm noticing that Safari debugger on OS X has a really tough time handling refresh on a remotely-debugged iOS Safari instance running Zone (and Angular 2). Either Safari will take 10x longer to reload the page, or it will crash on iOS with a notice: "A problem occured with this webpage so it was reloaded."

I was able to get one stack trace to show (and I'm not sure if it's related at all), but this problem is incredibly difficult to debug since the console closes when the refresh fails:

![screenshot 2015-09-01 09 40 06](https://cloud.githubusercontent.com/assets/11214/9607537/588636ba-508f-11e5-8c23-6bc42c085dd0.png)

The error corresponds to this code in zone.js 0.5.2:

``` javascript
      obj[setName] = function () {
        return global.zone[setName].apply(this, arguments);
      };

      var clearDelegate = obj[clearName];

      global.zone[clearName] = function (id) {
        if (ids[id]) {
          delete ids[id];
          global.zone.dequeueTask();
        }
        return clearDelegate.apply(this, arguments);
      };

      obj[clearName] = function () {
        return global.zone[clearName].apply(this, arguments); // Line 3985
      };
    }
  });
};
```

I'll try to create a reproducible test case, but any angular 2 demo should suffice (unless there is an issue with Ionic 2 which I find unlikely).

Brooooooklyn:
aha, I'm building ionic web app with zone.js too. And I'm trying to fix those issues in my zone.js file these day.

JiaLiPassion:
close for now.

PR #193, opened 2015-10-22
It may throw some errors when there were no handler to remove.

Brooooooklyn:
It may throw some errors when there were no handler to remove.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -96,6 +96,9 @@ function patchEventTargetMethods(obj) {
   // This is required for the addEventListener hook on the root zone.
   obj[keys.common.addEventListener] = obj.addEventListener;
   obj.addEventListener = function (eventName, handler, useCapturing) {
+    if (!handler) {
+      return;
+    }
     var eventType = eventName + (useCapturing ? '$capturing' : '$bubbling');
     var fn;
     //Ignore special listeners of IE11 & Edge dev tools, see https://github.com/angular/zone.js/issues/150
@@ -129,6 +132,9 @@ function patchEventTargetMethods(obj) {
   // This is required for the removeEventListener hook on the root zone.
   obj[keys.common.removeEventListener] = obj.removeEventListener;
   obj.removeEventListener = function (eventName, handler, useCapturing) {
+    if (!handler) {
+      return;
+    }
     var eventType = eventName + (useCapturing ? '$capturing' : '$bubbling');
     if (handler[boundFnsKey] && handler[boundFnsKey][eventType]) {
       var _bound = handler[boundFnsKey];
--- a/test/patch/element.spec.js
+++ b/test/patch/element.spec.js
@@ -113,6 +113,20 @@ describe('element', function () {
     expect(eventListener.handleEvent).not.toHaveBeenCalled();
   });
 
+  it('should have no effect while calling addEventListener without listener', function () {
+    expect(function() {
+      button.addEventListener('click', null);
+      button.addEventListener('click', undefined);
+    }).not.toThrowError();
+  });
+
+  it('should have no effect while calling removeEventListener without listener', function () {
+    expect(function() {
+      button.removeEventListener('click', null);
+      button.removeEventListener('click', undefined);
+    }).not.toThrowError();
+  });
+
 
   it('should only add a listener once for a given set of arguments', function() {
     var log = [];


lib/utils.js
```diff
@@ -130,7 +130,7 @@ function patchEventTargetMethods(obj) {
   obj[keys.common.removeEventListener] = obj.removeEventListener;
   obj.removeEventListener = function (eventName, handler, useCapturing) {
     var eventType = eventName + (useCapturing ? '$capturing' : '$bubbling');
```
vicb: From MDN: "Calling removeEventListener() with arguments that do not identify any currently registered EventListener on the EventTarget has no effect."

```
if (!handler) {
  return;
}
```

Is probably the best thing to do ?

Please also add test and check if `addEventListener` should be updated.

Brooooooklyn:
Updated, please check, thx.

lib/utils.js
```diff
@@ -96,6 +96,9 @@ function patchEventTargetMethods(obj) {
   // This is required for the addEventListener hook on the root zone.
   obj[keys.common.addEventListener] = obj.addEventListener;
   obj.addEventListener = function (eventName, handler, useCapturing) {
+    if (!handler || (typeof(handler) !== 'function' && typeof(handler.handleEvent) !== 'function')) {
+      return;
```
vicb: I would keep the first condition only and delegate in all other cases

vicb:
See my comments in the PR.

Also please squash your commits and reword the message to comply with angular convention : fix(event listener): ... See contributing.md in the angular/angular repo.

Thanks

lib/utils.js
Brooooooklyn: I'm not sure what do you mean...If i kept the first condition only then my test would fail...And it's not in line with most people's expectations. See https://github.com/angular/zone.js/issues/167  

And as you said
> From MDN: "Calling removeEventListener() with arguments that do not identify any currently registered EventListener on the EventTarget has no effect."

I think it's the same as addEventListener.

vicb: I'm am not sure how this PR relates to #167, could you please expand ?

I believe that `!handler` is the only condition that could make zone.js code fail. Any other value should not make zone.js fail and then there is no reason why we should not call the underlying `xEventListener` method - zone.js should not try to fix its users errors.

The less code we have the less bugs we have. For ex I think your code might fail on IE because of the `FunctionWrapper` thing (when dev tools are open, the handler could be a `FunctionWrapper` IIRC, see a few lines below your updates).

That being said, there might be some obvious thing I fail to get here ?

wrt to your tests, what would fail ?

Also please update the tests to reflect the original issue:

> It may throw some errors when there were no handler to remove.

```js
expect(function() {
  button.addEventListener('click', eventListenerA);
}).not.toThrowError();
```

Brooooooklyn: ah...It's my fault. I was wrong to attach https://github.com/angular/zone.js/issues/167 to eventListener's issue.  It seem to be another issue.

Anyway, when you fire `addEventListener('someEvent', {a: 123})` or `addEventListener('someEvent', {handleEvent: 'something not a function'})` in browser without zone.js, nothing would happen and have no effect. But when you are firing these in with zone.js, It would throw error. I think it doesn't seem like a good behavior. 

And about the issue about IE9, I had considered about it. Then codes would return when the handler isn't function and handler.handleEvent isn't function. It has no effect to the previous codes.

Brooooooklyn:
I have updated files....please check. Sorry to bother you.

test/patch/element.spec.js
```diff
@@ -113,6 +113,17 @@ describe('element', function () {
     expect(eventListener.handleEvent).not.toHaveBeenCalled();
   });
 
+  it('should have no effect while calling addEventListener / removeEventListener without listener', function () {
+    var eventListener;
+    expect(function() {
```
vicb: Could you please
- removed eventListener - only used undefined / null
- add a test for ie {a: 1} - you had a question about that
- split add & remove in 2 different tests

vicb:
You're not bothering me at all, you're fixing bug and that's great, thanks !

test/patch/element.spec.js
Brooooooklyn: I test in IE9/10/11 addEventListener('click', {a: 2})  , nothing happened, no error was thrown.
image

vicb:
landed as 5bcc6aefad6e2eb30f87ec0b9e9ffcc236971b75
thanks !

Issue #198, opened 2015-11-03
addEventListener has issues with a null callback


alxhub:
```
Uncaught TypeError: Cannot read property 'toString' of null
obj.addEventListener @ angular2.dev.js:1068
Gestures.add @ polymer.html:752
```

Polymer adds a `null` listener to the 'tap' event on `document` (something about enabling document-wide listeners):

```
// enable document-wide tap recognizer
if (event === 'tap') {
  Polymer.Gestures.add(document, 'tap', null);
}
```

Unfortunately zone.js's addEventListener has issues with a `null` callback, and throws a few exceptions (calling `toString()` on the callback and trying to set properties on it).

vicb:
The last commit in master tries to fix this issue... but in the wrong way as we only do not delegate to the original add/remove function when the handled is null. It should be fixed before next zone release.

@Brooooooklyn any chance you can give this issue a look ?

rvraaphorst:
Running into exact this issue, but can't update as (AFAIK) my jspm only knows v0.5.8. Any idea when the next version is released?

PR #199, opened 2015-11-04
origin addEventListener/removeEventListener should be called without eventListener

Brooooooklyn:
Should I commit amend or open a new pr?

### DIFF
--- a/test/patch/element.spec.js
+++ b/test/patch/element.spec.js
@@ -114,17 +114,31 @@ describe('element', function () {
   });
 
   it('should have no effect while calling addEventListener without listener', function () {
+    var eventListenerZone = zone.fork({
+      addEventListener: jasmine.createSpy('addEventListener')
+    });
     expect(function() {
-      button.addEventListener('click', null);
-      button.addEventListener('click', undefined);
+      eventListenerZone.run(function() {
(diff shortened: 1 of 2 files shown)


Brooooooklyn:
I tried to fix https://github.com/angular/zone.js/issues/198 here, please check @vicb

test/patch/element.spec.js
```diff
@@ -114,17 +114,25 @@ describe('element', function () {
   });
 
   it('should have no effect while calling addEventListener without listener', function () {
+    var originAddEventListenerFn = zone.addEventListener;
```
vicb: A nicer way to do that would be to fork the zone installing the spy at that time.

While your are modifying the test, please also check the args (`toHaveBeenCalledWith`)

vicb:
@Brooooooklyn thanks for the fix, it looks good to me - please update the tests according to the comments. Thanks.

Brooooooklyn:
updated, please check

vicb:
landed as 26e7f5177a06b28e9c4478eab5c6b95cd79da37f

Thanks @Brooooooklyn (you might want to add `fixes #<gh issue id>` when you fix gh issues).

PR #200, opened 2015-11-04
output of typeof operator should be string

Brooooooklyn:


lib/patch/mutation-observer.js
```diff
@@ -39,7 +39,7 @@ function patchClass(className) {
   var prop;
   for (prop in instance) {
     (function (prop) {
-      if (typeof global[className].prototype !== undefined) {
+      if (typeof global[className].prototype !== 'undefined') {
```
vicb: Good catch. Could you please add a test for this ?

Brooooooklyn: ah...how to build a test spec for this situation? I'm afraid that I have no experience to build such a test spec...

vicb:
landed as 19703e3

Issue #202, opened 2015-11-11
Add Indexed DB browser patch


gionkunz:
Hey there

I'm using Pouch DB within an Angular 2 project, which is using Indexed DB under the hood and I need to do a manual `ChangeDetectorRef.detectChanges()`, because the Indexed DB events are not proxied through the Zone for change detection.

Any change we can add a patch for Indexed DB?

Cheers
Gion

gionkunz:
`IDBDatabase` inherits from `EventTarget`, would it be that simple to add `IDBDatabase` to https://github.com/angular/zone.js/blob/master/lib/patch/event-target.js#L13-L32 ?

cherrydev:
Apparently this was fixed back in 84a251faabb422a86d91bafc58b1a461cfff6664

PR #204, opened 2015-11-12
Added property patches and event target methods as well as tests for Indexed DB

gionkunz:
This does solve #202 , although I still have problems with PuchDB as described in #203 . I tested in Firefox and Chrome.

### DIFF
--- a/lib/patch/event-target.js
+++ b/lib/patch/event-target.js
@@ -29,7 +29,13 @@ function apply() {
       'WorkerGlobalScope',
       'XMLHttpRequest',
       'XMLHttpRequestEventTarget',
-      'XMLHttpRequestUpload'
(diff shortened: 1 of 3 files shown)


cherrydev:
@gionkunz Any luck on getting this patch working with tests running?  I'm not sure why this failed to begin with, but I notice that at least currently, the idiom for the tests is to use the `global.zone` reference instead of just `zone`.

mhevery:
Could we get this rebased on the latest master?

gdi2290:
IndexedDB support would be great

mhevery:
@gdi2290 could you rebase this on master so that we could merge this in?

mhevery:
abandoned

PR #207, opened 2015-11-21
Replace unicode characters with ASCII

mgiuffrida:
Unicode characters in source code screw up some build systems, like Chromium.

### DIFF
--- a/test/lib/brick.js
+++ b/test/lib/brick.js
@@ -1833,8 +1833,8 @@ if (typeof WeakMap === "undefined") {
     var LEFT_MOUSE_BTN = 0;
     var GET_DEFAULT_LABELS = function() {
         return {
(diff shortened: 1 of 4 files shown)


mgiuffrida:
googlebot: oops, michaelpg@chromium.org is my chromium identity. this is also me.

test/lib/brick.js
```diff
@@ -1833,8 +1833,8 @@ if (typeof WeakMap === "undefined") {
     var LEFT_MOUSE_BTN = 0;
     var GET_DEFAULT_LABELS = function() {
         return {
-            prev: "←",
```
vicb: Coulf you please revert this change, `brick.js` is a 3rd party lib.
As it is used in tests only, it should not affect you ?

PR #210, opened 2015-11-25
Integrate WTF into zone.js

mhevery:
/cc: @IgorMinar @btford @vicb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ blink-core/
 blink-idl/
 parsed-idl/
 /node_modules
+.idea
--- a/lib/patch/functions.js
+++ b/lib/patch/functions.js
@@ -41,6 +41,14 @@ function patchSetClearFunction(obj, fnNames) {
         return clearDelegate.apply(this, arguments);
       };
 
+      global.zone[setName + 'Unpatched'] = function() {
+        return delegate.apply(obj, arguments);
+      };
+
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@ blink-core/
 blink-idl/
 parsed-idl/
 /node_modules
+.idea
```
vicb: -1 you should add it to your own global gitignore

IgorMinar: nah.. this line is to protect us from Misko not to protect Misko from our commits.

I'm fine with keeping it in.

mhevery:
I have updated all with your comments. It fails on IE9, but I don't know how to get it reproduced locally. Could anyone help.

mhevery:
I found the reason why IE9 was failing.

lib/patch/functions.js
```diff
@@ -1,96 +1,100 @@
...
-        var args = bindArgs(arguments);
-        id = delegate.apply(obj, args);
-        ids[id] = true;
-        return id;
+var wtf = require('../wtf');
+var Zone = require('../core').Zone;
+
+function patchSetClearFunction(window, fnNames) {
+  fnNames.forEach(function (setName) {
+    var repeating = setName == 'setInterval';
```
vicb: the caller should pass the info

mhevery: fixed

PR #215, opened 2015-12-11
Make sure we use the native Promise to schedule utask when they're available

vicb:
Due to https://github.com/jakearchibald/es6-promise/issues/140, es6-promise would always override the global Promise in Chrome (because of a very minor bug in Chrome).

Zone would only use a Promise to schedule utask when a native implementation is present, otherwise it would fallback to other mechanism depending on the environment.

Because of es6-promise overwriting the native Promise impl in Chrome, it would use mutation observers to schedule microtasks. This PR makes sure we check/use native Promises before es6-promise as a chance to overwrite it.

/cc @IgorMinar @mhevery @jeffbcross @btford @adamdbradley 

I've hit the bug while trying to update zone.js to 5.0.9 in angular2:
- zone.js was using Mutation Observer,
- zone.js monkey patch Mutation Observer but our impl happened to be flawed and [a fix](https://github.com/angular/zone.js/pull/200) has been merged into 5.0.9
- it caused my [zone 0.5.9 PR](https://github.com/angular/angular/pull/5754) to fail on Angular2 - this is also because Travis is using an old version of Chrome. Tests were green on my local machine.

Obviously this PR will most probably (to be confirmed) fix the problem in Chrome.

What is left to be tested is the behavior with a browser that has support for Mutation Observer but not for Promise (should be IE11 only).

TODO:
- [ ] test this PR with IE11


vicb:
I have investigated more and find the root cause - the actual issue has nothing to do with mutation observers but is related to https://github.com/angular/zone.js/commit/d44d699666c4e7c81fc56404af8f44bbaf5612e5

Declaring function inside blocks is not allowed in JS and lead to the following weird behavior in Chrome:
```js
// not valid
if (true) {
  function test() {
    alert('if');
  }
} else {
  function test() {
    alert('else');
  }
}

test(); // "else" !!!

// valid
var test;

if (true) {
  test = function () {
    alert('if');
  }
} else {
  test = function () {
    alert('else');
  }
}

test(); // "if"
```

Because of this behavior, zone.js uses `Symbol()` to generate unique keys in 0.5.9 - in 0.5.8 it would use the version based on plain string because of the above bugs.

It turns out that `Symbol` are flawed on Dartium (Chrome 39.0.2171) which lead to angular.js being broken when using zone 0.5.9.

btford:
Would be good to add an integration test with `es6-promsie` to be sure that we do the right thing when it's present (and maybe track changes to `es6-promise`).

btford:
When using a string key rather than a `Symbol`, we should be making sure that the property we add is not enumerable.

btford:
LGTM– chatted with @vicb in person and decided to create follow-up issues for the points I raised here, since both are sort of "separate" concerns from landing this fix.

btford:
Published a release as 0.5.10

PR #219, opened 2015-12-18
chore: Convert code to TypeScript

mhevery:


lib/zone.ts
```diff
@@ -0,0 +1,4 @@
+import './browser/zone';
+
+export var Zone = global.Zone;
+
```
vicb: I think we should apply the patches here

package.json
```diff
@@ -33,13 +38,15 @@
     "gulp-rename": "^1.2.2",
     "gulp-uglify": "^1.2.0",
     "jasmine-core": "^2.2.0",
-    "karma": "^0.12.31",
```
vicb: remove browserify

test/microtasks.spec.ts
```diff
@@ -1,16 +1,16 @@
 'use strict';
 
 describe('Microtasks', function () {
-  it('should execute microtasks enqueued in the root zone', function(done) {
+  it('should execute microtasks enqueued in the root global.zone', function(done) {
```
vicb: revert

test/setup.ts
```diff
@@ -0,0 +1,9 @@
+(<any>jasmine).DEFAULT_TIMEOUT_INTERVAL = 2000;
```
vicb: could you factorize some code between microtasks / non-microtasks entry points

vicb:
I think the bundle generation should also be updated, right ?

A few comments but the changes seem ok, good to see TS !

Before merging the PR, it would be good to test into an Ng2 app.

lib/zone.ts
mhevery: not applicable.

package.json
mhevery: done

test/microtasks.spec.ts
mhevery: done

test/setup.ts
mhevery: done

PR #220, opened 2015-12-22
docs(readme): fix syntax errors in readme demos

arliang:
fix syntax errors in readme demos


mhevery:
These examples are for the OLD API which has been since changed and the examples do not work on the NEW API. There is a comment at the top of the README (https://github.com/angular/zone.js/blob/master/README.md#deprecated-zonejs-pre-v0515) but we need to make this more explicit.

PR #229, opened 2016-01-05
zone.js does not support timeouts that are passed as strings for eval.

erichooshmand:
zone.js patches to window.setTimeout and window.setInterval do not allow strings containing javascript to be eval'd.

Using zone.js in conjunction with a large legacy codebase requires us to allow eval strings be passed to timeout calls.  Currently, as zone.js expects the value to be a function, strings meant for eval fail.  This edit resolves that issue.

### DIFF
--- a/dist/zone.js
+++ b/dist/zone.js
@@ -406,7 +406,7 @@ function patchSetClearFunction(obj, fnNames) {
         var id, fnRef = fn;
         arguments[0] = function () {
           delete ids[id];
-          return fnRef.apply(this, arguments);
+          return fnRef instanceof Function ? fnRef.apply(this, arguments) : eval.call(global, fnRef);  
         };
         var args = bindArgs(arguments);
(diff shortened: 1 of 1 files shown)


erichooshmand:
I signed it!

Brooooooklyn:
You should commit the Typescript files in `lib` dir, and add some tests, may I do these for you?

island205:
@Brooooooklyn +1

poxrud:
as @Brooooooklyn already pointed out you made changes to the generated distribution file. This file will get overwritted every time it gets rebuilt. Instead you should make your changes to the /lib/patch/functions.ts file. 

Also `eval.call(global, fnRef);` should be changed to `eval.call(this, fnRef);`
so that you are in the proper zone.

erichooshmand:
@Brooooooklyn Thanks, I will close this pull and create one there.  Also, please feel free to create the tests if you have time.

PR #231, opened 2016-01-07
Major Refactoring

mhevery:
@vicb could you have a look?


trevnorris:
Just going to point out a few things that I see (note, this is all coming from the perspective of node).

- Wrapping callbacks adds unnecessary overhead. Currently in node for this type of state propagation we simply either attach the request object (every async operation is a request/handle object) to the event, or we attach it to the C++ class directly.
- The later of the above is important because at times functions are generated in C++ using `FunctionTemplate` before being attached to the request object. This either requires the function to be passed back to JS (which is definitely unnecessary overhead) or ask v8 to create a native API for this case.
- It doesn't address the case where, say, a server receives an incoming connection. At the code level there's no attachment between the server's events and the new connection's events (no server callback above the libuv level fires that is). So that call chain must be manually added, if we want to trace the connection back to the server.
- `[EventListenerTask]` doesn't handle node's specific case where there will be exactly one response at some undetermined time in the future. While this case could be seen as only one case of `[EventListenerTask]`, the important point is it can only happen once so when the request has completed the underlying class can be immediately destructed.

That's the most I could get from this PR. Would have to see a spec for zones before could bring up more.

mhevery:
> Wrapping callbacks adds unnecessary overhead. Currently in node for this type of state propagation we simply either attach the request object (every async operation is a request/handle object) to the event, or we attach it to the C++ class directly.

Yes, you can do that. Instead of calling `Zone.current.wrap` you can just store the zone, then then call the callback using `storedZone.run(callback)`



> The later of the above is important because at times functions are generated in C++ using FunctionTemplate before being attached to the request object. This either requires the function to be passed back to JS (which is definitely unnecessary overhead) or ask v8 to create a native API for this case.

Yes you can do that. Just store the zone and then use `zone.run` to renter. Same as above.



> It doesn't address the case where, say, a server receives an incoming connection. At the code level there's no attachment between the server's events and the new connection's events (no server callback above the libuv level fires that is). So that call chain must be manually added, if we want to trace the connection back to the server.

I don't fully understand what you are asking but my understanding is that this would be covered by the `Task`



> [EventListenerTask] doesn't handle node's specific case where there will be exactly one response at some undetermined time in the future. While this case could be seen as only one case of [EventListenerTask], the important point is it can only happen once so when the request has completed the underlying class can be immediately destructed.

`EventListinerTask` is old thinking, the later SHA only has a `Task` and `TaskSpec` which tells us what kind of time it is, it it should be covered.

trevnorris:
> Yes you can do that. Just store the zone and then use zone.run to renter. Same as above.

So no callback can be run directly from native, unless v8 gives us an API to do so?

> I don't fully understand what you are asking but my understanding is that this would be covered by the Task

Not naturally like most other async operations. The creation of the new connection is triggered by the kernel and doesn't originate from another asynchronous call. Meaning there would be no active "zone" during class construction.

> EventListinerTask is old thinking, the later SHA only has a Task and TaskSpec which tells us what kind of time it is, the use case should be covered.

I'm just going from what documentation was included in the PR here: https://github.com/angular/zone.js/pull/231/files#diff-9c2a67a68b64149ebe4ab75a4eb143f6R96

Qard:
I would expect a connection would just be a task with a parent zone equal to the zone in which createServer() was called.

ofrobots:
Summoning @piscisaureus since he might have some interest in this discussion too.

> Wrapping callbacks adds unnecessary overhead

As @mhevery mentions, this can be avoided, but note that wrapped functions are *much* faster now: http://benediktmeurer.de/2015/12/25/a-new-approach-to-function-prototype-bind/.

> Not naturally like most other async operations. The creation of the new connection is triggered by the kernel and doesn't originate from another asynchronous call. Meaning there would be no active "zone" during class construction.

The connection should be traced back to the `createServer` zone as @Qard mentions.

trevnorris:
@ofrobots 
> The connection should be traced back to the createServer zone as @Qard mentions.

I'm very well aware of this. My point is that there is no global state that the connection can simply draw from, unlike operations like timers, and it's error prone to generate the new instance (both the C++ class and JS handle from a `FunctionTemplate`) then later attach the parent information. This is why I took such pains to propagate this at the native layer for `AsyncWrap`: https://github.com/nodejs/node/blob/v5.5.0/src/tcp_wrap.cc#L139-L140

This is one reason I implemented `AsyncWrap`. To filter as many I/O operations through a single location so they can be easily traced, and add properties to the handle on construction. Though, even so, handling these cases is error prone, and expecting this to be done using a JS API places a substantial burden on node (I'm not going to take time discussing a hypothetical native API until one is presented).

Hopefully next week I'll find time to write up a post-mortem on why domains have failed, and that will give insight into why a similar approach would have a difficult time succeeding.

As for the implementation itself, I'd suggest there is missing functionality/flexibility. In my initial AsyncListener patch every async path was unique, and an error handler could be added/removed in each one. For some example code (not using the exact original API):
```js
// Not using the event emitter style is key to reducing overhead.
const al = new AsyncListener(function errorHandler() { });
// Begin using the new al instance to intercept errors.
useAsyncListener(al);
// Startup a new TCP server and listen for incoming requests.
net.createServer(function(conn) {
  // If connection was received in first five seconds then will have
  // the "al" error handler.
  conn.on('data', function(chunk) {
    // If the incoming data is expected then remove the "al"
    // If this didn't have "al" then it's a noop.
    if (chunk.toString() === 'foobar')
      removeAsyncListener(al);
  });
}).listen(8080);
// Only add error handling to connections made in the first
// five seconds
setTimeout(function() {
  removeAsyncListener(al);
}, 5000);
```
While implementing this the advantage of being able to handle the branching asynchronous requests uniquely became apparent. Without needing to unnecessarily wrap or bind callbacks I was able to handle complex scenarios of nested error handlers with a straight forward API. It was even easy to dynamically add different error handlers to the same async path based on existing conditions (e.g. `addAsyncListener(condition ? al1 : al2)`).

Though even after this landed (and was working quite well) it was deemed that adding more error handling code wasn't a good idea and instead we should focus on hooks that users can use to implement their own error handling mechanisms. So for the last couple years I've been slowly figuring out how to allow `AsyncWrap` the flexibility so users can implement their own error handling but still have the reliability that there will be no gaps in coverage. If anything this is the direction I'd suggest taking.

ofrobots:
> Hopefully next week I'll find time to write up a post-mortem on why domains have failed, and that will give insight into why a similar approach would have a difficult time succeeding.

It would be awesome to have a document on this. While it is widely known that Domains didn't work well, but having a detailed post-mortem on them would be very instructive to anyone who wants to deal with async error handling. Note that zones are not domains.

At the most basic level, the problem that we need to solve in JS (not just Node) is the problem of async context propagation. To do this we need two things:

1. The host environment (e.g. Node) or language to have a concept of a context that is propagated across async boundaries. For Node this is pretty much `AsyncWrap` + support for handling timers and promises.
2. A solution to the [user-space queuing problem](https://docs.google.com/document/d/1tlQ0R6wQFGqCS5KeIw0ddoLbaSYx6aU7vyXOkv-wvlM/edit#heading=h.fy8dcsgxet4l). User-space libraries that do queueing / pooling of callbacks need to propagate context. This can only work if the concept of a context gets standardized, along with API that such user-space libraries can use to propagate it.

This is, at a most fundamental level, what zones are. I would like to start prototyping some of this Node so that this becomes more concrete, specially w.r.t. your performance concerns.

mhevery:
https://github.com/angular/zone.js/pull/231#issuecomment-175903367
> Not naturally like most other async operations. The creation of the new connection is triggered by the kernel and doesn't originate from another asynchronous call. Meaning there would be no active "zone" during class construction.

So I am not familiar with Node, but this would be covered in browser because in order to get a callback of new connection someone had to call `addEventListener` on the connection object. In other words an event had to be added to some event target. Here the `addEventListener` method is patched to capture the domain.

Is this not the case in Node? Isn't there some low level connection listening API? Does node not have a common way of listening on streams?

trevnorris:
@mhevery There's unfortunately a disconnect in the internal mechanics of node. One layer, the one where domains operate, are the JS handles that the users interact with. The other layer, where AsyncWrap operates, are the native class constructors that usually become the `_handle` property of the user facing JS handles.

Attaching the connection to the server on the JS level is easy enough. That information is contained in the logic path it needs to take anyway. It's cases like `TCPWrap`, on the native side, that require special logic to get the necessary information during class construction.

Take for example the `net.createServer()` API. The `listen()` method accepts a string or number. Number means it will open a TCP and string means it will open a Pipe. So the underlying `{TCP,Pipe}Wrap` isn't created until `listen()`, even though the user facing JS API has already been constructed.

This is fundamentally why I couldn't replace all of domain's mechanics with AsyncListener. The resulting API wasn't inferior. Just different. The reason I went this route is because a lot of experimentation demonstrated that it was more reliable placing the mechanics of asynchronous resource tracking closer to where the associated resources are actually created. In v8's case it's also quite a bit more performant.

trevnorris:
@mhevery So to summarize: After many implementation attempts I found the most reliable way was to handle resource tracking blow the JS layer, close to where the actual native handles are located. Then use JS APIs that take advantage of these mechanics by allowing callbacks to run at key locations in the handle's lifetime.

mhevery:
@trevnorris I see. So what does this mean for Zone API? What kinds of changes would you like to see?

trevnorris:
@mhevery While the `run()` syntax is more straight forward, I found the `enter`/`exit` type approach more flexible. For example (using more pseudo-code-ish from old days):
```js
const net = require('net');
const al1 = new AsyncListener('foo');

net.createServer(function(conn) {
  conn.on('data', function(chunk) {
    if (chunk.toString() === 'track me')
      al1.enable();
  });
  conn.on('close', function() {
    al1.disable();  // noop if not being used
  });
}).listen(8080);
```
Since we (node) can propagate the current "parent" async handle in C++ practically for free, we can at least trace the entire async object stack (even though we may be missing the actual stack trace) from any point.

This also allows each async path to be handled uniquely. So I could run `al.disable()` during an asynchronous operation and it would stop listening only on that asynchronous execution branch. Allowing other asynchronous operations that occur on the parent handle to still be tracked.

With this it's easy to create a `run()` interface for users that want to use the simpler interface.

Now there's a discussion on what happens once it's enabled. My implementation eventually came to focusing on state propagation and calling callbacks during the lifetime of the asynchronous handle (sure there's a specific name for this, but very sleep deprived atm). While my original error handling mechanism was more involved (allowing a stack of listeners to catch an error, each one returning whether the error was handled or not, causing the error to bubble to the next listener) basic implementation is as simple as stacking `uncaughtException` handlers and removing them once the callback has completed.

So while a user could implement their own error handler because the callbacks are available, it was also fairly easy to write the "official" error handling mechanism that essentially used the same callbacks.

Data propagation is also important. While AsyncWrap currently uses a simple uid attached to the native constructed class, I understand why this isn't viable in JS. Though the proposed methods and the AsyncWrap's implementation do not conflict and in fact may work well together.

An aside as to why the uid is useful for data propagation, is because creating a `Map` and storing relevant data there showed to have very little overhead. Plus since `AsyncWrap` can notify the user when it's been destructed, the map entry can be removed without fear of memory leak (this is useful because most of node's handles are cleaned up manually via API calls, so we can spare the GC extra work). Also because it allows to transverse and identify handles from native code, which is useful while debugging resource issues.

Sorry, this is a lot of brain dump. Take what you can, but I'll organize these thoughts into bullet points and post back tomorrow.

Qard:
From APM perspective, I'd like to see different notifications when a callback is given and when the async operation is kicked off. That'd cover situations like server.listen() and promise.then().

mhevery:
@Qard 

> From APM perspective, I'd like to see different notifications when a callback is given 

I believe this is covered by `zoneSpec.onIntercept`.

> and when the async operation is kicked off. 

One can wrap the `Task.callback` to achieve this.

I am toying with the idea of a separate `onInvoke` between Zones vs Tasks, but so for have erred on the side of simpler API, as it is not clear to me what would the additional distinction buy me. 



> That'd cover situations like server.listen() and promise.then().

`Promise.then` is covered in the monkey patch mode, which allows us to process promises synchronously.

mhevery:
bee8a60d918fd6d737f8f525cb8d886f69bf06af

PR #192, opened 2015-10-07
chore(saucelabs): update browser list

mlaval:
As suggested by @vicb , the list of browsers which are tested in Sauce Labs should be aligned with the one from Angular.
As a bonus, let's add Edge.

PR #235, opened 2016-01-26
fix(WebSocket): don't patch EventTarget methods twice

tlancina:
On older versions of Chrome, patching via property descriptors isn't available, but unlike Safari, window.EventTarget exists and is already patched, so patching EventTarget methods on WebSocket.prototype again causes them to call the first patched method which then infinitely calls itself.

I didn't provide a test because the tests fail on any older Chrome version (not sure the exact range, tested on 39 and 41).  I don't have access to SauceLabs, but I believe https://github.com/angular/zone.js/pull/192 will cause failure on at least the Android 4.4 case.

This is fairly high priority for Ionic as it seems to crash Android on some devices when using livereload.

### DIFF
--- a/lib/patch/websocket.ts
+++ b/lib/patch/websocket.ts
@@ -3,7 +3,11 @@ import * as utils from '../utils';
 // we have to patch the instance since the proto is non-configurable
 export function apply() {
   var WS = (<any>global).WebSocket;
-  utils.patchEventTargetMethods(WS.prototype);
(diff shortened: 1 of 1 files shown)


tlancina:
I signed it :pencil2:

mhevery:
@vicb Could you merge this and cut a new Zone release. This is blocking Ionic.

tlancina:
Thanks!

PR #236, opened 2016-01-26
fix(utils): add the ability to prevent the default action of onEvents

poxrud:
Add the ability to prevent the default action of onEvent (onclick, onpaste, etc..) from occurring 
by returning false. 

For example, you should be able to stop text from being pasted into a textbox with:
```javascript
textarea.onpaste = function() { 
  return false;
}
```

I discovered this during my investigation of issue #232, which this PR fixes. 
I suspect it will also fix issue https://github.com/meteorhacks/zones/issues/80

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -48,8 +48,17 @@ export function patchProperty(obj, prop) {
     }
 
     if (typeof fn === 'function') {
-      this[_prop] = fn;
-      this.addEventListener(eventName, fn, false);
+      
+      function wrapFn(event) {
+        var result;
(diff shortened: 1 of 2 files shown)


poxrud:
@mhevery 
This PR is now incompatible with POST-v0.6.0., however I've verified that this issue still exists. 
Should I submit a new PR for this?

mhevery:
Yes please!

mhevery:
Any update on a rebase?

mhevery:
Actually, it applies cleanly, and tests, pass so I am merging it in.

PR #238, opened 2016-01-27
0.5.11

vicb:


vicb:
@btford could you please take a look and release if ok.

Note: angular2 needs [to be patched](https://github.com/angular/angular/pull/6086) before zone 0.5.11 is used

/cc @mhevery

btford:
published

PR #241, opened 2016-01-29
fix(utils): access new XMLHttpRequest().responseBlob error

Brooooooklyn:
https://bugs.webkit.org/show_bug.cgi?id=44721

In some browsers that use webkit, It would fail when patchClass function try to access responseBlob in XMLHttpRequest instance.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -158,7 +158,13 @@ export function patchClass(className) {
   var prop;
   for (prop in instance) {
     (function (prop) {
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
@vicb review it please, it's would cause application crush in some android browser.

Brooooooklyn:
@mhevery

Brooooooklyn:
@jeffbcross

PR #242, opened 2016-01-31
chore(typings): publish TS typings in the npm distro

### LINKED ISSUE #237: Distribute .d.ts files from typescript compilation
Users should no longer fetch zone typings from DefinitelyTyped, now that it is written in TS we can distribute typings along with the .js files like rxjs and angular2 do.
This is needed so that user code (including angular2) can rely on moduleResolution=node to find the zone typings, rather than pollute with an ambient module.

alexeagle:
Fixes #237

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "zone.js",
   "version": "0.5.11",
   "description": "Zones for JavaScript",
-  "main": "lib/zone.js",
+  "main": "build/lib/zone.js",
+  "typings": "build/lib/core.d.ts",
   "files": [
     "lib",
     "dist"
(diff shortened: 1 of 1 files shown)


alexeagle:
cc @mhevery 
This is the easy way: just run tsc before publishing to npm, and the build directory will be included in the distro.
Right now, if you `require('zone.js')` it is broken, because our `"main":"lib/zone.js"` does not exist.
After this change, you can `require("zone.js")` but more importantly you can `import {Zone} from 'zone.js';` and Typescript will get the type of `Zone`, without user needing to `typings install zone.js` manually.

This is needed for https://github.com/angular/angular/pull/6267

mhevery:
LGTM

PR #253, opened 2016-02-11
Use typings; tsd is deprecated

mikekidder:
Was going to push earlier, but saw you had updated tsd.  :)

package.json - removed tsd, adds typings.... the rest is just from alpha sorting


mikekidder:
CLA, I signed it!

package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.5.14",
   "description": "Zones for JavaScript",
   "main": "build/lib/zone.js",
-  "typings": "build/lib/core.d.ts",
```
alexeagle: Don't remove this line - this is an output of the package. It allows users to
`import {Zone} from 'zone.js';`
and the typescript type-checker can find it.

alexeagle:
LGTM after that fix

package.json
mikekidder: Ah, good to know. Wasn't sure if that was a tsd leftover.

mikekidder:
Closing as this looks to be covered in 0.6.1

Issue #258, opened 2016-02-17
Add a test for Universal


jteplitz:
We should have a test that checks that zones load in node and that everything is patched properly in node. Unfortunately, AFAIK karma does not run in node so we'll need to add a separate test runner for node.

PR #259, opened 2016-02-17
fix(WebWorker): Patch WebSockets and XMLHttpRequest in WebWorker

### LINKED ISSUE #249: WebSockets are not patched in WebWorker
WebSocket patching has gotten mixed in with property description patching, which does not run on a WebWorker. As such WebSockets aren't patched in a Worker even though the API is available.

jteplitz602:
I manually verified that this works in node, but we should have a way to automate this so we don't introduce the same bug in the future (see #258)

Closes #249

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -5,11 +5,15 @@ module.exports = function (config) {
     basePath: '',
     files: [
       'test/browser_entry_point.ts',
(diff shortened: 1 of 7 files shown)


jeffbcross:
LGTM, waiting on Travis

PR #275, opened 2016-03-15
feat(zonespec): add a spec for asynchronous tests

juliemr:
This spec is constructed with a done callback and a fail callback,
and waits until all asynchronous tasks are completed before
exiting.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -98,6 +98,10 @@ gulp.task('build/wtf.min.js', function(cb) {
   return generateBrowserScript('./lib/zone-spec/wtf.ts', 'wtf.min.js', true, cb);
 });
 
+gulp.task('build/async-test.js', function(cb) {
+  return generateBrowserScript('./lib/zone-spec/async-test.ts', 'async-test.js', false, cb);
+});
+
 gulp.task('build', [
   'build/zone.js',
   'build/zone.js.d.ts',
@@ -108,7 +112,8 @@ gulp.task('build', [
   'build/long-stack-trace-zone.js',
   'build/long-stack-trace-zone.min.js',
   'build/wtf.js',
-  'build/wtf.min.js'
+  'build/wtf.min.js',
+  'build/async-test.js'
 ]);
 
 
--- a/lib/zone-spec/async-test.ts
+++ b/lib/zone-spec/async-test.ts
@@ -1,0 +1,96 @@
+(function() {
+  class AsyncTestZoneSpec implements ZoneSpec {
+    _finishCallback: Function;
+    _failCallback: Function;
+    _pendingMicroTasks: boolean = false;
+    _pendingMacroTasks: boolean = false;
+    _alreadyErrored: boolean = false;
+    runZone = Zone.current;
+
+    constructor(finishCallback: Function, failCallback: Function, namePrefix: string) {
+      this._finishCallback = finishCallback;
+      this._failCallback = failCallback;
+      this.name = 'asyncTestZone for ' + namePrefix;
+    }
+
+    _finishCallbackIfDone() {
+      if (!(this._pendingMicroTasks || this._pendingMacroTasks)) {
+        // We do this because we would like to catch unhandled rejected promises.
+        // To do this quickly when there are native promises, we must run using an unwrapped
+        // promise implementation.
+        var symbol = (<any>Zone).__symbol__;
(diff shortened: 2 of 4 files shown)


juliemr:
WIP - still needs tests.

juliemr:
Tests are now added! Please take a look. cc @mhevery @vikerman

juliemr:
This will be updated on top of https://github.com/angular/zone.js/pull/275

dist/async-test.js
```diff
@@ -0,0 +1,137 @@
+/******/ (function(modules) { // webpackBootstrap
```
mhevery: please revert the dist folder. I will rebuild when we are ready to cut release

karma.conf.js
```diff
@@ -51,7 +51,7 @@ module.exports = function (config) {
 
     logLevel: config.LOG_INFO,
 
-    browsers: ['Firefox'],
+    browsers: ['Chrome'],
```
mhevery: was this meant to be here? Maybe a separate SHA?

gulpfile.js
```diff
@@ -108,7 +112,8 @@ gulp.task('build', [
   'build/long-stack-trace-zone.js',
   'build/long-stack-trace-zone.min.js',
   'build/wtf.js',
-  'build/wtf.min.js'
+  'build/wtf.min.js',
+  'build/async-test.js'
 ]);
```
vikerman: I don't know the full details but just looking at the other specs - Do we also need a async-test.min.js?

lib/zone-spec/async-test.ts
```diff
@@ -0,0 +1,96 @@
...
+              }
+            }, 0);
+          });
+        }
+      }
+    }
+
+    // ZoneSpec implementation below.
+
+    name: string = 'asyncTestZone';
```
vikerman: Isn't the name already set in the constructor?

lib/zone-spec/async-test.ts
```diff
@@ -0,0 +1,96 @@
+(function() {
+  class AsyncTestZoneSpec implements ZoneSpec {
+    _finishCallback: Function;
+    _failCallback: Function;
+    _pendingMicroTasks: boolean = false;
+    _pendingMacroTasks: boolean = false;
+    _pendingEventTasks: boolean = false;
+    _alreadyErrored = false;
```
vikerman: _alreadyErrored: boolean = false ?

test/async-test.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    var atz = Zone.current.fork(testZoneSpec);
+
+    atz.run(function() {
+      setTimeout(() => {
+        finished = true;
+      });
+    });
+  });
+
+  it('should call finish after a setTimeout is done', (done) => {
```
vikerman: Repeated?

test/async-test.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    atz.run(function() {
+      // Promise rejection does not cause onHandleError to trigger
+      // until the next Macro task, so that there's a chance
+      // for handlers to be added.
+      Promise.reject('my reason');
+    });
+
+  });
+});
```
vikerman: The zone spec has explicit logic to test whether micro, macro and event tasks are empty. So should we have tests that mix the task types and wait for all of them to drain?

dist/async-test.js
juliemr: done.

gulpfile.js
juliemr: Nope, I didn't add the build steps to make a min version, since this will only be for testing.

karma.conf.js
juliemr: Oops, reverted.

lib/zone-spec/async-test.ts
juliemr: done.

lib/zone-spec/async-test.ts
juliemr: Yes, moved this to just the declaration.

test/async-test.spec.ts
juliemr: Yup, removed.

test/async-test.spec.ts
juliemr: Done! Added some. Also, that uncovered some bugs with the event task searching, so I fixed it up. Basically, we never want to wait on event tasks.

PR #277, opened 2016-03-17
fix: allow clearTimeout() with undefined/null

### LINKED ISSUE #276: clearTimeout with null/undefined throws exception
When using `clearTimeout` with an `undefined` or `null` value on a page with Zone.js, the following exception is thrown:

```
zone.js:115 Uncaught TypeError: Cannot read property 'zone' of undefined
```

Plunker: https://plnkr.co/edit/dDEXwEVm3hbmleKU179b?p=preview

This should not happen, as [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/clearTimeout): "Passing an invalid ID to clearTimeout does not have any effect (and doesn't throw an exception)."

teropa:
Closes #276

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -84,7 +84,8 @@ function patchTimer(
 
   var clearNative = patchMethod(window, cancelName, () => function(self: any, args: any[]) {
     var task: Task = args[0];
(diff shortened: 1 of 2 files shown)


mhevery:
Closed by #279

PR #281, opened 2016-03-18
fix(zonespec): Do not crash on error if last task had no data

sjelin:
I have no  idea what I'm doing in this codebase so hopefully I didn't break anything


sjelin:
@mhevery are you the person who should review this?  It's not a very big change - a lot of the changed files are generated files.  I only edited lib/zone-spec/long-stack-trace.ts

sjelin:
Changes in `dist` reverted - thanks to @juliemr for the information

PR #282, opened 2016-03-18
chore(package.json): use browser field

gdi2290:
see https://github.com/defunctzombie/package-browser-field-spec

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "zone.js",
   "version": "0.6.3",
   "description": "Zones for JavaScript",
-  "main": "dist/zone.js",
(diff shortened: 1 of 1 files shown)


gdi2290:
also consider `zone.js/browser` and `zone.js/node`. you can expose js and d.ts files at the root to requires dist/

`./node.js`
```
require('dist/zone-node.js');
```
`./node.d.ts`
```
import 'dist/zone';
```



`./browser.js`
```
require('dist/zone.js');
```
`./browser.d.ts`
```
import 'dist/zone';
```


then you can also do `zone.js/browser` and `zone.js/node` with types

PR #284, opened 2016-03-18
fix(browser/utils): calling removeEventListener twice with the same a…

### LINKED ISSUE #283: Calling `EventTarget.removeEventListener` twice causes errors
Consider the following code snippet:

```js
var elem = document.body;
var listener = function (event) {};
elem.addEventListener('click', listener, false);
elem.removeEventListener('click', listener, false);
elem.removeEventListener('click', listener, false);
```

According to the [DOM spec](https://dom.spec.whatwg.org/#dom-eventtarget-removeeventlistener), the second `removeEventListener` should do nothing.  However, with `zone.js` loaded it produces the following error message:

```
	Error: Task does not support cancellation, or is already canceled.
	    at ZoneDelegate.cancelTask (/usr/local/google/home/sjelin/zone.js/test/browser_entry_point.ts:382:24)
	    at Zone.cancelTask (/usr/local/google/home/sjelin/zone.js/test/browser_entry_point.ts:292:45)
	    at zoneAwareRemoveEventListener (/usr/local/google/home/sjelin/zone.js/test/browser_entry_point.ts:970:25)
	    at HTMLBodyElement.removeEventListener (eval at createNamedFn (/usr/local/google/home/sjelin/zone.js/test/browser_entry_point.ts:1051:18), <anonymous>:3:46)
	    at Object.<anonymous> (/usr/local/google/home/sjelin/zone.js/test/browser_entry_point.ts:2354:19)
```

Basically, the problem is that [the patched `removeEventListener`](https://github.com/angular/zone.js/blob/74b4b97c1449819d7a96ce6b60ab860c20e7867f/lib/browser/utils.ts#L190) will [cancel the task](https://github.com/angular/zone.js/blob/74b4b97c1449819d7a96ce6b60ab860c20e7867f/lib/browser/utils.ts#L200), but never removes it from the list of registered tasks.  So when the second `removeEventListener` happens it tries to cancel the task again, leading to the problem.  I suggest we start removing it from the list of registered tasks, unless there is a reason it has to stay there.

sjelin:
…rgs should not cause errors

Closes #283

### DIFF
--- a/lib/browser/utils.ts
+++ b/lib/browser/utils.ts
@@ -195,7 +195,7 @@ function zoneAwareRemoveEventListener(self: any, args: any[]) {
   // - When `addEventListener` is called on the global context in strict mode, `this` is undefined
   // see https://github.com/angular/zone.js/issues/190
   var target = self || _global;
-  var eventTask = findExistingRegisteredTask(target, handler, eventName, useCapturing, false);
+  var eventTask = findExistingRegisteredTask(target, handler, eventName, useCapturing, true);
   if (eventTask) {
     eventTask.zone.cancelTask(eventTask);
   } else {
--- a/test/browser/element.spec.ts
+++ b/test/browser/element.spec.ts
@@ -257,6 +257,13 @@ describe('element', function () {
       button.click();
       expect(log).toEqual('');
     });
+
+    it('should be able to deregister the same event twice', function() {
+      var listener = (event) => {};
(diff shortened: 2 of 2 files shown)


sjelin:
@mhevery

arsenys:
Just checked, this fixes the all the problems I've described in the comments for issue #283

PR #291, opened 2016-03-21
chore: initial node fork

gdi2290:
nothing here until we eventually patch everything


mhevery:
I am going to close this one, since @juliemr is working on proper Node.js support now.

PR #294, opened 2016-03-22
feat(zonespec): add a spec for synchronous tests

vikerman:
Will throw an exception if any asynchornous microtask or macrotask is scheduled. Allows event tasks to be scheduled.


juliemr:
LGTM.

I spent a while thinking about whether we want to fail if a microtask is scheduled, or if that's too restrictive. But I think I like this commit the way it is, because you're ensuring that everything written in the test really has already run by the time we get to the end of the function.

mhevery:
LGTM

PR #295, opened 2016-03-22
Update utils.ts

alexeagle:
Suppress closure compiler warnings about unknown 'process' variable. Needed to compile angular2 apps with closure compiler.


alexeagle:
This also suppresses an error about the `Zone` symbol being unknown.

PR #298, opened 2016-03-24
chore: Fix type definitions

### LINKED ISSUE #297: Exported typings are not a module
Exported external package typings file '**/node_modules/zone.js/dist/zone.js.d.ts' is not a module. Please contact the package author to update the package definition.

It appears that the typescript build is not outputting a module in it's definitions output.

park9140:
Fixes exported type definitions so that you can import from typescript using ```import { Zone } from 'zone.js'``` and not have the compiler complain about it not being a module.

Resolves #297, still needs a resolution to access the correct ZoneSpec via import without declaring a module manually.

### DIFF
--- a/lib/zone-spec/async-test.ts
+++ b/lib/zone-spec/async-test.ts
@@ -1,3 +1,5 @@
+import { Zone, ZoneSpec, Task, ZoneDelegate, HasTaskState } from '../zone';
+
 (function() {
   class AsyncTestZoneSpec implements ZoneSpec {
     _finishCallback: Function;
--- a/lib/zone-spec/long-stack-trace.ts
+++ b/lib/zone-spec/long-stack-trace.ts
@@ -1,3 +1,4 @@
+import { Zone, ZoneDelegate, Task, ZoneSpec } from '../zone';
 'use strict';
 (function() {
   const NEWLINE = '\n';
--- a/lib/zone-spec/sync-test.ts
+++ b/lib/zone-spec/sync-test.ts
@@ -1,3 +1,5 @@
+import { Zone, ZoneSpec, Task, ZoneDelegate, MicroTask } from '../zone';
+
 (function() {
   class SyncTestZoneSpec implements ZoneSpec {
     runZone = Zone.current;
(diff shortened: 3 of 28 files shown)


gdi2290:
you could also add this snippet to the zone.ts file
```typescript
// Ensure this is treated as a module.
declare module "zone.js" {}
```

park9140:
@gdi2290 I considered that. I just figured it was better to push the project toward a correct implementation, rather than the current clumsy mish mosh of ambient dependency, and external modules.

gdi2290:
no worries, your version is correct and preferred :+1:

lib/zone-spec/async-test.ts
```diff
@@ -1,3 +1,5 @@
+import { Zone, ZoneSpec, Task, ZoneDelegate, HasTaskState } from '../zone';
```
mhevery: We can not add this code here. If we do then Zones will be included in the `async-test.js` bundle as well as `zone.js` bundle. 

Zone's is meant as a polyfil, and so it should be ambiently available.

lib/zone-spec/long-stack-trace.ts
```diff
@@ -1,3 +1,4 @@
+import { Zone, ZoneDelegate, Task, ZoneSpec } from '../zone';
```
mhevery: same here

lib/zone-spec/sync-test.ts
```diff
@@ -1,3 +1,5 @@
+import { Zone, ZoneSpec, Task, ZoneDelegate, MicroTask } from '../zone';
+
```
mhevery: same here

lib/zone-spec/wtf.ts
```diff
@@ -1,3 +1,5 @@
+import { Zone, ZoneSpec, Task, ZoneDelegate } from '../zone';
+
```
mhevery: same here

lib/zone.ts
```diff
@@ -119,7 +119,7 @@
  * zones are children of the root zone.
  *
  */
-interface Zone {
+export interface Zone {
```
mhevery: This will prevent the Zone from being imported ambiently. Sorry we can't do it that way.

mhevery:
Thanks for you hard work, but we can't accept this PR. The issue is that Zone needs to be treated as if it is part of platform. ie. it needs to have ambient definition. Your change makes it so that it must be imported explicitly.

PR #300, opened 2016-03-25
WIP - treat XHRs as Macrotasks

juliemr:


lib/zone-spec/async-test.ts
```diff
@@ -20,13 +20,16 @@
...
         var symbol = (<any>Zone).__symbol__;
         var NativePromise: typeof Promise = <any>window[symbol('Promise')];
-        if (NativePromise) {
-          NativePromise.resolve(true)[symbol('then')](() => {
-            if (!this._alreadyErrored) {
-              this.runZone.run(this._finishCallback);
-            }
-          });
-        } else {
+        // TODO(julie): This isn't good enough if we have an error thrown within an event callback.
```
juliemr: @mhevery thoughts on if we can be better here?

lib/browser/browser.ts
```diff
@@ -34,6 +34,78 @@ patchClass('FileReader');
 propertyPatch();
 registerElementPatch(_global);
 
+// Treat XMLHTTPRequest as a macrotask.
+patchXHR(_global);
```
mhevery: I wonder if it may be better to move this XHR patching to `XMLHttpRequest` since we already doing property patching there. Take a look and let me know what you think.

juliemr: Not entirely sure what you mean here - instead of patching XMLHttpRequest with `patchEventTargetMethods` like everything else, we could special-case it, but that wouldn't work for browsers that have a global `EventTarget` if I understand correctly (https://github.com/angular/zone.js/blob/9b3e77988a54824ee97a3277cc192ab442cdb788/lib/browser/event-target.ts).

We could also have a special case for XMLHttpRequest in `patchOnProperties` - I can try that out if it sounds good to you.

lib/browser/browser.ts
```diff
@@ -34,6 +34,78 @@ patchClass('FileReader');
...
+    return task;
+  }
+
+  function placeholderCallback() {
+  }
+
+  function clearTask(task: Task) {
+    var data = <XHROptions>task.data;
+    // Note - ideally, we would call data.target.removeEventListener here, but it's too late
+    // to prevent it from firing. So instead, we store info for the event listener.
```
mhevery: I think you may need to cancel the task here?

lib/zone-spec/async-test.ts
mhevery: chatted in person. setTimeout is fine.

test/browser/XMLHttpRequest.spec.ts
```diff
@@ -44,6 +84,32 @@ describe('XMLHttpRequest', function () {
     });
   }));
 
+  it('should allow canceling of an XMLHttpRequest', function(done) {
+    var spy = jasmine.createSpy('spy');
+    var req;
+
+    testZone.run(function() {
+    req = new XMLHttpRequest();
+    req.onreadystatechange = function() {
```
mhevery: you should also verify that there are no outstanding tasks.

mhevery:
LGTM

lib/browser/browser.ts
mhevery: You are right, let's keep it here.

test/browser/XMLHttpRequest.spec.ts
juliemr: done.

lib/zone-spec/async-test.ts
juliemr: Cleaned up the code and removed old comments.

juliemr:
I updated, squashed, and got all tests passing!

@mhevery would you like to take another look?

mhevery:
Looks good, please merge.

juliemr:
Merged as fd39f97e20fe49028590b4f3f94f13a2249fdc46

Issue #290, opened 2016-03-21
"More tasks executed then were scheduled" error when using zone.js and jquery.layout.


NikGovorov:
Please try to resize a window or hover on any resize pane:
http://plnkr.co/edit/coyaquV0zuwwVGLGAGUE?p=preview

born2net:
+1 seeing same issues...

born2net:
same issues in 0.12
```
More tasks executed then were scheduled.
```

NikGovorov:
Confirming, same issue in  0.6.6: http://plnkr.co/edit/XLXdc1EoEzOE7O4VoPQO?p=preview

NikGovorov:
@mhevery, Could you please re-open the issue?

ghaiat:
+1
Same issue here

fxck:
yes, it didn't help

born2net:
this is a big showstopper for me...
I can't release ng 0.12 Kitchen sink: http://ng2.javascriptninja.io
due to this bug...

ng team, PLEASE remove the  Throw Exceptions in zone.js 

Regards,

Sean

born2net:
please re-open

arsenys:
I have a same bug, "More tasks executed then were scheduled." started to appear in my application that uses [Semantic-UI](http://semantic-ui.com/) for many components. Please re-open.

MaikeMota:
Same problem here. The "More tasks executed then were scheduled" message always appear when has interation with semantic-ui elements.

![image](https://cloud.githubusercontent.com/assets/4526602/14023714/6c15ab80-f1c4-11e5-862c-4f645b6ffccd.png)

![image](https://cloud.githubusercontent.com/assets/4526602/14023734/870c8b48-f1c4-11e5-8217-e7ebe3e1dd01.png)

I'm using 2.0.0-beta.12 version.

mhevery:
This will show up in beta.13

mhevery:
OK, it seems like there as another issue with the same error message. Could someone please file a new issue with a plunker demonstrating the problem.

frankwallis:
@mhevery - I think it is the same issue, this is using zone.js@0.6.6: 
http://plnkr.co/edit/XrjNURLeN5bhPrkXnohG?p=preview

SelcukTekin:
Also want to leave my input here that we are experiencing the same problem which is incredibly easy to reproduce. Just create a function that creates a timeout every 10 seconds that triggers itself. 

[...]

PR #301, opened 2016-03-25
fix(setTimeout): allow clearTimeout to be called in setTimeout callback

frankwallis:
This is a fix for #290 - "More tasks executed than were scheduled" error.

The change is to set task.cancelFn  = null *before* invoking a task for non-periodic tasks. As it is no longer possible to cancel the task which is just about to be invoked, I think this should be ok. I also updated the setTimeout tests to create new zones to run the tests in, as initially the test was incorrectly passing due to the task counts not being 0 to begin with.


frankwallis:
I have signed the CLA

PR #307, opened 2016-04-03
use const and let instead of var

Brooooooklyn:
Use const and let instead of var
And use semicolon in interface


Brooooooklyn:
@mhevery I'm wondering if use `let` and `const` is a better way to write typescript, and I remove the `var` in lib files.

And also , I fix some style in codes such as `==` and ',' in interface and type

mhevery:
Yes `let` and `const` are better. Old habits die hard.

PR #308, opened 2016-04-04
chore: 6.8 build and release

juliemr:


dist/async-test.js
```diff
@@ -59,29 +59,13 @@
...
-	                // To do this quickly when there are native promises, we must run using an unwrapped
-	                // promise implementation.
-	                var symbol = Zone.__symbol__;
-	                var NativePromise = window[symbol('Promise')];
-	                if (NativePromise) {
-	                    NativePromise.resolve(true)[symbol('then')](function () {
-	                        if (!_this._alreadyErrored) {
-	                            _this.runZone.run(_this._finishCallback);
+	                this.runZone.run(function () {
+	                    setTimeout(function () {
```
mhevery: do we want to run the setTimeout in the current zone? (I don't know the answer, just thinking)

dist/zone.js
```diff
@@ -76,6 +76,60 @@
...
+	        };
+	        return zone.scheduleMacroTask('XMLHttpRequest.send', placeholderCallback, options, scheduleTask, clearTask);
+	    }; });
+	    var clearNative = utils_1.patchMethod(window.XMLHttpRequest.prototype, 'abort', function (delegate) { return function (self, args) {
+	        var task = findPendingTask(self);
+	        if (task && typeof task.type == 'string') {
+	            task.zone.cancelTask(task);
+	        }
+	        else {
+	            throw new Error('tried to abort an XHR which has not yet been sent');
```
mhevery: Do we want to throw errors, which browser would have not thrown? I think zone should be un-observable, in the sense that it should not throw extra errors,

mhevery:
Looks reasonable.

dist/async-test.js
juliemr: I don't think so, because we don't want to trace our own setTimeout, but I also am not sure.

dist/zone.js
juliemr: Done - I made a separate `fix` commit to do this in lib, and then re-ran the build.

juliemr:
Merged as 8827e1e9f782cc62bbc379ab9de0bfc6f72823ff and 54aec092e4c3ddd394bee95aa3c0fe426050188d

PR #317, opened 2016-04-12
chore: lots of cleanup in preparation for a node.js entry point

juliemr:
Move common files and tests into `common` folders.
Rename test util to `test-util` for clarity.
Remove unused `brick` test.


lib/browser/define-property.ts
```diff
@@ -1,5 +1,5 @@
-import {zoneSymbol} from "./utils";
-
+import {zoneSymbol} from "../common/utils";
+// TODO(julie): does this belong in common?
```
mhevery: yes, I think `common` is the right place

mhevery:
LGTM

lib/browser/define-property.ts
juliemr: I'll investigate more and do this in a follow-up.

PR #318, opened 2016-04-12
TEST - see if we can remove propertyPatch for all browsers

juliemr:
Please do not review yet, just getting this to run on Travis.


juliemr:
Cool, we can't.

PR #321, opened 2016-04-12
test(node): add task to run a subset of tests on node

juliemr:
`npm run test-node` or `gulp test/node` will now run tests
on node.js. These are also run on TravisCI. Clean up gulp
build.


test/node_entry_point.ts
```diff
@@ -0,0 +1,13 @@
+// Setup tests for Zone without microtask support
+import '../lib/zone';
+import '../lib/node/node';
+import '../lib/zone-spec/long-stack-trace';
+
+// Setup test environment
+import './test-env-setup';
+
+// List all tests here:
+import './common/microtasks.spec';
```
mhevery: can we share this list with we the browser somehow? That way a test file added to the common would be added to both together?

mhevery:
LGTM. Could you pull common test into a common file?

juliemr:
Done - merged as 89530fe79a028dd6b854e2fb02e9e9b70424fab0

Issue #323, opened 2016-04-14
Cordova pause event error from a null target


adamdbradley:
Cordova fires a `pause` event when the app is put into the background, but this error is thrown from within `patchViaCapturingAllTheEvents()`:
![image](https://cloud.githubusercontent.com/assets/452425/14514195/2745ffaa-01b4-11e6-80d5-89f9d07eca25.png)

It looks as though `event.target` is `null`:
![image](https://cloud.githubusercontent.com/assets/452425/14514226/5e29da1e-01b4-11e6-92b4-64c6eca56fe4.png)

This only happens on Cordova's `pause` event, but is not a problem with the others:
https://cordova.apache.org/docs/en/latest/cordova/events/events.html#pause

mhevery:
@juliemr could you look at this?

juliemr:
This is the easy fix: https://github.com/angular/zone.js/pull/327

But I haven't found the magical combination of steps necessary to reproduce the bug outside of cordova. @adamdbradley do you have a script to create a targetless event that I can add to our tests without setting up cordova on our CI?

This also doesn't address Zones possibly not properly including or wrapping Cordova's custom events - it just makes sure that it won't outright fail.

adamdbradley:
@juliemr I dug through cordova and it looks like the `pause` event ultimately [calls fireDocumentEvent](https://github.com/apache/cordova-js/blob/c517ca811b4948b630e0b74dbae6c9637939da24/src/cordova.js#L138)

Inside fireDocumentEvent it then uses the [createEvent method](https://github.com/apache/cordova-js/blob/c517ca811b4948b630e0b74dbae6c9637939da24/src/cordova.js#L86)

Next, it passes the created `evt` to [documentEventHandlers[type].fire(evt)](https://github.com/apache/cordova-js/blob/c517ca811b4948b630e0b74dbae6c9637939da24/src/cordova.js#L141) 

Which is cordova's [Channel.prototype.fire](https://github.com/apache/cordova-js/blob/24ab6855470f2dc0662624b597c98585e56a1666/src/common/channel.js#L198)

The `fireArgs` variable from within `Channel.prototype.fire` has the `null` `target`: 
![image](https://cloud.githubusercontent.com/assets/452425/14592783/0da37fa0-04e8-11e6-942b-2d6c80d333e7.png)

[...]

PR #327, opened 2016-04-14
fix(property-descriptor): do not fail for events without targets

juliemr:
See #323

### DIFF
--- a/lib/browser/property-descriptor.ts
+++ b/lib/browser/property-descriptor.ts
@@ -66,8 +66,12 @@ function patchViaCapturingAllTheEvents() {
     const property = eventNames[i];
     const onproperty = 'on' + property;
     document.addEventListener(property, function (event) {
-      let elt = <Node>event.target, bound;
-      const source = elt.constructor['name'] + '.' + onproperty;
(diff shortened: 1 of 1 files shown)


lib/browser/property-descriptor.ts
```diff
@@ -66,8 +66,10 @@ function patchViaCapturingAllTheEvents() {
     const property = eventNames[i];
     const onproperty = 'on' + property;
     document.addEventListener(property, function (event) {
-      let elt = <Node>event.target, bound;
-      const source = elt.constructor['name'] + '.' + onproperty;
+      let elt = <Node>event.target, bound, source;
+      if (elt) {
+        source = elt.constructor['name'] + '.' + onproperty;
+      }
```
mhevery: could we add an else branch which would at assume `"unknown"` constructor name and then add the `onproperty`? It sucks that the fix now makes source `undefined`, which is hard for debugging.

mhevery:
Please take care of the comment, then then it LGTM, to merge.

lib/browser/property-descriptor.ts
juliemr: Done.

Issue #328, opened 2016-04-15
Accessing XMLHttpRequest#responseBlob breaks older browsers


tkazec:
`patchClass()` is causing Zone.js to break in older browsers like Android 4.1 – see angular/angular#8000.

Hacky fix: https://github.com/Buzzvil/zone.js/commit/1c3f040cba8a1745829c0f8143620564c250b895

Brooooooklyn:
same issue with you

It's a very old webkit bug https://bugs.webkit.org/show_bug.cgi?id=44721

tkazec:
As far as I know, this was part of a spec, then removed. Some older browsers have a buggy implementation that causes DOM Exception 11 to be thrown when touched.

Since Zone.js is used in Angular 2, and Angular 2 advertises Android 4.1 support, this needs a fix ASAP.

tkazec:
All good now, thanks.

PR #329, opened 2016-04-15
fix(XHR.responseBlob): don't access XHR.responseBlob on old android webkit

aicest:


Brooooooklyn:
@mhevery  fix https://github.com/angular/zone.js/issues/328

mhevery:
Could you add a test for this?

tkazec:
Can we get a new release?

PR #330, opened 2016-04-18
feat: Add a zone spec for fake async test zone.

vikerman:
This implementation is forked from the fake async from Angular 2 and adds the following

1) Implements own scheduler queue that doesn't depend on the jasmine implementation of a scheduler queue.

2) Handles errors in timer/microtask callbacks better so that pending timers are removed properly on error (No need for a clearPendingTimers at the end of a test handling errors). The queue processing is stopped after the first error and can be restarted.

3) Throw an exception if a XMLHttpRequest is used in a fakeAsync test - Useful for avoiding cases for ex. someone uses the online template compiler in a fakeAsync test but forgot to use the template cache and ends up doing a XHR in the test.

### DIFF
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -1,0 +1,256 @@
+(function() {
+  interface ScheduledFunction {
+    endTime: number;
+    id: number,
+    func: Function;
+    args: any[];
+    delay: number;
+  }
+  
+  class Scheduler {
+    // Next scheduler id.
+    public nextId: number = 0;
+    
+    // Scheduler queue with the tuple of end time and callback function - sorted by end time.
+    private _schedulerQueue: ScheduledFunction[] = [];
+    // Current simulated time in millis.
+    private _currentTime: number = 0;
+    
+    constructor() {}
+    
+    scheduleFunction(cb: Function, delay: number, args: any[] = [], id: number = -1) : number {
+      let currentId: number = id < 0 ? this.nextId++ : id;
+      let endTime = this._currentTime + delay;
+            
+      // Insert so that scheduler queue remains sorted by end time.
+      let newEntry: ScheduledFunction = {
+        endTime: endTime,
+        id: currentId,
+        func: cb,
+        args: args,
+        delay: delay
+      }
+      let i = 0;
+      for (; i < this._schedulerQueue.length; i++) {
+        let currentEntry = this._schedulerQueue[i];
+        if (newEntry.endTime < currentEntry.endTime) {
+          break;
+        }
+      }
+      this._schedulerQueue.splice(i, 0, newEntry);
+      return currentId;
+    }
+    
+    removeScheduledFunctionWithId(id: number): void {
+      for (let i = 0; i < this._schedulerQueue.length; i++) {
+        if (this._schedulerQueue[i].id == id) {
+          this._schedulerQueue.splice(i, 1);
+          break;
+        }
+      }
+    }
+    
+    tick(millis: number = 0): void {
+      this._currentTime += millis;
+      while (this._schedulerQueue.length > 0) {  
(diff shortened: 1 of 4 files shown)


juliemr:
The Travis error seems to be legitimate.

test/zone-spec/fake-async-test.spec.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    it('should process microtasks before timers', () => {
+      fakeAsyncTestZone.run(() => {
+        let log = [];
+
+        Promise.resolve(null).then((_) => log.push('microtask'));
+
+        setTimeout(() => log.push('timer'), 9);
+
+        let id = setInterval(() => log.push('periodic timer'), 10);
```
juliemr: is is unused

test/zone-spec/fake-async-test.spec.ts
```diff
@@ -0,0 +1,307 @@
...
+
+  describe('timers', () => {
+    it('should run queued zero duration timer on zero tick', () => {
+      fakeAsyncTestZone.run(() => {
+        let ran = false;
+        setTimeout(() => { ran = true }, 0);
+
+        expect(ran).toEqual(false);
+
+        testZoneSpec.tick();
```
juliemr: Question: is `tick()` just an alias for `tick(0)` - and what was the previous behavior? Or should it try to clear out any non-periodic timers?

lib/zone-spec/fake-async-test.ts
```diff
@@ -0,0 +1,260 @@
...
+      while (this._schedulerQueue.length > 0) {  
+        let current = this._schedulerQueue[0];        
+        if (this._currentTime < current.endTime) {
+          // Done processing the queue since it's sorted by endTime.
+          break;
+        } else {
+          // Time to run scheduled function. Remove it from the head of queue.
+          this._current = this._schedulerQueue.shift();
+          let retval = this._current.func.apply(global, current.args);
+          if (!retval) {
```
juliemr: What if the callback returns 0 or null?

lib/zone-spec/fake-async-test.ts
```diff
@@ -0,0 +1,260 @@
...
+    tick(millis: number = 0): void {
+      this._currentTime += millis;
+      while (this._schedulerQueue.length > 0) {  
+        let current = this._schedulerQueue[0];        
+        if (this._currentTime < current.endTime) {
+          // Done processing the queue since it's sorted by endTime.
+          break;
+        } else {
+          // Time to run scheduled function. Remove it from the head of queue.
+          this._current = this._schedulerQueue.shift();
```
juliemr: Why store `_current` as a member on `this` instead of just keeping it a local variable?

lib/zone-spec/fake-async-test.ts
juliemr: Ah, I see the wrapper below. Never mind

juliemr:
Looks good overall!

vikerman:
Travis error - I need a better way to check whether the macro task is a XHR related call. Will check with Misko.

lib/zone-spec/fake-async-test.ts
vikerman: Good catch - Initial version of it used _current to cancel currently scheduled setInterval. But now that's refactored out into _fnAndFlush.

test/zone-spec/fake-async-test.spec.ts
vikerman: tick(millis) has a default value of 0 for millis. This was from the previous behavior of fakeAsync and provides an easy shorthand to flush the timer and microtask queue without advancing the time.

test/zone-spec/fake-async-test.spec.ts
vikerman: Removed.

PR #338, opened 2016-04-29
Add ref/unref to timers for Node.js

### LINKED ISSUE #336: After Zone.js patches things setTimeout(...).unref() doesn't work in Node.js
From https://github.com/angular/zone.js/blob/master/lib/node/node.ts#L16, I've inserted two lines that show the problem:

```js
setTimeout(function () {}, 50).unref() // works
if (shouldPatchGlobalTimers) {
  timers_1.patchTimer(_global, set, clear, 'Timeout');
  timers_1.patchTimer(_global, set, clear, 'Interval');
  timers_1.patchTimer(_global, set, clear, 'Immediate');
}
setTimeout(function () {}, 50).unref() // throws error
```

I'm importing `angular2-universal/polyfills` which imports `zone.js/dist/zone-node.js`.
Using Node v4.4.3.

matthewloring:
Fixes #336

### DIFF
--- a/lib/common/timers.ts
+++ b/lib/common/timers.ts
@@ -36,7 +36,17 @@ export function patchTimer(
         delay: (nameSuffix === 'Timeout' || nameSuffix === 'Interval') ? args[1] || 0 : null,
         args: args
       };
-      return zone.scheduleMacroTask(setName, args[0], options, scheduleTask, clearTask);
+      var task = zone.scheduleMacroTask(setName, args[0], options, scheduleTask, clearTask) ;
+      if (!task) {
+        return task;
+      }
+      // Node.js must additionally support the ref and unref functions.
+      var handle = (<TimerOptions>task.data).handleId;
+      if ((<any>handle).ref && (<any>handle).unref) {
+        (<any>task).ref = (<any>handle).ref.bind(handle);
+        (<any>task).unref = (<any>handle).unref.bind(handle);
+      }
+      return task;
     } else {
       // cause an error by calling it directly.
       return delegate.apply(window, args);
--- a/test/common/setInterval.spec.ts
+++ b/test/common/setInterval.spec.ts
@@ -1,5 +1,5 @@
 'use strict';
-import {zoneSymbol} from "../../lib/common/utils";
+import {isNode, zoneSymbol} from "../../lib/common/utils";
 
 describe('setInterval', function () {
 
@@ -25,6 +25,10 @@ describe('setInterval', function () {
       };
       expect(Zone.current.name).toEqual(('TestZone'));
       cancelId = setInterval(intervalFn, 10);
+      if (isNode) {
+        expect(typeof cancelId.ref).toEqual(('function'));
+        expect(typeof cancelId.unref).toEqual(('function'));
+      }
 
       // This icky replacer is to deal with Timers in node.js. The data.handleId contains timers in
       // node.js. They do not stringify properly since they contain circular references.
--- a/test/common/setTimeout.spec.ts
+++ b/test/common/setTimeout.spec.ts
@@ -1,4 +1,4 @@
-import {zoneSymbol} from '../../lib/common/utils';
+import {isNode, zoneSymbol} from '../../lib/common/utils';
 
 describe('setTimeout', function () {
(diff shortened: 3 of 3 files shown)


matthewloring:
cc @mhevery @ofrobots @jmdobry

test/common/setInterval.spec.ts
```diff
@@ -1,6 +1,12 @@
 'use strict';
 import {zoneSymbol} from "../../lib/common/utils";
 
+var isNode = true;
```
mhevery: 1) This is a better test `isNode = typeof window !== 'undefined'`.
2) could we add this to a separate file and import it from there?

test/common/setInterval.spec.ts
```diff
@@ -25,6 +31,9 @@ describe('setInterval', function () {
       };
       expect(Zone.current.name).toEqual(('TestZone'));
       cancelId = setInterval(intervalFn, 10);
+      if (isNode) {
+        expect(typeof cancelId.ref).toEqual(('function'));
```
mhevery: also add test for `unref`

test/common/setTimeout.spec.ts
```diff
@@ -1,5 +1,11 @@
 import {zoneSymbol} from '../../lib/common/utils';
 
+var isNode = true;
+try {
```
mhevery: same heer.

test/common/setTimeout.spec.ts
```diff
@@ -22,6 +28,9 @@ describe('setTimeout', function () {
       };
       expect(Zone.current.name).toEqual(('TestZone'));
       cancelId = setTimeout(timeoutFn, 3);
+      if (isNode) {
+        expect(typeof cancelId.ref).toEqual(('function'));
```
mhevery: also `unref`

lib/common/timers.ts
```diff
@@ -36,7 +36,17 @@ export function patchTimer(
...
       };
-      return zone.scheduleMacroTask(setName, args[0], options, scheduleTask, clearTask);
+      var task = zone.scheduleMacroTask(setName, args[0], options, scheduleTask, clearTask) ;
+      if (!task) {
+        return task;
+      }
+      // Node.js must additionally support the ref and unref functions.
+      var handle = (<TimerOptions>task.data).handleId;
+      if ((<any>handle).ref && (<any>handle).unref) {
+        (<any>task).ref = (<any>handle).ref.bind(handle);
```
mhevery: What do these functions do? Do they need to be wrapped with zone specific logic?

test/common/setInterval.spec.ts
vicb: wrt to 1 it will not work in web workers, right ?

test/common/setTimeout.spec.ts
matthewloring: Done.

test/common/setInterval.spec.ts
matthewloring: Done.

test/common/setInterval.spec.ts
matthewloring: 1) I am not familiar with the web platform or web workers. I can replace this with any check that is desired.
2) Where is the best place for this to live? Should it get its own file?

lib/common/timers.ts
matthewloring: They determine whether the existence of a pending timer should keep the process alive so work can be completed. This should be independent of zones behavior and should not require zone logic.

test/common/setInterval.spec.ts
mhevery: I think this should do the trick:

1) `var isNode = typeof global == 'object';`
2) We should import `isNode` from `../../lib/common/utils`

test/common/setTimeout.spec.ts
matthewloring: Done.

test/common/setInterval.spec.ts
matthewloring: Done.

takeshi:
I want to apply this patch to use @google/cloud-debug with zone.js on GAE  Node.JS. 
But this patch is not published to npm repository.  Please publish it.

PR #348, opened 2016-05-19
Fixed the home page example. Now it works as described below.

KwintenP:
Adapted the example in the README.md. It didn't work no more. 
With these changes, it should work as described underneath the example.


KwintenP:
I signed it! :)

PR #375, opened 2016-07-12
feat: return setTimeout Id in ZoneTask.toString

jahtalab:


jahtalab:
I signed it!

dist/zone.js.d.ts
```diff
@@ -372,6 +372,10 @@ interface TaskData {
      * Delay in milliseconds when the Task will run.
      */
     delay?: number;
+    /**
+     * identifier returned by the native setTimeout.
+     */
+    handleId?: number;
```
mhevery: TaskData is generic and can not have task specific data.

lib/zone.ts
```diff
@@ -395,6 +395,11 @@ interface TaskData {
    * Delay in milliseconds when the Task will run.
    */
   delay?: number;
+
+  /**
+   * identifier returned by the native setTimeout. 
+   */
+  handleId?: number;
```
mhevery: TaskData is generic and can not have task specific data.

PR #383, opened 2016-07-24
Pass actual error to failCallback

### LINKED ISSUE #379: AsyncTestZoneSpec calls failCallback with error message
Looking at the code for the `AsyncTestZoneSpec`, specifically [line 50](https://github.com/angular/zone.js/blob/master/lib/zone-spec/async-test.ts#L50), I see that the caught error is not passed to the `failCallback`, but rather only the error message.

In the case where a JavaScript error, such as a TypeError, is thrown during an `async` wrapped test, the callstack for the original error is lost.

Currently, the `async` function knows how to handle both `string` and `Error` in the `failCallback`, so why can't the original error be passed to it?

Example Test:

```javascript
let myUndefined; // deliberately left undefined, to cause a TypeError

it('should have a stack trace for TypeError with async()', async(() => {
    setTimeout(() => {
        expect(myUndefined.value).toBe(1);
    }, 300);
}));
```

Current Stack Trace:

```
    Failed: Cannot read property 'value' of undefined
    Error: Cannot read property 'value' of undefined
        at AsyncTestZoneSpec.eval [as _failCallback] (node_modules/@angular/core/testing/async.js:34:38)
        at AsyncTestZoneSpec.onHandleError (node_modules/zone.js/dist/async-test.js:107:30)
        at ZoneDelegate.handleError (node_modules/zone.js/dist/zone.js:327:40)
        at Zone.runTask (node_modules/zone.js/dist/zone.js:259:45)
        at ZoneTask.invoke (node_modules/zone.js/dist/zone.js:423:34)
```

schmuli:
When handling an error in the `AsyncTestZoneSpec`, instead of just passing the error message to the `_failCallback`, pass the actual error instance already available. This ensures that the complete stack trace is printed when logging a failed test.

Closes #379

### DIFF
--- a/lib/zone-spec/async-test.ts
+++ b/lib/zone-spec/async-test.ts
@@ -47,7 +47,7 @@
       // Let the parent try to handle the error.
       const result = parentZoneDelegate.handleError(targetZone, error);
       if (result) {
-        this._failCallback(error.message ? error.message : 'unknown error');
+        this._failCallback(error);
         this._alreadyErrored = true;
       }
       return false;
--- a/test/zone-spec/async-test.spec.ts
+++ b/test/zone-spec/async-test.spec.ts
@@ -208,7 +208,7 @@ describe('AsyncTestZoneSpec', function() {
       var testZoneSpec = new AsyncTestZoneSpec(() => {
         done.fail('expected failCallback to be called');
       }, (err) => {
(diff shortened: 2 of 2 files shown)


schmuli:
@juliemr, can you tell me if this is a valid change to the existing code?

PR #395, opened 2016-08-12
fix(browser): use XMLHttpRequest.DONE constant on target instead of t...

### LINKED ISSUE #394: XMLHttpRequest use readyState constants on target instead global interface
Some libraries (e.g. [offline.js](https://github.com/HubSpot/offline/issues/220)) patch XMLHttpRequest, but they don't expose readyState constants (for example XMLHttpRequest.DONE is undefined which causes the macrotask to never get invoked). 
Given that those constants are part of the [standard](https://www.w3.org/TR/XMLHttpRequest/#xmlhttprequest) only for the actual instance, we should use those constants on the target instead of global XMLHttpRequest

jahtalab:
..he global interface

fixes #394

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -55,7 +55,7 @@ function patchXHR(window: any) {
   function scheduleTask(task: Task) {
     var data = <XHROptions>task.data;
     data.target.addEventListener('readystatechange', () => {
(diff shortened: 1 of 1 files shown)


mhevery:
Your change fails the continuous build. Could you have a look?

jahtalab:
@mhevery , Just needed a rebase, Done.

PR #396, opened 2016-08-14
Proxy zone spec

mhevery:
This is prerequisite to fix: https://github.com/angular/angular/issues/10503

### DIFF
--- a/lib/jasmine/jasmine.ts
+++ b/lib/jasmine/jasmine.ts
@@ -4,11 +4,6 @@
 if (!Zone) {
   throw new Error('zone.js does not seem to be installed');
 }
-
-const SET_TIMEOUT = '__zone_symbol__setTimeout';
-const _global = typeof window == 'undefined' ? global : window;
-
-
 // When you have in async test (test with `done` argument) jasmine will
 // execute the next test synchronously in the done handler. This makes sense
 // for most tests, but now with zones. With zones running next test
@@ -17,10 +12,14 @@ const _global = typeof window == 'undefined' ? global : window;
 // it. We override the `clearStack` method which forces jasmine to always
 // drain the stack before next test gets executed.
 (<any>jasmine).QueueRunner = (function (SuperQueueRunner) {
+  const originalZone = Zone.current;
   // Subclass the `QueueRunner` and override the `clearStack` method.
 
   function alwaysClearStack(fn) {
-    _global[SET_TIMEOUT](fn, 0);
+    const zone: Zone = Zone.current.getZoneWith('JasmineClearStackZone')
+        || Zone.current.getZoneWith('ProxyZoneSpec')
+        || originalZone;
+    zone.scheduleMicroTask('jasmineCleanStack', fn);
   }
 
   function QueueRunner(options) {
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -80,6 +80,7 @@
     private _scheduler: Scheduler = new Scheduler();
     private _microtasks: Function[] = [];
     private _lastError: Error = null;
+    private _uncaughtPromiseErrors: {rejection: any}[] = Promise[Zone['__symbol__']('uncaughtPromiseErrors')];
         
     pendingPeriodicTimers: number[] = [];
     pendingTimers: number[] = [];
@@ -172,7 +173,8 @@
     }
 
     private _resetLastErrorAndThrow(): void {
-      let error = this._lastError;
+      let error = this._lastError || this._uncaughtPromiseErrors[0];
+      this._uncaughtPromiseErrors.length = 0;
       this._lastError = null;
       throw error;
     }
@@ -188,14 +190,17 @@
 
     flushMicrotasks(): void {
       FakeAsyncTestZoneSpec.assertInZone();
-      while (this._microtasks.length > 0) {
-        let microtask = this._microtasks.shift();
-        microtask();
-        if (this._lastError !== null) {
+      const flushErrors = () => {
+        if (this._lastError !== null || this._uncaughtPromiseErrors.length) {
           // If there is an error stop processing the microtask queue and rethrow the error.
           this._resetLastErrorAndThrow();
         }
       }
+      while (this._microtasks.length > 0) {
+        let microtask = this._microtasks.shift();
+        microtask();
+      }
+      flushErrors();
     }
 
     // ZoneSpec implementation below.
--- a/lib/zone-spec/proxy.ts
+++ b/lib/zone-spec/proxy.ts
@@ -1,0 +1,128 @@
+(function () {
+  class ProxyZoneSpec implements ZoneSpec {
+    name: string = 'ProxyZone';
+
+    private _delegateSpec: ZoneSpec;
+
+    properties: {[k: string]: any} = {'ProxyZoneSpec': this};
+    propertyKeys: string[] = null;
+
+    static get(): ProxyZoneSpec {
+      return Zone.current.get('ProxyZoneSpec');
+    }
+
+    static isLoaded(): boolean {
+      return  ProxyZoneSpec.get() instanceof ProxyZoneSpec;
+    }
+
+    static assertPresent(): ProxyZoneSpec {
+      if (!this.isLoaded()) {
+        throw new Error(`Expected to be running in 'ProxyZone', but it was not found.`);
+      }
+      return ProxyZoneSpec.get();
+    }
+
+    constructor(private defaultSpecDelegate: ZoneSpec = null) {
+      this.setDelegate(defaultSpecDelegate);
+    }
+
+
+    setDelegate(delegateSpec: ZoneSpec) {
+      this._delegateSpec = delegateSpec;
+      this.propertyKeys && this.propertyKeys.forEach((key) => delete this.properties[key]);
+      this.propertyKeys = null;
+      if (delegateSpec && delegateSpec.properties) {
+        this.propertyKeys = Object.keys(delegateSpec.properties);
+        this.propertyKeys.forEach((k) => this.properties[k] = delegateSpec.properties[k]);
+      }
+    }
+
+    getDelegate() {
+      return this._delegateSpec;
+    }
+
+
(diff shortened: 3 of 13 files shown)


lib/zone.ts
```diff
@@ -134,13 +134,23 @@ interface Zone {
...
+   * [ZoneSpec.properties] to configure the set of properties associated with the current zone.
    *
    * @param key The key to retrieve.
    * @returns {any} Tha value for the key, or `undefined` if not found.
    */
   get(key: string): any;
   /**
+   * Returns a Zone which defines a `key`.
+   *
+   * Recursively search the parent Zone until a Zone which has a `key` is found.
```
juliemr: Can you change this to mention that `key` is a property? Something like:

```
* Recursively search the parent Zone until a Zone which has a property `key` is found.
```

lib/zone.ts
```diff
@@ -134,13 +134,23 @@ interface Zone {
...
+   * otherwise.
+   *
+   * @param key The key to retrieve.
+   * @returns {Zone} Tha Zone which defines the `key`, `null` if not found.
+   */
+  getZone(key: string): Zone;
```
juliemr: Maybe consider renaming to `getZoneWith(key: string)`

lib/zone.ts
```diff
@@ -1057,6 +1075,9 @@ const Zone: ZoneType = (function(global: any) {
     constructor(executor: (resolve : (value?: R | Thenable<R>) => void,
                            reject: (error?: any) => void) => void) {
       const promise: ZoneAwarePromise<R> = this;
+      if (!(promise instanceof ZoneAwarePromise)) {
+        throw new Error('Must be an instanceof Promise.');
```
juliemr: Should this be `Must be an instanceof ZoneAwarePromise`?

test/common/Promise.spec.ts
```diff
@@ -225,13 +240,13 @@ describe('Promise', ifEnvSupports('Promise', function () {
         }).run(() => {
           zone = Zone.current;
           task = Zone.currentTask;
-          Promise.reject('rejectedErrorShouldBeHandled');
+          Promise.reject('rejectedError3ShouldBeHandled');
```
juliemr: This seems like a typo?

test/zone-spec/proxy.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+
+    it('should assert that it is in or out of ProxyZone', () => {
+      expect(() => ProxyZoneSpec.assertPresent()).toThrow();
+      expect(ProxyZoneSpec.isLoaded()).toBe(false);
+      expect(ProxyZoneSpec.get()).toBe(undefined);
+      proxyZone.run(() => {
+        expect(ProxyZoneSpec.isLoaded()).toBe(true);
+        expect(() => ProxyZoneSpec.assertPresent()).not.toThrow();
+      expect(ProxyZoneSpec.get()).toBe(proxyZoneSpec);
```
juliemr: Indentation error.

test/zone-spec/proxy.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      });
+    });
+
+    it('should reset properties', () => {
+      expect(proxyZone.get('key')).toBe(undefined);
```
juliemr: This test would be more readable if 'key' were a nonsense string like `expect(proxyZone.get('foo'))` - the 'key' is kind of confusing in the documentation so I think it's important to be clear here that it's not a magic string.

test/zone-spec/proxy.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+      proxyZoneSpec.setDelegate(otherDelegate);
+      expect(proxyZoneSpec.getDelegate()).toEqual(otherDelegate);
+      proxyZoneSpec.resetDelegate();
+      expect(proxyZoneSpec.getDelegate()).toEqual(defaultDelegate);
+    }); 
+  });
+
+  describe('forwarding', () => {
+    const proxyZoneSpec = new ProxyZoneSpec();
```
juliemr: Can these be called in a `beforeEach`? When Jasmine calls `describe` can be confusing since all `describe` blocks will get called before any tests (I think...)

lib/zone-spec/proxy.ts
```diff
@@ -0,0 +1,128 @@
+(function () {
+  class ProxyZoneSpec implements ZoneSpec {
+    name: string = 'ProxyZone';
+
+    private _delegateSpec: ZoneSpec;
+
+    properties: {[k: string]: any} = {'ProxyZoneSpec': this};
```
juliemr: What about `ProxyZoneFlag` or `ProxyZonePlaceholder`?

juliemr: Ah no, I see that it returns itself. Probably doesn't make sense to change it then.

juliemr:
CLA bot: this is fine.

Bunch of little comments but overall looks good.

juliemr:
Travis error looks like it might be legitimate.

test/common/Promise.spec.ts
IgorMinar: +1

I'm not sure how this change cleans up the tests.

also the commit label should be `chore(test)` should be `test(promise):`

lib/zone.ts
```diff
@@ -141,6 +141,16 @@ interface Zone {
    */
   get(key: string): any;
   /**
+   * Returns a Zone which defines a `key`.
+   *
+   * Recursively search the parent Zone until a Zone which has a `key` is found.
+   * otherwise.
+   *
+   * @param key The key to retrieve.
```
IgorMinar: you are not retrieving a key. `The key to use for identification of the returned zone.`

lib/zone.ts
```diff
@@ -141,6 +141,16 @@ interface Zone {
    */
```
IgorMinar: Tha -> The value

lib/zone.ts
IgorMinar: getZoneWithProperty ?

definitely not just `getZone` as that's too ambiguous

lib/zone-spec/fake-async-test.ts
```diff
@@ -188,14 +190,18 @@
...
+        if (this._lastError !== null || this._uncaughtPromiseErrors.length) {
           // If there is an error stop processing the microtask queue and rethrow the error.
           this._resetLastErrorAndThrow();
         }
       }
+      flushErrors();
+      while (this._microtasks.length > 0) {
+        let microtask = this._microtasks.shift();
+        microtask();
+        flushErrors();
```
IgorMinar: wouldn't this throw prematurely? I believe the spec-ed behavior and certainly chrome's behavior is that the rejection handler can be registered at any point within the current turn of the event loop:

```
// this does't throw
p = Promise.reject('foo'); Promise.resolve('x').then(() =>  p.then(() => {}, () => {}));


// this throws
p = Promise.reject('foo'); setTimeout(() => { p.then(() => {}, () => {}) }, 0);
```

lib/jasmine/jasmine.ts
```diff
@@ -4,11 +4,6 @@
 if (!Zone) {
```
IgorMinar: the commit message contains unbalanced parentheses

lib/jasmine/jasmine.ts
```diff
@@ -17,10 +12,14 @@ const _global = typeof window == 'undefined' ? global : window;
...
 (<any>jasmine).QueueRunner = (function (SuperQueueRunner) {
+  const originalZone = Zone.current;
   // Subclass the `QueueRunner` and override the `clearStack` method.
 
   function alwaysClearStack(fn) {
-    _global[SET_TIMEOUT](fn, 0);
+    const zone: Zone = Zone.current.getZone('JasmineClearStackZone')
+        || Zone.current.getZone('ProxyZoneSpec')
+        || originalZone;
+    zone.scheduleMicroTask('jasmineCleanStack', fn);
```
IgorMinar: isn't this going to result in StackOverflowErrors since we are not clearing the stack?

I believe that jasmine calls the "clearStack" method only once in a while when it actually needs to clear the stack.

lib/zone.ts
mhevery: `ZoneAwarePromise` is an internal name which the user does not see.

lib/zone-spec/proxy.ts
mhevery: leaving as is.

test/zone-spec/proxy.spec.ts
mhevery: fixed

test/zone-spec/proxy.spec.ts
mhevery: good idea, done

test/zone-spec/proxy.spec.ts
mhevery: done

lib/zone.ts
mhevery: fixed

lib/zone.ts
mhevery: fixed

lib/zone.ts
mhevery: updated

lib/zone.ts
mhevery: going with `getZoneWith`

lib/zone-spec/fake-async-test.ts
mhevery: good point. moved after the loop.

lib/jasmine/jasmine.ts
mhevery: LOL, Yes it does. fixed

lib/jasmine/jasmine.ts
mhevery: Good observation, but it will not happen. The reason is that is is the top most frame which  does the microtask execution, and so it will never nest. This method will always return and only later execute the microtask.

mhevery:
Closed with: 5f519de5dc9b5db45e25991deb15db2cf150e38f

PR #408, opened 2016-08-19
fix(tasks): do not drain the microtask queue early.

juliemr:
Fixes a bug where a event task invoked inside another task
would drain the microtask queue too early. This would mean that microtasks
would be called unexpectedly in the middle of what should have been
a block of synchronous code.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -504,6 +504,9 @@ const Zone: ZoneType = (function(global: any) {
   if (global.Zone) {
     throw new Error('Zone already loaded.');
   }
+
+  let _invokingTasks = 0;
+
   class Zone implements AmbientZone {
     static __symbol__: (name: string) => string = __symbol__;
 
@@ -835,10 +838,16 @@ const Zone: ZoneType = (function(global: any) {
       this.callback = callback;
       const self = this;
       this.invoke = function () {
+        let drainAfter: boolean;
         try {
+          drainAfter = !_invokingTasks;
+          _invokingTasks++;
           return zone.runTask(self, this, <any>arguments);
         } finally {
-          drainMicroTaskQueue();
+          if (drainAfter) {
+            drainMicroTaskQueue();
+          }
+          _invokingTasks--;
         }
       };
     }
--- a/test/browser/element.spec.ts
+++ b/test/browser/element.spec.ts
@@ -50,6 +50,22 @@ describe('element', function () {
     button.dispatchEvent(clickEvent);
   });
 
+  fit('should not call microtasks early when an event is invoked', function() {
+    var log = '';
+    var logFunction = function logFunction () {
+      log += 'a';
+    };
+
+    Zone.current.scheduleMicroTask('test', () => {
+      log += 'microtask';
+    });
+    button.addEventListener('click', logFunction);
+
+    button.click();
+
+    expect(log).toEqual('a');
+  });
+
   it('should work with addEventListener when called with an EventListener-implementing listener', function () {
     var eventListener = {
       x: 5,
--- a/test/common/zone.spec.ts
+++ b/test/common/zone.spec.ts
@@ -150,6 +150,33 @@ describe('Zone', function () {
       ]);
     });
   });
+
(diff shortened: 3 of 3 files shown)


juliemr:
I added another commit showing a failing test that my change does not yet deal with. @mhevery - I could use feedback on whether this test is legitimate or not. I _think_ it's showing unintended consequences of invoking an event task, and we should change the behavior here.

juliemr:
Conceptually - why do we need to drain the microtask queue after invoking a task? Shouldn't the scheduling from `scheduleQueueDrain` take care of it?

lib/zone.ts
```diff
@@ -504,6 +504,9 @@ const Zone: ZoneType = (function(global: any) {
   if (global.Zone) {
     throw new Error('Zone already loaded.');
   }
+
+  let _invokingTasks = 0;
```
mhevery: Could we rename this to: `numberOfNestedTaskFrames`

lib/zone.ts
```diff
@@ -835,10 +838,16 @@ const Zone: ZoneType = (function(global: any) {
       this.callback = callback;
       const self = this;
       this.invoke = function () {
+        let drainAfter: boolean;
         try {
+          drainAfter = !_invokingTasks;
```
mhevery: I don't think `drainAfter` is needed.

lib/zone.ts
```diff
@@ -835,10 +838,16 @@ const Zone: ZoneType = (function(global: any) {
...
+          _invokingTasks++;
           return zone.runTask(self, this, <any>arguments);
         } finally {
-          drainMicroTaskQueue();
+          if (drainAfter) {
```
mhevery: `if (numberOfNestedTaskFrames == 1) {`

test/common/zone.spec.ts
```diff
@@ -150,6 +150,33 @@ describe('Zone', function () {
...
+  describe('invoking tasks', () => {
+    var log;
+    function noop() {}
+
+
+    beforeEach(() => {
+      log = [];
+    });
+
+    it('should not drain the microtask queue too early', () => {
```
mhevery: change to `should drain the microtask queue on the outter most nested task.`

test/common/zone.spec.ts
```diff
@@ -150,6 +150,33 @@ describe('Zone', function () {
...
+
+      var macro = z.scheduleMacroTask('test', () => {
+        event.invoke();
+        // At this point, we should not have invoked the microtask.
+        expect(log).toEqual([
+          'eventTask'
+        ]);
+      }, null, noop, noop);
+
+      macro.invoke();
```
mhevery: Yes, this is a valid test.

test/browser/element.spec.ts
```diff
@@ -50,6 +50,22 @@ describe('element', function () {
     button.dispatchEvent(clickEvent);
   });
 
+  fit('should not call microtasks early when an event is invoked', function() {
```
mhevery: remove `fit`

mhevery:
Draining of microtask queue should happen as the last thing before we turn the control over to the C code. This means that it is the last thing in the top most frame should do. The code in `scheduleQueueDrain ` you are referring to should be changed from:
`if (!_drainScheduled && !_currentTask && _microTaskQueue.length == 0) {`
to 
`if (!_drainScheduled && numberOfNestedTaskFrames == 0 && _microTaskQueue.length == 0) {`

What this code does is to say hay, somehow you are not in a task, and this is a problem, because queue draining will not happen, so let me schedule a manual queue drain. Why we are not in task? because when the original Zone is loaded the code is running outside any Zone task. It is only after the zone is loaded and a task is scheduled do we have a way to drain the queue at the right moment. So you can think of `scheduleQueueDrain` as a way to cheat to bootstrap ourselves. So this method is to kickstart the task process.

mhevery:
Have a look at my implementation: https://github.com/angular/zone.js/pull/409

mhevery:
closed by ff88bb43d29711cd6d6c575f6855590455463ae3

PR #409, opened 2016-08-19
fix(tasks): do not drain the microtask queue early.

mhevery:
Fixes a bug where a event task invoked inside another task
would drain the microtask queue too early. This would mean that microtasks
would be called unexpectedly in the middle of what should have been
a block of synchronous code.

### DIFF
--- a/test/browser/element.spec.ts
+++ b/test/browser/element.spec.ts
@@ -51,16 +51,15 @@ describe('element', function () {
   });
 
   it('should not call microtasks early when an event is invoked', function(done) {
-    // we have to run this test in setTimeout to guarantee that we are running in an existing task
-    setTimeout(() => {
-      var log = '';
+    var log = '';
+    button.addEventListener('click', () => {
       Zone.current.scheduleMicroTask('test', () => log += 'microtask;');
-      button.addEventListener('click', () => log += 'click;');
-      button.click();
-
-      expect(log).toEqual('click;');
-      done();
+      log += 'click;'
     });
+    button.click();
+
+    expect(log).toEqual('click;');
+    done();
   });
 
   it('should call microtasks early when an event is invoked', function(done) {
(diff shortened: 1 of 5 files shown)


test/browser/element.spec.ts
```diff
@@ -50,6 +50,48 @@ describe('element', function () {
...
+     * 3. Pay the cost of throwing an exception in event tasks and verifying that we are the
+     *    top most frame.
+     */
+    global[Zone['__symbol__']('setTimeout')](() => {
+      var log = '';
+      Zone.current.scheduleMicroTask('test', () => log += 'microtask;');
+      button.addEventListener('click', () => log += 'click;');
+      button.click();
+
+      expect(log).toEqual('click;microtask;');
```
juliemr: Tests are showing this coming up as `microtask;click;`

test/browser/element.spec.ts
```diff
@@ -50,6 +50,51 @@ describe('element', function () {
...
+     *
+     * Given that this can only arise when someone is emulating clicks on DOM in a synchronous
+     * fashion we have few choices:
+     * 1. Ignore as this is unlikely to be a problem outside of tests.
+     * 2. Monkey patch the event methods to increment the _numberOfNestedTaskFrames and prevent
+     *    eager drainage.
+     * 3. Pay the cost of throwing an exception in event tasks and verifying that we are the
+     *    top most frame.
+     *
+     * For now we are choosing to ignore it and assume that this arrises in tests only.
```
juliemr: Update this comment to mention that we patch Jasmine, please.

test/jasmine-patch.spec.ts
```diff
@@ -1,3 +1,8 @@
+beforeEach(() => {
+  // assert that each jasmine run has a task, so that darinMicrotask works properly.
```
juliemr: typo: drainMicrotask

juliemr:
Two tiny comments, then LGTM.

test/browser/element.spec.ts
mhevery: added.

test/jasmine-patch.spec.ts
mhevery: fixed

PR #412, opened 2016-08-22
Fix spelling error of 'describe' in jasmine patch

abierbaum:
Looks like a spelling error was preventing the jasmine patch from patching "describe".

### DIFF
--- a/lib/jasmine/jasmine.ts
+++ b/lib/jasmine/jasmine.ts
@@ -31,7 +31,7 @@
 
   // Monkey patch all of the jasmine DSL so that each function runs in appropriate zone.
   const jasmineEnv = jasmine.getEnv();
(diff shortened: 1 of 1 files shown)


abierbaum:
I signed it!

lib/jasmine/jasmine.ts
```diff
@@ -31,7 +31,7 @@
 
   // Monkey patch all of the jasmine DSL so that each function runs in appropriate zone.
   const jasmineEnv = jasmine.getEnv();
-  ['desribe', 'xdescribe', 'fdescribe'].forEach((methodName) => {
+  ['describe', 'xdescribe', 'fdescribe'].forEach((methodName) => {
```
abierbaum: Just noticed this while looking through some stack traces.  No idea if it is needed but looks like a typo.

PR #415, opened 2016-08-24
fix(property-descriptor): do not use document object in web worker

urimor:
In propertyDescriptorPatch(), when running is Safari browser, patchViaCapturingAllTheEvents() is called
which uses document.addEventListener().
When this code runs in a web worker, there is no document.
Fix by changing document.addEventListener() to self.addEventListener.


urimor:
Oups, was not ment to be sent yet, will open an issue instead

PR #420, opened 2016-09-01
feat: assert that right ZoneAwarePromise is available

mhevery:


jelbourn:
LGTM

lib/zone.ts
```diff
@@ -508,6 +513,15 @@ const Zone: ZoneType = (function(global: any) {
   class Zone implements AmbientZone {
     static __symbol__: (name: string) => string = __symbol__;
 
+    static assertZonePatched() {
+      if (global.Promise !== ZoneAwarePromise) {
+        throw new Error("Zone.js has detected that ZoneAwarePromise `(window|global).Promise` " +
+            "has been overwritten.\n" +
+            "Most likely cause is that a Promise polyfill has been loaded " +
+            "after Zone.js (If you must load polyfills do so before Zone.js.)");
```
IgorMinar: "(Polyfilling promise api is not necessary when zone.js is loaded. If you must load one, do so before loading zone.js.)"

mhevery: updated

PR #421, opened 2016-09-01
fix(ZoneDelegate): updateTaskCount before scheduling the task

### LINKED ISSUE #377: Synchronous XHR request through zonejs results in:  More tasks executed then were scheduled
I am logging a bug report:

There seems to be a bug in the bookkeeping of `XMLHttpRequest` in zone.js.

If I  trigger a synchronous `XMLHttpRequest` when zonejs is loaded on the page I get the following error:
`Error: More tasks executed then were scheduled.`

The error does not occur if the request is asynchronous.

Here is a simple plunker with a repro: https://plnkr.co/edit/7Cn3JknfDIIHMIdqcuo0?p=preview

I have not looked into this extensively, but the error occurs here:
`next` is set to -1 which triggers the error condition.

```
ZoneDelegate.prototype._updateTaskCount = function (type, count) {
	            var counts = this._taskCounts;
	            var prev = counts[type];
	            var next = counts[type] = prev + count;
	            if (next < 0) {
	                throw new Error('More tasks executed then were scheduled.');
	            }
	            if (prev == 0 || next == 0) {
	                var isEmpty = {
	                    microTask: counts.microTask > 0,
	                    macroTask: counts.macroTask > 0,
	                    eventTask: counts.eventTask > 0,
	                    change: type
	                };
	                try {
	                    this.hasTask(this.zone, isEmpty);
	                }
	                finally {
	                    if (this._parentDelegate) {
	                        this._parentDelegate._updateTaskCount(type, count);
	                    }
	                }
	            }
	        };
```

jahtalab:
fixes #377

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -737,6 +737,10 @@ const Zone: ZoneType = (function(global: any) {
 
     scheduleTask(targetZone: Zone, task: Task): Task {
       try {
+        if (targetZone == this.zone) {
+          this._updateTaskCount(task.type, 1);
+        }
+      } finally {
         if (this._scheduleTaskZS) {
           return this._scheduleTaskZS.onScheduleTask(this._scheduleTaskDlgt, this.zone, targetZone, task);
         } else if (task.scheduleFn) {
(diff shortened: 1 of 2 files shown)


mhevery:
Merged as: 6e2f13c

PR #424, opened 2016-09-01
fix(jasmine): propagate all arguments of it/describe/etc...

mhevery:


juliemr:
Travis failure looks legitimate, but also looks like it was caused by the previous commit.

lib/jasmine/jasmine.ts
```diff
@@ -39,14 +39,16 @@
   });
   ['it', 'xit', 'fit'].forEach((methodName) => {
     let originalJasmineFn: Function = jasmineEnv[methodName];
-    jasmineEnv[methodName] = function(description: string, specDefinitions: Function) {
-      return originalJasmineFn.call(this, description, wrapTestInZone(specDefinitions));
+    jasmineEnv[methodName] = function(description: string, specDefinitions: Function, timeout: number) {
```
juliemr: Let's either add the `timeout` param both here and in 47 or in neither, it's weird to just add it to one.

mhevery: done

PR #426, opened 2016-09-02
fix(zone): remove superfluous comparison

mprobst:


mhevery:
removed in separate PR

PR #431, opened 2016-09-09
fix(promise): support more aggressive optimization.

mprobst:
ZoneAwarePromise does not technically implement or inherit the static
side of Promise, so aggressive optimizers may drop static methods such
as `all`, `race`, `reject`, and `resolve`.

This change explicitly exports them into quoted properties, which is
harmless for normal use and protects users of more aggressive
optimizers, such as Closure Compiler in typed mode.


vicb:
@mhevery @mprobst what about failing CI ?

```
Your test run exceeded 120 minutes. 
One possible solution is to split up your test run.
```

mprobst:
Must be a flake, how would this change cause any change in tests?

vicb:
yes it was, master is green. Sorry for the noise.

PR #438, opened 2016-09-12
feat(node): patch most fs methods

alxhub:


mhevery:
LGTM, will merge after CI goes green.

Issue #439, opened 2016-09-13
Calling fetch() with no arguments prevents page from loading when Chrome Developer Tools are open


pe8ter:
On [this line](https://github.com/angular/zone.js/blob/master/lib/zone.ts#L1148) there appears to be an on-purpose erroneous call to `fetch()` which normally requires an argument:

    const fetchPromise = global['fetch']();

This error causes the Chrome Developer Tools to pause which means my page won't load. I realize I can turn off the *Pause on Exceptions* flag but then I can't use that feature for my own code. It's not ideal to miss possible exceptions because a dependency forces me to disable this browser feature.

Tragetaschen:
Bootstrapping in Edge fails entirely

jeffbcross:
Closing as dupe of #436

jramstedt:
I'm still getting paused by this error in 0.6.23 release.

pe8ter:
This bug still occurs in Zone.js version 0.6.25, but both this issue and https://github.com/angular/zone.js/issues/436 are closed. Since that issue deals with MS Edge and this one is Chrome-only I think this issue should be reopened.

mhevery:
@pe8ter please re-open a new issue with new repro steps, as this repro steps no longer cause the issue.

Issue #437, opened 2016-09-10
zone.js issue with NativeScript


warren-bank:
<pre>
* file: ExampleProject/node_modules/nativescript-angular/application.js
  line: 2
  code: `import "zone.js/dist/zone-node"`
* file: ExampleProject/node_modules/nativescript-angular/node_modules/zone.js/dist/zone-node.js
  line: 670
  code: `var timers = require('timers')`
* note: in NativeScript, "timers" are re-implemented (ie: setTimeout, setInterval, etc)
  repo: https://github.com/NativeScript/NativeScript/tree/master/tns-core-modules/timer
  code: `var timer = require("timer")`
  note: "timer" NOT "timers"
* question:
    which module was most recently changed, such that they are no-longer compatible?
  answer:
  - zone.js
      https://github.com/angular/zone.js/blob/master/dist/zone-node.js
      https://github.com/angular/zone.js/commits/master/dist/zone-node.js
    note:
      2 commits within the last 5 hours
    commits:
      https://github.com/angular/zone.js/tree/55f815017df6bbda23eb0d1f61795f6f0412f51e/dist/zone-node.js
        date: Aug 22, 2016, 10:29 AM PDT
        line: 54
        code: `var timers = __webpack_require__(4);`
      https://github.com/angular/zone.js/tree/44c46fa5a615cfa3ca6ad9ea431e216ec6795bdf/dist/zone-node.js
        date: Sep 10, 2016, 9:04 AM PDT
        line: 664
        code: `var timers = require('timers');`
      https://github.com/angular/zone.js/tree/15d298b877f9da819055dad9d97cd50195c9d9f5/dist/zone-node.js
        date: Sep 10, 2016, 9:05 AM PDT
        line: 670
        code: `var timers = require('timers');`
* workaround?:
    - write a wrapper module named "timers", which exports "timer"
</pre>

warren-bank:
sorry, I should backtrack a bit and mention that this causes any NativeScript app (written with angular 2) to crash in Android (not sure about iOS, not a fan) when the app is launched with a stack trace that includes the following information:

<pre>
Unhandled Exception:
  Error calling module function

Error:
  Failed to find module: "timers", relative to: /app/tns_modules

StackTrace:
  - Frame: function: "require"
    Frame: file: "/data/data/org.nativescript.ExampleProject/files/app/tns_modules/zone.js/dist/zone-node.js", line: 670, column: 14
  - Frame: function: "require"
[...]

PR #440, opened 2016-09-14
Add a nativescript zone implementation

hdeshev:
The [nativescript-angular](https://github.com/NativeScript/nativescript-angular/) mobile renderer was using the zone-node.js bundle until 0.6.18+ added patching for the `timers` module and `fetch` promise-likes. Those are incompatible with NativeScript and cause runtime errors. The best solution to support zone.js in NativeScript in a way that minimizes future breakage seems to involve creating a separate bundle.

I based this bundle on the node one, just removing the unsupported patches. I also added a rudimentary test task that runs under node (to avoid spinning up device emulators and complicate build infrastructure).

`nativescript-angular` vendors the `zone-nativescript.js` bundle that this PR adds, but it would be great if we get rid of that in the future.

This PR takes care of #437 and the `fetch()` change is somewhat related to #439.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -60,6 +60,10 @@ gulp.task('build/zone-node.js', ['compile-esm'], function(cb) {
   return generateScript('./lib/node/node.ts', 'zone-node.js', false, cb);
 });
 
+gulp.task('build/zone-nativescript.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/nativescript/nativescript.ts', 'zone-nativescript.js', false, cb);
+});
+
 // Zone for the browser.
 gulp.task('build/zone.js', ['compile-esm'], function(cb) {
 
@@ -127,6 +131,7 @@ gulp.task('build', [
   'build/zone.js.d.ts',
   'build/zone.min.js',
   'build/zone-node.js',
+  'build/zone-nativescript.js',
   'build/jasmine-patch.js',
   'build/jasmine-patch.min.js',
   'build/long-stack-trace-zone.js',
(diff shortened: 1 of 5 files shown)


hdeshev:
I signed it!

dist/zone-nativescript.js
```diff
@@ -0,0 +1,842 @@
+var Zone$1 = (function (global) {
```
mhevery: Please Remove the changes to the `dist` folder. We update the dist folder on releases only.

gulpfile.js
```diff
@@ -169,3 +174,31 @@ gulp.task('test/node', ['compile'], function(cb) {
   jrunner.addSpecFiles(specFiles);
   jrunner.execute();
 });
+
+gulp.task('test/nativescript', ['compile'], function(cb) {
```
mhevery: Please also update the `package.json` file so that native scripts can be run using `npm run test-nativescript` command.

lib/zone.ts
```diff
@@ -1148,7 +1148,7 @@ const Zone: ZoneType = (function(global: any) {
       const fetchPromise = global['fetch']();
       // ignore output to prevent error;
       fetchPromise.then(() => null, () => null);
-      if (fetchPromise.constructor != NativePromise) {
+      if (fetchPromise.constructor != NativePromise && fetchPromise.constructor != ZoneAwarePromise) {
```
mhevery: Can you break this change into a separate commit. I don't think it has anything to do with NativeScript

test/nativescript_entry_point.ts
```diff
@@ -0,0 +1,20 @@
...
+import '../lib/zone-spec/sync-test';
+import '../lib/zone-spec/task-tracking';
+import '../lib/zone-spec/wtf';
+
+// Setup test environment
+import './test-env-setup';
+
+// List all tests here:
+import './common_tests';
+import './nativescript_tests';
```
mhevery: `nativescript_tests.ts` seems to be an empty file. Was that intended?

dist/zone-nativescript.js
hdeshev: Done.

gulpfile.js
hdeshev: Done

lib/zone.ts
hdeshev: Already done. The PR contains both commits.

test/nativescript_entry_point.ts
hdeshev: I thought it would look consistent with the node_tests.ts setup, but I can delete it. Your call.

PR #442, opened 2016-09-14
fix: use fetch(about://blank) to prevent exception on MS Edge

### LINKED ISSUE #436: Zone JS problem with Edge
I am using the latest EDGE browser, and this is the error I am getting, while loading Zone js

capture

### LINKED ISSUE #439: Calling fetch() with no arguments prevents page from loading when Chrome Developer Tools are open

mhevery:
Closes #436
Closes #439

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -1145,7 +1145,7 @@ const Zone: ZoneType = (function(global: any) {
   if (NativePromise) {
     patchThen(NativePromise);
     if (typeof global['fetch'] !== 'undefined') {
-      const fetchPromise = global['fetch']();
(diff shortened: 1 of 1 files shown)


jeffbcross:
LGTM

PR #451, opened 2016-09-16
fix: make fetch promise patching safe

mlaval:
If the `fetchPromise` is a `ZoneAwarePromise`, then patching it creates infinite loop of `.then()`.


vicb:
Please add a test

mlaval:
For the test, the only way I found to reproduce this issue is to load a `window.fetch` polyfill (e.g. https://github.com/github/fetch) in browsers which do not support this feature: http://caniuse.com/#feat=fetch

With the polyfill and without this fix, Safari 10 errors with `RangeError: Maximum call stack size exceeded.`, and IE11 with `Error: Out of stack space`.

PR #454, opened 2016-09-18
fix: run all timers in passage of time in a single fakeAsync's tick call

futurizing:
The `tick()` funtion should run a timer created during that passage of time if that timer is scheduled to run in that passage of time. I write it in a hurry. Any suggestion is welcome.

This fix #https://github.com/angular/angular/issues/10299.


futurizing:
I signed it!

PR #457, opened 2016-09-19
fix(bundling): switch to using umd bundles

### LINKED ISSUE #456: Zone.js browser bundle pollutes global namespace
The `browser.ts` file (https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts) is not being bundled inside any closure, resulting in all of its top-level functions and variables becoming globals.

This was discovered because this behaviour was breaking Plunker's preview by overriding `window.name` to `alert`. Down the rabbit hole we go and find ourselves here.

juliemr:
This means that we will stop polluting the global namespace when
zone.js is included.

Fixes #456

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,7 +18,17 @@ function generateScript(inFile, outFile, minify, callback) {
   inFile = path.join('./build-esm/', inFile).replace(/\.ts$/, '.js');
   var parts = [
     gulp.src('./build-esm/lib/**/*.js')
-        .pipe(rollup({ entry: inFile}))
+        .pipe(rollup({
+          entry: inFile,
+          format: 'umd',
+          banner: '/**\n'
+              + '* @license\n'
+              + '* Copyright Google Inc. All Rights Reserved.\n'
+              + '*\n'
+              + '* Use of this source code is governed by an MIT-style license that can be\n'
+              + '* found in the LICENSE file at https://angular.io/license\n'
+              + '*/'
+        }))
         .pipe(rename(outFile)),
   ];
   if (minify) {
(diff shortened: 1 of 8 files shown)


juliemr:
note: I've included the updates from running `npm run prepublish` here because it makes the review more meaningful.

Let me know if you'd like me to revert the changes to stuff in `dist/`

dist/async-test.js
```diff
@@ -1,3 +1,9 @@
+(function (global, factory) {
```
mhevery: We should add copyright here:

```
/**
 * @license
 * Copyright Google Inc. All Rights Reserved.
 *
 * Use of this source code is governed by an MIT-style license that can be
 */
```

mhevery: Also while it is nice to see final product, do revert `dist` folder. before checking in.

mhevery: We should add copyright here:

```
/**
 * @license
 * Copyright Google Inc. All Rights Reserved.
 *
 * Use of this source code is governed by an MIT-style license that can be
 */
```

mhevery: Also while it is nice to see final product, do revert dist folder. before checking in.

dist/async-test.js
```diff
@@ -1,3 +1,9 @@
...
+    typeof exports === 'object' && typeof module !== 'undefined' ? factory() :
+    typeof define === 'function' && define.amd ? define(factory) :
+    (factory());
+}(this, (function () { 'use strict';
+
 (function () {
```
mhevery: All of the specs have their own closure which auto-detects global. We should remove these?

juliemr: Done

dist/async-test.js
juliemr: Added copyright.

ACK for reverting dist folder, will do before checking in.

juliemr:
Responded to comments and updated the dist files - I'll squash and revert the dist files before merging if this looks good to you.

PR #460, opened 2016-09-20
fix(zonespecs): revert unwrapping of zonespecs which actually require…

juliemr:
… global

This reverts them back to 0.6.23 release status.


vicb:
LGTM if it makes angular/angular CI green

PR #462, opened 2016-09-20
fix: Support clearing the timeouts with numeric IDs

### LINKED ISSUE #461: Support clearing the timeouts with numerical IDs for Polymer compatibility
Currently, calling `clearTimeout(timeoutId: number)`, as well as the other patched timer methods, clears the timeout in browser, but does not cancel the corresponding zone task.

### Details

The patched timer methods, such as `setTimeout`, return a `Task` instance. At the same time, by the DOM specs, the return value should be numeric.

Polymer, for example, depends on the `setTimeout` return value to be numeric. In [`Polymer.Async` implementation](https://github.com/Polymer/polymer/blob/ff6e884ef4f309d41491333860a8bc9c2f178696/src/lib/async.html#L22), the `setTimeout` output is processed with a tilde operator `~`.

Given a `Task` instance, the numeric operations will implicitly invoke `task.toString()`. This way Zone’s task is spec-compatible. But the issues pop up when the resulting numbers are reused for clearing the timeouts.

### Impact

This issue could block the protractor integration tests for the Angular apps that use Polymer elements. The tests timeout trying to sync with the Angular page, they wait for the stable flag, which is never raised because of some timeouts cleared, while their tasks are still pending.

### Possible solutions

Ideally, the patched timer methods should return numeric IDs, as required by the spec. But that would be a big change, possibly breaking the code that depends on current Zone’s `Task` returns.

Another option is to extend the patched timer methods so that the `clearTimeout` and such can be called with a numeric timeout ID. This requires storing the tasks in an array indexed by timeoutIDs, which is easy.

platosha:
Fixes #461

### DIFF
--- a/lib/common/timers.ts
+++ b/lib/common/timers.ts
@@ -19,14 +19,21 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
   setName += nameSuffix;
   cancelName += nameSuffix;
 
+  const tasksByHandleId: Object = {};
+
   function scheduleTask(task: Task) {
     const data = <TimerOptions>task.data;
-    data.args[0] = task.invoke;
+    data.args[0] = function() {
+      task.invoke.apply(this, arguments);
+      delete tasksByHandleId[data.handleId];
+    };
     data.handleId = setNative.apply(window, data.args);
+    tasksByHandleId[data.handleId] = task;
     return task;
   }
 
   function clearTask(task: Task) {
+    delete tasksByHandleId[(<TimerOptions>task.data).handleId];
     return clearNative((<TimerOptions>task.data).handleId);
   }
 
@@ -59,7 +66,7 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
 
   clearNative =
       patchMethod(window, cancelName, (delegate: Function) => function(self: any, args: any[]) {
-        var task: Task = args[0];
+        var task: Task = typeof args[0] === 'number' ? tasksByHandleId[args[0]] : args[0];
         if (task && typeof task.type === 'string') {
           if (task.cancelFn && task.data.isPeriodic || task.runCount === 0) {
             // Do not cancel already canceled functions
--- a/test/common/setTimeout.spec.ts
+++ b/test/common/setTimeout.spec.ts
@@ -102,6 +102,27 @@ describe('setTimeout', function() {
     expect(typeof(cancelId.toString())).toBe('number');
   });
 
+  it('should allow cancelation by numeric timeout Id', function (done) {
+    // Node returns complex object from setTimeout, ignore this test.
+    if (isNode) {
+      done();
+      return;
+    }
+
+    const testZone = Zone.current.fork(Zone['wtfZoneSpec']).fork({ name: 'TestZone' });
+    testZone.run(() => {
+      const spy = jasmine.createSpy('spy');
+      const task: Task = <any>setTimeout(spy, 0);
+      const cancelId: number = <any>task;
+      clearTimeout(0 + cancelId);
+      setTimeout(function () {
+        expect(spy).not.toHaveBeenCalled();
+        expect(task.runCount).toEqual(-1);
+        done();
+      }, 1);
+    });
+  });
+
   it('should pass invalid values through', function() {
     clearTimeout(null);
     clearTimeout(<any>{});
--- a/test/common/setTimeout.spec.ts
+++ b/test/common/setTimeout.spec.ts
@@ -57,7 +57,7 @@ describe('setTimeout', function() {
       setTimeout(function() {
         expect(spy).not.toHaveBeenCalled();
         done();
-      });
+      }, 1);
     });
   });
 
@@ -72,7 +72,7 @@ describe('setTimeout', function() {
           clearTimeout(cancelId);
           done();
         });
-      });
+      }, 1);
     });
   });
 


platosha:
I signed it!

mhevery:
- [ ] Please rebase to master
- [ ] You need a test to demonstrate the issue.

lib/common/timers.ts
```diff
@@ -16,6 +16,8 @@ export function patchTimer(
   setName += nameSuffix;
   cancelName += nameSuffix;
 
+  var tasksByHandleId: Array<Task> = [];
```
mhevery: This array will grow forever. I think using a hash may be better here.

lib/common/timers.ts
```diff
@@ -55,6 +58,11 @@ export function patchTimer(
 
   clearNative = patchMethod(window, cancelName, (delegate: Function) => function(self: any, args: any[]) {
     var task: Task = args[0];
+    if (task && typeof task === 'number') {
+        var handleId = <number>task;
+        task = tasksByHandleId[handleId];
+        tasksByHandleId[handleId] = undefined;
```
mhevery: also this also has to be cleared when the task expires

platosha:
@mhevery added the test.

lib/common/timers.ts
platosha: Done.

lib/common/timers.ts
platosha: Done.

lib/common/timers.ts
```diff
@@ -19,14 +19,21 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
   setName += nameSuffix;
   cancelName += nameSuffix;
 
+  var tasksByHandleId: Object = {};
```
vicb: please use `const`

test/common/setTimeout.spec.ts
```diff
@@ -102,6 +102,27 @@ describe('setTimeout', function() {
...
   });
 
+  it('should allow cancelation by numeric timeout Id', function (done) {
+    // Node returns complex object from setTimeout, ignore this test.
+    if (isNode) {
+      done();
+      return;
+    }
+
+    var testZone = Zone.current.fork(Zone['wtfZoneSpec']).fork({ name: 'TestZone' });
```
vicb: const, const, const, let instead of var

lib/common/timers.ts
platosha: Done.

test/common/setTimeout.spec.ts
platosha: Done.

test/common/setTimeout.spec.ts
```diff
@@ -102,6 +102,27 @@ describe('setTimeout', function() {
...
+      return;
+    }
+
+    const testZone = Zone.current.fork(Zone['wtfZoneSpec']).fork({ name: 'TestZone' });
+    testZone.run(() => {
+      const spy = jasmine.createSpy('spy');
+      const task: Task = <any>setTimeout(spy, 0);
+      const cancelId: number = <any>task;
+      clearTimeout(0 + cancelId);
+      setTimeout(function () {
```
vicb: should you set a timeout of ie 10 here ? 
I don't think there is any guarantee on the order for the same timeout value (even if most impl should work as expected)

platosha: Right.

Done. Set the second timeout of 1.

Please note that the same point also applies to another test in this file as well, see https://github.com/platosha/zone.js/blob/02d0784bb9e9cfb44d2eaa32c12be538370b0a0b/test/common/setTimeout.spec.ts#L57

vicb: could you please also modify the other occurence ?

lib/common/timers.ts
```diff
@@ -60,6 +67,9 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
   clearNative =
       patchMethod(window, cancelName, (delegate: Function) => function(self: any, args: any[]) {
         var task: Task = args[0];
+        if (task && typeof task === 'number') {
+          task = tasksByHandleId[<number>task];
+        }
```
platosha: Done.

Can’t drop the `<number>` type casting in the `tasksByHandleId[<number>task]` index expression, though, the TypeScript compiler gives error.

test/common/setTimeout.spec.ts
platosha: Done, two places. Modified in a separate commit.

lib/common/timers.ts
```diff
@@ -19,14 +19,21 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
   setName += nameSuffix;
   cancelName += nameSuffix;
 
+  const tasksByHandleId: Object = {};
+
   function scheduleTask(task: Task) {
     const data = <TimerOptions>task.data;
-    data.args[0] = task.invoke;
+    data.args[0] = function() {
+      task.invoke.apply(this, arguments);
```
vicb: should `this` be `task` here ?
or `task.invoke(...arguments)` ?

lib/common/timers.ts
```diff
@@ -60,6 +67,9 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
   clearNative =
       patchMethod(window, cancelName, (delegate: Function) => function(self: any, args: any[]) {
         var task: Task = args[0];
+        if (typeof task === 'number') {
+          task = tasksByHandleId[<number>task];
```
vicb: remove `<number>` should be implied by the `typeof` check above

lib/common/timers.ts
platosha: No.

Before the change, `task.invoke` was used as a timer callback. Timer callbacks are executed with some other context: `this` is a `Timeout` instance in Node.js, `window` in browers, `undefined` in strict mode.

Within this change, we use another function as a timer callback, but in the function we still call `task.invoke` with the original timer callback context. As it was called before the change.

Note, that the `invoke` method of `ZoneTask`, which is called here, also passes the context further, so it ends as the original user-provided task callback context. Changing it is not good. Consider:

```javascript
setTimeout(function() {
  console.log(this);
  // should log `window` in browsers, not a `ZoneTask` instance or anything else
});
```

This doesn’t fail any Zone.js tests, though. Seems like there is a lack of tests for the timer callbacks context.

PR #463, opened 2016-09-20
fix(promise): include stack trace in an unhandlerd promise

mhevery:


test/common/Promise.spec.ts
```diff
@@ -244,13 +245,13 @@ describe('Promise', ifEnvSupports('Promise', function () {
         }).run(() => {
           zone = Zone.current;
           task = Zone.currentTask;
-          Promise.reject('rejectedErrorShouldBeHandled');
-          expect(promiseError).toBe(null);
+          error = new Error('rejectedErrorShouldBeHandled');
+          Promise.reject(error);
```
juliemr: Why remove the expectation that `promiseError` is null here?

juliemr:
Other than that question, LGTM.

test/common/Promise.spec.ts
mhevery: Opps, that was on accident, will add it back.

PR #467, opened 2016-09-28
Don't patch itself recursively

Freak613:
I'm using fetch polyfill (https://github.com/camsong/fetch-ie8).
I'm loading this polyfill before zone.js loading, to get fetch patched by zone.

In this block fetch has been called to patch its returning Promise realization.
It happens, that at the moment of the call, global Promise already has been patched by Zone, fetch returns ZoneAwarePromise as a result, and then zone patch its own promise which leads to recursive calls on the first attempt of using fetch.


Freak613:
I signed it! lol

mhevery:
Duplicate of #451

Issue #416, opened 2016-08-25
Error when running zone.js in a web worker in Safari browser


urimor:
The problem is in patchViaCapturingAllTheEvents() which is the workaround for patching on-properties in browsers other than Chrome & Firefox.

The patch is made by calling document.addEventListener() but in a web worker, the call to document causes error.

In our fork we fixed this (c359ce719ddaa73317cb40e78ba516fdab149837) by changing document.addEventListener() to self.addEventListener().

Thanks

RonGros:
Hi, any comment from Google on this? this actually means that NO Angular 2 application can run on Safari if it uses web workers! sounds like a SUPER SHOWSTOPPER, doesn't it?

nadavsinai:
also see -  (bd095bf) use of .hasOwnProperty - changed to use Object.getOwnPropertyNames - a different issue also affecting safari webworkers in version 0.6.21+

PR #469, opened 2016-09-28
Fix use of zone.js in safari webworkers

nadavsinai:
There's 2 problems fixed here :

-  (9eee113) use of document - changed to self
-  (bd095bf) use of .hasOwnProperty - changed to use Object.getOwnPropertyNames

Since this was done a while ago there's a couple of merge commits pulling changes from upstream, if you need to eliminate them I can rebase and re apply them.

However since the issue we opened for of the same problems (#416) did not receive any response from you for over a month I will spare myself from the extra work until proven useful to anyone else but us...

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -347,7 +347,7 @@ export function patchMethod(target: any, name: string,
                                       name: string) => (self: any, args: any[]) => any): Function
 {
   let proto = target;
-  while (proto && !proto.hasOwnProperty(name)) {
+  while (proto && Object.getOwnPropertyNames(proto).indexOf(name) === -1) {
     proto = Object.getPrototypeOf(proto);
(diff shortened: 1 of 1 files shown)


vicb:
please rebase on master

nadavsinai:
OK,will do, I will naturally omit the merge commits and  pick the actual relevant fixes, but should I also pick the build commit of the fixed 0.6.25 or should I leave it as is for you to do a build (probably with verbump) ?

vicb:
we do update built files on version bumps.
thanks !

nadavsinai:
done, now rebased on master.

nadavsinai:
I really don't understand the error in your CI, something that is completely unrelated to the changes made here.
locally the gulp build task works for me

PR #471, opened 2016-09-30
fix: stop using class extensd as it breaks rollup

mhevery:


lib/jasmine/jasmine.ts
```diff
@@ -35,43 +40,43 @@
     let originalJasmineFn: Function = jasmineEnv[methodName];
     jasmineEnv[methodName] = function(description: string, specDefinitions: Function) {
       return originalJasmineFn.call(this, description,  wrapDescribeInZone(specDefinitions));
-    } 
+    }
```
vicb: missing ; also below

lib/jasmine/jasmine.ts
```diff
@@ -35,43 +40,43 @@
...
   });
   ['it', 'xit', 'fit'].forEach((methodName) => {
     let originalJasmineFn: Function = jasmineEnv[methodName];
     jasmineEnv[methodName] = function(description: string, specDefinitions: Function, timeout: number) {
       arguments[1] = wrapTestInZone(specDefinitions);
       return originalJasmineFn.apply(this, arguments);
-    } 
+    }
```
vicb: here

lib/jasmine/jasmine.ts
mhevery: fixed

lib/jasmine/jasmine.ts
mhevery: fixed

PR #475, opened 2016-10-02
fix: fluent interface for EventEmitter

### LINKED ISSUE #470: PR #430 Breaks Stream interface in Node.js
PR #430  breaks fluent interface nature of **Node.js** Stream:

To reproduce, uncomment `require("zone.js")`. The code will crash because **createInterface** and **on** do not return _self_ which prevents cascading

``` js
//require("zone.js")
const fs = require("fs")

require('readline').createInterface({
  input: fs.createReadStream('/etc/hosts')
}).on('line', function (line) {
  console.log(line)
}).on('close', function () {
  console.log('done')
});
```

jakubmal:
Fixes #470 by returning self from `addListener`, `on`, `prependListener`, `removeListener` in Node.js.

### DIFF
--- a/lib/node/events.ts
+++ b/lib/node/events.ts
@@ -8,6 +8,12 @@
 
 import {makeZoneAwareAddListener, makeZoneAwareListeners, makeZoneAwareRemoveListener, patchMethod} from '../common/utils';
 
+const callAndReturnFirstParam = (fn: (self: any, args: any[]) => any) => {
+  return (self: any, args: any[]) => {
+    fn(self, args);
+    return self;
+  };
+};
 
 // For EventEmitter
 const EE_ADD_LISTENER = 'addListener';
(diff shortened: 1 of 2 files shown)


lib/node/events.ts
```diff
@@ -9,9 +9,9 @@ const EE_LISTENERS = 'listeners';
 const EE_ON = 'on';
 
 
-const zoneAwareAddListener = makeZoneAwareAddListener(EE_ADD_LISTENER, EE_REMOVE_LISTENER, false, true);
-const zoneAwarePrependListener = makeZoneAwareAddListener(EE_PREPEND_LISTENER, EE_REMOVE_LISTENER, false, true);
-const zoneAwareRemoveListener = makeZoneAwareRemoveListener(EE_REMOVE_LISTENER, false);
+const zoneAwareAddListener = makeZoneAwareAddListener(EE_ADD_LISTENER, EE_REMOVE_LISTENER, false, true, true);
```
vicb: IMO adding a "magic" boolean param is not the best way to to this.

Can you add `makeChainable...` that wraps the underlying method and return self ?

jakubmal: Sure, let me know what you think now!

lib/common/utils.ts
```diff
@@ -363,3 +363,9 @@ export function patchMethod(target: any, name: string,
   return delegate;
 }
 
+export function makeNamedFnChainable(fn: (self: any, args: any[]) => any) {
```
vicb: - pls move this fn where it is used (ie `events.ts`).
- use arrow fn,
- wouldn't `callAndReturnFirstParam(...)` be more explicit ?

percinor:
https://github.com/angular/universal/issues/568
I  have modified  zonejs23 or 25  util.ts and event.ts  same like  above ,but also error on ng2 universal project . if  i work express+angular2 +mysql ，its no problem , how to solve it , thanks

tamascsaba:
@vicb, jakubmal ??

jakubmal:
Sorry, I was short on time lately, I should be able get this ready this week!

lib/common/utils.ts
jakubmal: I would prefer named functions, because stack trace when debugging, but I've got nothing against good policies!

jakubmal:
All issues addressed! :)

jakubmal:
@alxhub ?

spenoir:
I'm hoping this PR will fix my `TypeError: Cannot read property 'on' of undefined` when using websockets in Angular 2. Has it been released?

JiaLiPassion:
@spenoir, yes this has been released.

PR #479, opened 2016-10-07
make it be array anyway

tflori:
In the rest of the class we act as we are sure this has to be array (`this._uncaughtPromiseErrors[0]`, `this._uncaughtPromiseErrors.length` usw). So it has to be an array even if something with zone is not like expected.

### DIFF
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -86,7 +86,7 @@
     private _microtasks: Function[] = [];
     private _lastError: Error = null;
     private _uncaughtPromiseErrors: {rejection: any}[] =
-        Promise[Zone['__symbol__']('uncaughtPromiseErrors')];
+        Promise[Zone['__symbol__']('uncaughtPromiseErrors')] || [];
 
     pendingPeriodicTimers: number[] = [];
     pendingTimers: number[] = [];


tflori:
restart plz :-/

lib/zone-spec/fake-async-test.ts
```diff
@@ -86,7 +86,7 @@
     private _microtasks: Function[] = [];
     private _lastError: Error = null;
     private _uncaughtPromiseErrors: {rejection: any}[] =
-        Promise[Zone['__symbol__']('uncaughtPromiseErrors')];
+        Promise[Zone['__symbol__']('uncaughtPromiseErrors')] || [];
```
mhevery: Could you provide a test showing how this could be undefined? I fear that while `|| []` solves the runtime issue it creates another issue which is that the errors will not be reported.

tflori: good point. I get an error in my tests where I mock the promise to be able to check the status..

hopefully I find time for this...

mhevery:
But if someone overwrites the global Promise then you have bigger problems
then just not being able to access the array. We should rather make the
Global Promise non-overwritable.

tflori:
Hm - thats weird. We overwrite the global Promise with a ZoneAwarePromise but if someone overwrites this Promise we have a Problem? Consider mozilla makes the Promise unwriteable like the navigator object. We can not install our ZoneAwarePromise.

Honestly I don't know where is the problem with another Promise implementation in our zone. Currently I'm just coping in the __zone_symbol__* attributes to my Promise. Why I have a own Promise implementation? Because I want to ask synchronously if the Promise has been resolved or rejected (it makes testing much easier). In the background I'm using the Promise implementation that was there before.

mhevery:
Why not use our fake-sync feature which can turn your async tests using
promises to sync.
https://github.com/angular/zone.js/blob/master/test/zone-spec/fake-async-test.spec.ts#L83
https://github.com/angular/angular/blob/14ee75924b6ae770115f7f260d720efa8bfb576a/modules/%40angular/examples/testing/ts/fake_async.ts#L14
https://github.com/angular/angular/blob/41c8c309730d36e46012f6461e8c94d84b13554d/modules/%40angular/core/test/animation/animation_player_spec.ts#L15

tflori:
Because it is still so complicated to ask this: `expect(Promise.resolve(true)).toHaveBeenResolvedWith(true);`

Currently I'm trying to create different Promise matchers that work with the ZoneAwarePromise but everything seems to be unrealistic complicated. (why a promise can not have a public status?)

But back to topic: yes It can be undefined and the code will throw an exception. Why not to prevent such wrong errors?

mhevery:
Our `ZoneAwarePromise` have `__symbol__('state')` and `__symbol__('value')` which you can use to read the current zone state. Does that help?

mhevery:
Closing as stale. Reopen as new PR if you can address my concerns

PR #483, opened 2016-10-11
[WIP]fix(patchProperty): fix when property is function with property

Brooooooklyn:
property patch may cause some other libary that use `function with property` as property.

see https://github.com/ReactiveX/rxjs/blob/master/src/observable/dom/AjaxObservable.ts#L357

and simple example:
http://plnkr.co/edit/oX1v8tTjTog1T57Z0JnI?p=preview

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -65,23 +65,43 @@ export function patchProperty(obj, prop) {
   // substr(2) cuz 'onclick' -> 'click', etc
   const eventName = prop.substr(2);
   const _prop = '_' + prop;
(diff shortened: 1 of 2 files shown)


Brooooooklyn:
Couldn't pass test in android and safari

Brooooooklyn:
I'm trying fix this issue in https://github.com/ReactiveX/rxjs/pull/2025

here is an example about this issue if you guys are interested

http://plnkr.co/edit/oX1v8tTjTog1T57Z0JnI?p=preview

PR #485, opened 2016-10-12
fix(node): crash when calling listeners() for event with no listeners

alxhub:


.gitignore
```diff
@@ -8,3 +8,4 @@ parsed-idl/
 .vscode
 npm-debug.log
 build-esm/
+/yarn.lock
```
mhevery: revert please

alxhub: Done

PR #486, opened 2016-10-12
docs(readme): add note about newrelic

gdi2290:
perhaps we should create a FAQ?


gdi2290:
newrelic does the same tracking for errors so they also patch Promise. Yes, anything that patches Promise should also be included before zone.js

JiaLiPassion:
async-listener and continuous-local-storage (which use the async-listener) should also load before zone.js

PR #430, opened 2016-09-07
feat(node): patch outgoing http requests to capture the zone

alxhub:

Issue #491, opened 2016-10-19
Error: Can't set headers after they are sent.


Hongbo-Miao:
I put the error log here, since it is a little bit long
http://pastie.org/10945816

The reproduce codes app is here
https://github.com/Hongbo-Miao/zonejs-bug

It uses universal-starter, just added **socket.io**, see in [second commit](https://github.com/Hongbo-Miao/zonejs-bug/commit/216a69e589c31f8d9f9157f3be0c57eae111433f).

To see the error, `npm start` and open page [http://localhost:3000/](http://localhost:3000/).

The app works with zone.js 0.6.21, and it shows the error starts from 0.6.22 and also shows in latest 0.6.26

JiaLiPassion:
I made a pull request to fix this issue, for now you can change your package.json to make zone.js reference to https://github.com/JiaLiPassion/zone.js.git to see the result.

MrCroft:
This still occurs in `0.8.20`.
I can't believe how old this issue is :)

My scenario, I have an SSR Response Service:
```
import { Injectable, Inject } from '@angular/core';

import { Response } from 'express';
import { RESPONSE } from '@nguniversal/express-engine';

@Injectable()
export class SsrResponseService {
  constructor(@Inject(RESPONSE) private response: Response) {}

  redirect(url: string, code?: number) {
    if (!!code) {
      return this.response.redirect(code, url);
    } else {
      return this.response.redirect(url);
    }
  }
}
```

Then, in a component, based on some logic, I might need to redirect the user:
```
this.responseService.redirect('/some-other-url', 301);
```

The redirect works, but the server logs:
```
Unhandled Promise rejection: Can't set headers after they are sent. ; 
Zone: <root> ; Task: Promise.then ; Value: Error: Can't set headers after they are sent.
```

Full stack trace [here](https://gist.github.com/MrCroft/c8a659567a3b248744b62a7cc04f061d)

Any ETA on this fix?

mike-suggitt:
Bump - same here

JiaLiPassion:
@mike-suggitt , @MrCroft , I don't think the `Can't set headers after they are sent` is `zone.js` error, it is just `express http response` error.

JiaLiPassion:
@MrCroft , sorry I am not familiar with `SSR`, you can ask at `stackoverflow`.

MrCroft:
[...]

Issue #484, opened 2016-10-12
Zone.js higher than 0.6.21 breaks angular-universal https calls


AlexYankee:
Calls to https:// end up with error code: ECONNRESET, as described in https://github.com/angular/universal-starter/issues/172

JiaLiPassion:
This is because the EventEmitter.once is not handled properly, I have create a pull request, you can use my repository for test, https://github.com/JiaLiPassion/zone.js.git.

John0x:
Any updates on this?

JiaLiPassion:
This has been fixed, you can use the latest zone.js 0.7.1 to test.

PR #500, opened 2016-11-04
Fix EventEmitter.once issue and add EventEmitter removeAllListeners patch.

JiaLiPassion:
I found several issues of patching EventEmitter's listener related methods.

with zone.js 0.6.22 release, addListener, prependListener, removeListener are patched in this commit.
https://github.com/angular/zone.js/commit/100b82bbac4d720be0cb7ead11398e7fe95d1132

1. issue #484 EventEmitter.once was not removed properly.
the issue can be described as the following test case.

` emitter.once('test', shouldNotRun);
   emitter.removeListener('test', shouldNotRun);`

in 0.6.22, the once listener was not removed.

the reason is when removeListener call util.ts, findExistingRegisteredTask method, it will look for the correct event task by check the eventTask.data.handler === handler, but if the handler is attached by EventEmitter.once method, the handler will not attached directly, instead it will be attached through a _onceWrap (https://github.com/nodejs/node/blob/master/lib/events.js#L286), and keep the original handler to the onceWrap.listener property, so we should not only check the data.handler but also the data.handler.listener(https://github.com/angular/zone.js/pull/500/commits/2ee4711e96cad3b27bce79ea9290800c929986f1#diff-2e7dc744ee86a31cde46d41cda5e42c0R714)

2. issue #491 removeListeners was not patched
the issue can be described as the following test case.
`emitter.on('test', expectZoneA);
  emitter.on('test', expectZoneA);
  emitter.removeAllListeners('test');`

in 0.6.22, the listener was not removed.

the reason is based on the issue #491,  in the  zone.js 0.6.22 release,  to fix #430,  addListener, removeListener, prependListener, and listeners have been patched, but removeAllListeners is not, In issue #491, socket-io will call removeAllListeners to cache and remove the listeners on request event when attach server, but because the removeAllListeners is not patched and handled correctly by zone(the response listener was added to event tasks because the addListener is patched but not removed because removeAllListeners are not patched), the request event will be handled multiple times and cause express response throw "can't set header after response sent " error.
I add a patch of removeAllListeners based on the #430 release.

3. when add listener by prependListener, the listener should be added at the beginning of the listener array.
the issue can be described as the following test case.

`emitter.on('test', listenerA);
  emitter.prependListener('test', listenerB);
  expect(emitter.listeners('test')).toEqual([listenerB, listenerA]);`

in 0.6.22, the order is still listenerA, listenerB

The reason is in 0.6.22 prependListener do the same logic with addListener.

4. use the newest fetch to fix Object #<Response> has no method 'blob' error

 please review the source code of those 2 issues, Thank you.


jakubmal:
Nice! I might be mistaken, but you should not commit dist versions, they are updated, when publishing a new version, right?

JiaLiPassion:
@jakubmal  Yes, you are right, I will modify my Pull Request. Thank you.

JiaLiPassion:
I modify the package.json to use the newest fetch to avoid travis CI build error(Object #<Response> has no method 'blob').

package.json
```diff
@@ -64,6 +64,6 @@
     "ts-loader": "^0.6.0",
     "tslint": "^3.15.1",
     "typescript": "^2.0.2",
-    "whatwg-fetch": "^1.0.0"
+    "whatwg-fetch": "https://github.com/jimmywarting/fetch.git"
```
mhevery: Why is this change needed? This does not seem right.

PR #507, opened 2016-11-18
Better handling of stack traces in Zones

mhevery:
feat(Error): Rewrite Error stack frames to include zone

The errors caused by the Zone contain extra stack frames. This feature
removes the confusing stack frames as well as appends the zone for each
stack frame to get a better understanding of the error messages.

Improved Error:
```
Error.spec.js:54 Error: Inside [InnerZone]
    at insideRun (Error.spec.js:31) [InnerZone]
    at Zone.run (zone.js:100) [<root> => InnerZone]
    at testFn (Error.spec.js:29) [<root>]
    at Zone.run (zone.js:100) [ProxyZone => <root>]
    at Object.eval (Error.spec.js:19) [ProxyZone]
```

Original Error:
```
Error.spec.js:53 Error: Inside
    at ZoneAwareError (zone.js:652)
    at insideRun (Error.spec.js:31)
    at ZoneDelegate.invoke (zone.js:216)
    at Zone.run (zone.js:100)
    at testFn (Error.spec.js:29)
    at ZoneDelegate.invoke (zone.js:216)
    at Zone.run (zone.js:100)
    at Object.eval (Error.spec.js:19)
```

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -519,6 +519,21 @@ interface EventTask extends Task {
   /* TS v1.8 => type: 'eventTask'; */
 }
 
+/**
+ * Extend the Error with additional fields for rewritten stack frames
+ */
+interface Error {
+  /**
+   * Stack trace where extra frames have been removed and zone names added.
+   */
+  zoneAwareStack?: string;
+
+  /**
+   * Original stack trace with no modiffications
+   */
+  originalStack?: string;
+}
+
 /** @internal */
 type AmbientZone = Zone;
 /** @internal */
@@ -545,7 +560,7 @@ const Zone: ZoneType = (function(global: any) {
 
 
     static get current(): AmbientZone {
-      return _currentZone;
+      return _currentZoneFrame.zone;
     };
     static get currentTask(): Task {
       return _currentTask;
@@ -606,19 +621,17 @@ const Zone: ZoneType = (function(global: any) {
 
     public run(
         callback: Function, applyThis: any = null, applyArgs: any[] = null, source: string = null) {
-      const oldZone = _currentZone;
-      _currentZone = this;
+      _currentZoneFrame = new ZoneFrame(_currentZoneFrame, this);
       try {
         return this._zoneDelegate.invoke(this, callback, applyThis, applyArgs, source);
       } finally {
-        _currentZone = oldZone;
+        _currentZoneFrame = _currentZoneFrame.parent;
       }
     }
 
     public runGuarded(
         callback: Function, applyThis: any = null, applyArgs: any[] = null, source: string = null) {
-      const oldZone = _currentZone;
-      _currentZone = this;
+      _currentZoneFrame = new ZoneFrame(_currentZoneFrame, this);
       try {
         try {
           return this._zoneDelegate.invoke(this, callback, applyThis, applyArgs, source);
@@ -628,7 +641,7 @@ const Zone: ZoneType = (function(global: any) {
           }
         }
       } finally {
-        _currentZone = oldZone;
+        _currentZoneFrame = _currentZoneFrame.parent;
       }
     }
 
@@ -641,8 +654,7 @@ const Zone: ZoneType = (function(global: any) {
             '; Execution: ' + this.name + ')');
       const previousTask = _currentTask;
       _currentTask = task;
-      const oldZone = _currentZone;
-      _currentZone = this;
+      _currentZoneFrame = new ZoneFrame(_currentZoneFrame, this);
       try {
         if (task.type == 'macroTask' && task.data && !task.data.isPeriodic) {
           task.cancelFn = null;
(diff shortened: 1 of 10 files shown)


lib/zone.ts
```diff
@@ -1232,5 +1253,151 @@ const Zone: ZoneType = (function(global: any) {
...
+   */
+  function ZoneAwareError() {
+    // Create an Error.
+    let error: Error = NativeError.apply(this, arguments);
+
+    // Save original stack trace
+    error.originalStack = error.stack || '';
+
+    // Process the stack trace and rewrite the frames.
+    if (ZoneAwareError['rewriteStack']) {
```
vsavkin: Since generating errors isn't a performance concern (we don't throw millions of errors), have you considered just creating a new stack instead of modifying the existing one in place?  The code will be easier for follow.

lib/zone.ts
```diff
@@ -1232,5 +1253,151 @@ const Zone: ZoneType = (function(global: any) {
...
+  /**
+   * This is ZoneAwareError which processes the stack frame and cleans up extra frames as well as
+   * adds zone information to it.
+   */
+  function ZoneAwareError() {
+    // Create an Error.
+    let error: Error = NativeError.apply(this, arguments);
+
+    // Save original stack trace
+    error.originalStack = error.stack || '';
```
vsavkin: Why `|| ''`?  Seems odd that the original stack is actually not the original stack :) ?

lib/zone.ts
```diff
@@ -1232,5 +1253,151 @@ const Zone: ZoneType = (function(global: any) {
...
+            runTaskFrame = true;
+          } else if (fnName.indexOf('run') !== -1) {
+            runFrame = true;
+          } else {
+            frameType = FrameType.blacklist;
+          }
+          blacklistedStackFrames[frame] = frameType;
+          // Once we find all of the frames we can stop looking.
+          if (runFrame && runGuardedFrame && runTaskFrame) {
+            ZoneAwareError['rewriteStack'] = true;
```
vsavkin: Seems odd that we are setting a global here. It will never be reset to false. Consider the following three errors:
`A B A'`, where A and A' are essentially the same error triggered in the same place in the code.  And B is the one that triggers this onHandleError. 

Because of this global A and A' will be handled differently, but they should not.

lib/zone.ts
```diff
@@ -519,6 +519,21 @@ interface EventTask extends Task {
...
+interface Error {
+  /**
+   * Stack trace where extra frames have been removed and zone names added.
+   */
+  zoneAwareStack?: string;
+
+  /**
+   * Original stack trace with no modiffications
+   */
+  originalStack?: string;
```
mhevery: It is the original Error.stack before it is rewritten to include zone information.

lib/zone.ts
```diff
@@ -1232,5 +1253,151 @@ const Zone: ZoneType = (function(global: any) {
...
+
+  /*
+   * This code patches Error so that:
+   *   - It ignores un-needed stack frames.
+   *   - It Shows the associated Zone for reach frame.
+   */
+
+  enum FrameType {
+    /// Skip this frame when printing out stack
+    blacklist,
```
mhevery: fixed

lib/zone.ts
mhevery: fixed

lib/zone.ts
mhevery: I am not sure how one could do that? Can you tell me what you have in mind?

lib/zone.ts
mhevery: This code is part of the detection, which tries to answer if we can rewrite stack on this platform. It is simply an indicator if a given platform supports it. For the same platform it will never be different.

lib/zone.ts
```diff
@@ -1232,5 +1253,151 @@ const Zone: ZoneType = (function(global: any) {
...
+        if (error.stack) {
+          Error['stackRewrite'] = true;
+        }
+        throw error;
+      });
+    });
+  };
+  // Cause the error to extract the stack frames.
+  detectZone.runTask(detectZone.scheduleMacroTask('detect', detectRunFn, null, () => null, null));
+  console.log(blacklistedStackFrames);
```
mhevery: removed

test/common/Error.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+describe('ZoneAwareError', () => {
+  // If the environment does not supports stack rewrites, then these tests will fail
+  // and there is no point in running them.
+  if (!Error['stackRewrite']) return;
+
+  it('should show zone names in stack frames and remove extra frames', () => {
+    var rootZone = Zone.current;
```
mhevery: done

test/common/Error.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    while(rootZone.parent) {
+      rootZone = rootZone.parent;
+    }
+
+    var innerZone = rootZone.fork({name: 'InnerZone'});
```
mhevery: done

vsavkin:
LGTM

PR #514, opened 2016-11-22
fix: Run tests on both the build as well as the dist folder

mhevery:


.travis.yml
```diff
@@ -22,6 +22,7 @@ before_script:
   - ./scripts/sauce/sauce_connect_block.sh
 
 script:
-  - node_modules/.bin/karma start karma-sauce-jasmine.conf.js --single-run
-  - node_modules/.bin/karma start karma-sauce-mocha.conf.js --single-run
+  - node_modules/.bin/gulp build
+  - node_modules/.bin/karma start karma-dist-sauce-jasmine.conf.js --single-run
+  - node_modules/.bin/karma start karma-build-sauce-mocha.conf.js --single-run
```
IgorMinar: wouldn't it be safer to run both mocha and jasmine in both build and dist mode?

the tests don't take much time to run, so what's the downside?

IgorMinar:
otherwise lgtm

Issue #515, opened 2016-11-22
0.7.1 Breaking Error


antonyRoberts:
Upgrading to zone 0.7.1 from 0.6.26, with @JiaLiPassion zone-node file, breaks app on start with 

TypeError: Error.captureStackTrace is not a function.
>nodemon] 1.11.0
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: C:\apps\UniversalMaterialBreak\dist/**/* src/index.html
[nodemon] starting `node dist/server/index.js`
C:\apps\UniversalMaterialBreak\node_modules\depd\lib\compat\index.js:39
  Error.captureStackTrace(obj)
        ^

>TypeError: Error.captureStackTrace is not a function
    at callSiteToString (C:\apps\UniversalMaterialBreak\node_modules\depd\lib\compat\index.js:39:9)
    at Object.get [as callSiteToString] (C:\apps\UniversalMaterialBreak\node_modules\depd\lib\compat\index.js:60:15)
[nodemon] app crashed - waiting for file changes before starting...

[Repo showing issue.](https://github.com/antonyRoberts/UniversalMaterialBreak)

JiaLiPassion:
[zone-node.js.zip](https://github.com/angular/zone.js/files/608041/zone-node.js.zip)

it seems ZoneAwareError not inherit captureStackTrace method, I made a pull request #516 to fix this one. you can use the attched zone-node.js for test.

antonyRoberts:
After your patch getting 
>C:\apps\UniversalMaterialBreak\node_modules\depd\index.js:252
  var file = callSite.getFileName() || '<anonymous>'
                      ^
>TypeError: callSite.getFileName is not a function
    at callSiteLocation (C:\apps\UniversalMaterialBreak\node_modules\depd\index.js:252:23)
    at depd (C:\apps\UniversalMaterialBreak\node_modules\depd\index.js:112:14)
    at Object.<anonymous> (C:\apps\UniversalMaterialBreak\node_modules\express\lib\router\index.js:21:32)
    at Module._compile (module.js:573:32)
    at Object.Module._extensions..js (module.js:582:10)
    at Module.load (module.js:490:32)
    at tryModuleLoad (module.js:449:12)
    at Function.Module._load (module.js:441:3)
    at Module.require (module.js:500:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (C:\apps\UniversalMaterialBreak\node_modules\express\lib\application.js:17:14)
[...]

Issue #505, opened 2016-11-16
Patch process.nextTick()


sjelin:
node's [`process.nextTick()`](https://nodejs.org/api/process.html#process_process_nexttick_callback_args) is used by a lot of libraries (`q`, for instance) but isn't patched by zone.js:

```js
require('zone.js');
Zone.current.fork({}).run(() => {
  Zone.current.foo = 'bar';
  process.nextTick(() => {
    console.log(Zone.current.foo); // undefined
  });
})
```

`process.nextTick` was basically `setImmediate` before `setImmediate` existed.  Now it's like `setImmediate` except it gets scheduled in a higher priority queue.

JiaLiPassion:
Yeah, I agree with you, zone.js should patch process.nextTick, process.nextTick is a microtask , if we patch it like 
`var nativeNextTick = process.nextTick;
  process.nextTick = function() {
    var args = arguments;
    args[0] = Zone.current.wrap(args[0], 'process.nextTick');
    nativeNextTick.apply(this, args);
 }`

the callback will be inzone.
and the nextTick will run before setImmediate/setTimeout but after promise.resolve, because ZoneAwarePromise will not use v8 native microtask queue, should zone.js also push process.nextTick into microtask queue of zone?

sjelin:
I don't think wrapping will effect the order at all?  `Zone.wrap` doesn't make a function asynchronous, and the callback is being scheduled by `process.nextTick` in either case.  But I think your patch is fine regardless.

JiaLiPassion:
yes, the zone.wrap will not change the order, what I want to say is patch it just like the code I posted  can only resolve the issue that process.nextTick callback not inzone, but it will still be executed after promise.resolve().  such behaviors are different with the native implementations.

sjelin:
After reading more about micro tasks, I agree.

sjelin:
I think we need another task queue, sadly

JiaLiPassion:
I believe in nodejs native implementation process.nextTick and promise.resolve use the same microtask queue, so maybe we can patch process.nextTick using the ZoneAwarenessPromise microTaskQueue...

sjelin:
Nope:
```js
var p = (s) => { Promise.resolve().then(() => { console.log(s); }); }
var n = (s) => { process.nextTick(() => { console.log(s); }); }

n(1);
p(3);
n(2);
p(4);
```
Outputs:
```
1
2
3
4
```
On node v6.9.1

[...]

PR #516, opened 2016-11-23
Patch captureStackTrace/prepareStackTrace to ZoneAwareError, patch process.nextTick, fix removeAllListeners bug

JiaLiPassion:
1. ZoneAwareError didn't add captureStackTrace method, so it will failed when user call Error.captureStackTrace. for example issue #515 

2. ZoneAwareError didn't add Error.prepareStackTrace callback, so if user add prepareStackTrace callback to Error, the callback will not be executed. for example issue #515 

https://github.com/v8/v8/wiki/Stack-Trace-API

3. patch process.nextTick with zone.scheduleMicroTask, so process.nextTick can be in zone, and can be handled by ZoneDelegate task related callback(onScheduleTask/onInvokeTask) #505 

4. modify a typo, FrameType.trasition->FrameType.transition

5. fix EventEmitter.removeAllListeners bug #518, when call EventEmitter.removeAllListeners without eventName, should remove all eventListener, and add some test cases for EventEmitter.removeListener/newListener.

6. update fetch to 2.0.1 in package.json


lib/node/node.ts
```diff
@@ -30,6 +30,13 @@ if (shouldPatchGlobalTimers) {
   patchTimer(_global, set, clear, 'Immediate');
 }
 
+// patch process.nextTick
+var nativeNextTick = process.nextTick;
```
Brooooooklyn: may use `const` better

JiaLiPassion: yes, you are right, thank you.

lib/common/utils.ts
```diff
@@ -280,9 +280,15 @@ export function makeZoneAwareRemoveAllListeners(fnName: string, useCapturingPara
   const defaultUseCapturing = useCapturingParam ? false : undefined;
 
   return function zoneAwareRemoveAllListener(self: any, args: any[]) {
+    var target = self || _global;
+    if (args.length === 0) {
+      // remove all listeners without eventName
+      target[EVENT_TASKS] = [];
+      target[symbol]();
+      return;
```
KrauseStefan: The real `removeAllListeners(...)`
return `this`
This wrapper should do the same
https://github.com/nodejs/node/blob/master/lib/events.js#L404

lib/common/utils.ts
```diff
@@ -280,9 +280,15 @@ export function makeZoneAwareRemoveAllListeners(fnName: string, useCapturingPara
...
+    }
     var eventName = args[0];
     var useCapturing = args[1] || defaultUseCapturing;
-    var target = self || _global;
     var eventTasks = findAllExistingRegisteredTasks(target, eventName, useCapturing, true);
```
KrauseStefan: When all all listeners are removed shouldn't all the eventTasks also be chanceled?
I believe we need to look into how to use `findAllExistingRegisteredTasks` to get and cancel all eventTasks.

lib/common/utils.ts
KrauseStefan: In typescript I believe you probably want to use either `let` or `const` as far as I can tell all except the logic in the loop can use `const`.

lib/common/utils.ts
JiaLiPassion: the cancelTask will just call the original EventEmitter.removeListener,  I think we don't need to cancel it, because when we call target[symbol]() https://github.com/angular/zone.js/pull/516/files#diff-f5e084bcb7c43d977d56c3b791bbca2dR287, the EventEmitter.removeAllListeners will do it for us.

and the cancelTask call here is also not necessary either I think, https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L290

nakedcity:
When are you planning to  merge this?

lib/node/node.ts
```diff
@@ -83,3 +85,28 @@ if (httpClient && httpClient.ClientRequest) {
...
+
+function patchNextTick() {
+  let setNative = null;
+
+  function scheduleTask(task: Task) {
+    const args = task.data;
+    args[0] = function() {
+      task.invoke.apply(this, arguments);
+    };
+    setNative.apply(process, args);
```
sjelin: Shouldn't this be `setNative.apply(self, args)` using the `self` variable coming from `patchMethod`?  Not actually a problem since `nextTick` appears to bind callbacks to `global`, but still

lib/node/node.ts
```diff
@@ -83,3 +85,28 @@ if (httpClient && httpClient.ClientRequest) {
     }
   };
 }
+
+function patchNextTick() {
+  let setNative = null;
```
sjelin: It seems weird to use this local variable instead of moving `scheduleTask` inside of `patchMethod`, but I guess you're trying to limit nested closures?

JiaLiPassion: yeah, I agree, I just keep the coding style to be the same with timer patch, https://github.com/angular/zone.js/blob/master/lib/common/timers.ts#L17

lib/node/node.ts
sjelin: Do we know it's `process`?

```js
var nextTick = process.nextTick;
nextTick.call(myObject, function() {
  console.log(myObject); // If nextTick didn't always bind to `global`,
                         // this would be `myObject`.  However, with the 
                         // zone.js patch, this would be `process`
});
```

sjelin: (I edited my code in the above comment, since it used to make no sense)

JiaLiPassion: Yeah, but I can't imagine when we need to call like nextTick.call(myObject, ...), in my understanding we should always call nextTick.call(process, ...)

sjelin: Anyway, I'd prefer to see it the other way, but it honestly doesn't matter

JiaLiPassion: > But it is allowable to use a different context, and so this could be a problem in the future. With setTimeout the DOM spec requires you to call it on an object of type Window or WorkerGlobalScope (i.e. the global scope)

I didn't know such kind of usage, thank you for the information. in this PR, I'll just keep it this way just to make the coding style be the same with patchTimer. Thanks again.

PR #517, opened 2016-11-24
docs(changelog): use conventional changelog

ocombe:
I added conventional changelog like you have on angular/angular, I added a gulp script and also a npm script for it.
I haven't hooked it up in the release process because I don't know how you want to do it (manually before a new release?).

I also fixed the install / gulp tsc so that it works on windows (I applied the same fix that you have on angular/angular).

And I regenerated the complete changelog.


mhevery:
Thank you.

PR #521, opened 2016-11-25
feat(Zone): check global for rootZoneSpec

### LINKED ISSUE #513: Configuring Root zone
Is there a way to configure `onScheduleTask` and `onInvokeTask` on the root zone?
Or other interceptions for that matter?

jahtalab:
closes #513

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -952,7 +952,13 @@ const Zone: ZoneType = (function(global: any) {
   const symbolPromise = __symbol__('Promise');
   const symbolThen = __symbol__('then');
 
-  let _currentZoneFrame = new ZoneFrame(null, new Zone(null, null));
+  const symbolRootZoneSpec = '__rootZoneSpec__'
+  var rootZone: Zone = new Zone(null, null);
+  if (global[symbolRootZoneSpec]) {
+    rootZone = <Zone>rootZone.fork(global[symbolRootZoneSpec]);
+  }
+
+  let _currentZoneFrame = new ZoneFrame(null, rootZone);
   let _currentTask: Task = null;
   let _microTaskQueue: Task[] = [];
   let _isDrainingMicrotaskQueue: boolean = false;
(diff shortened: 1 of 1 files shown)


JiaLiPassion:
As described in comment of #513,  I am not sure my thinking is correct or not, but I think we shouldn't allow access ZoneSpec at any level.

lib/zone.ts
```diff
@@ -952,7 +952,14 @@ const Zone: ZoneType = (function(global: any) {
   const symbolPromise = __symbol__('Promise');
   const symbolThen = __symbol__('then');
 
-  let _currentZoneFrame = new ZoneFrame(null, new Zone(null, null));
+  const symbolRootZoneSpec = '__rootZoneSpec__'
+  var rootZone: Zone = new Zone(null, null);
```
JiaLiPassion: We should use let instead of var

lib/zone.ts
```diff
@@ -952,7 +952,13 @@ const Zone: ZoneType = (function(global: any) {
...
+  if (global[symbolRootZoneSpec]) {
+    rootZone = <Zone>rootZone.fork(global[symbolRootZoneSpec]);
```
jahtalab: I think we should so that we don't clutter global space.
I can't see any reason not to.

jahtalab:
Could you please rerun the tests on travis?
I'm not sure what is causing the failure.

JiaLiPassion:
Sorry, I have no permission to rerun the test on travis, the error seems to be timeout, it should be ok to run again, in your PR, a ';' is missing in line 955, maybe you can commit the ';' and it will trigger a new travis test run.

JiaLiPassion:
It seems to be the fetch issue, I got the same problem before, I'll check it later.

jahtalab:
just rebased, tests are passing now. can we merge this one?

mhevery:
Sorry, but this is by design. RootZone should be indistinguishable for no zone at all. This requirement is here so that if we standardize the Zones in TC39, the browser vendors will not have to pay the cost for all of the applications. All zones must be entered explicitly.

Issue #522, opened 2016-11-26
incorrect environment detection in nw.js


adros:
Module `lib/common/utils.ts` incorrectly sets variable `isBrowser` to false when used in nw.js (because `isNode==true` because global variable `process` exists).

This causes that WebSockets are not correctly patched (and possibly other problems as well).

PR #523, opened 2016-11-26
isBrowser is now correctly detected (as true) in nw.js environment

adros:
fixex #522


JiaLiPassion:
With this PR, in nw.js env, isNode and isBrowser will all be true?  so the code here will still not be processed as browser? https://github.com/angular/zone.js/blob/master/lib/browser/property-descriptor.ts#L18

and I made the same mistake before, you should not commit "dist" folder cause it is for release. you can check here,https://github.com/angular/zone.js/blob/master/DEVELOPER.md

adros:
@JiaLiPassion you're right. Property `isNode` should be false to get it work.
I'll close this PR and create new one with correct behaviour (and without changes in dist)
Thank you

adros:
signed it!

2016-11-26 11:46 GMT+01:00 googlebot <notifications@github.com>:

PR #524, opened 2016-11-26
'isNode' will now be false in nw.js environment

### LINKED ISSUE #522: incorrect environment detection in nw.js

adros:
so 'isBrowser' will be true and 'propertyDescriptorPatch' will be executed correctly

closes #522


JiaLiPassion:
sorry, my earlier suggestion is not corrent,  in nw.js or electron, isNode and isBrowser should both be true, so the node related api and browser method can both be patched.

PR #531, opened 2016-12-08
Make sure there is no inner fn decl

vicb:
/ref https://github.com/angular/angular/issues/13311

### DIFF
--- a/test/main.ts
+++ b/test/main.ts
@@ -7,27 +7,29 @@
  */
 
 declare var __karma__: {
-  loaded: Function, start: Function
-  error: Function
+  loaded: Function,
+  start: Function,
+  error: Function,
 };
 
-__karma__.loaded = function () { };
+__karma__.loaded = function() {};
 (window as any).global = window;
 
-System.config({ defaultJSExtensions: true });
+System.config({defaultJSExtensions: true});
 
 System.import('/base/build/test/browser-zone-setup').then(() => {
-  let testFrameworkPatch = typeof (window as any).Mocha !== 'undefined'
-      ? '/base/build/test/test-env-setup-mocha' : '/base/build/test/test-env-setup-jasmine';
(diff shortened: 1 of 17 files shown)


test/ws-webworker-context.ts
```diff
@@ -8,6 +8,6 @@
 
 declare function importScripts(path: string): void;
 
-importScripts('/base/build/lib/zone.js');
-importScripts('/base/node_modules/systemjs/dist/system.src.js');
-importScripts('/base/build/test/zone_worker_entry_point.js');
+    importScripts('/base/build/lib/zone.js');
```
mhevery: this indentation does not seem right

test/mocha-patch.spec.ts
```diff
@@ -1,17 +1,17 @@
...
-declare function setup(fn: () => void): void;
-declare function teardown(fn: () => void): void;
-declare function suiteSetup(fn: () => void): void;
-declare function suiteTeardown(fn: () => void): void;
-declare function before(fn: () => void): void;
-declare function after(fn: () => void): void;
-//
-
-import {ifEnvSupports} from './test-util';
+    declare function test(description: string, testFn: () => void): void;
```
mhevery: Indentation does not seem right here

test/main.ts
```diff
@@ -6,28 +6,28 @@
...
 
-declare var __karma__: {
-  loaded: Function, start: Function
+declare var __karma__: {loaded: Function, start: Function
   error: Function
 };
 
-__karma__.loaded = function () { };
-(window as any).global = window;
+  __karma__.loaded = function() {};
```
mhevery: Indentation here does not look right.

mhevery:
Could we have the tslint run as part of the travis that way we don't go out of sync again?

test/ws-webworker-context.ts
vicb: clang bug, we have to live with this for now :(

Issue #538, opened 2016-12-12
Wrong currentZone value inside of onInvoke hook


davidich:
Hi, there.

According to comments I found in zone.ts file, onInvoke hook has to receive currentZone as second parameter:

`/**
   * Allows interception of the callback invocation.
   *
   * @param parentZoneDelegate Delegate which performs the parent [ZoneSpec] operation.
   * @param currentZone The current [Zone] where the current interceptor has beed declared.
   * @param targetZone The [Zone] which originally received the request.
   * @param delegate The argument passed into the `run` method.
   * @param applyThis The argument passed into the `run` method.
   * @param applyArgs The argument passed into the `run` method.
   * @param source The argument passed into the `run` method.
   */
  onInvoke?:
      (parentZoneDelegate: ZoneDelegate, currentZone: Zone, targetZone: Zone, delegate: Function,
       applyThis: any, applyArgs: any[], source: string) => any;`

I created an interceptZone, which I suppose should be passed as current zone to my hook method. But for some reason I receive the same value as targetZone.
I created a demo [HERE](http://plnkr.co/edit/iwdW5x3EURKopuwtiX4s?p=preview), which depicts the described problem.

Am I missing something?

Thanks,
Oleksiy

JiaLiPassion:
It seems a bug when currentZoneSpec has not onInvoke callback or maybe zone.js designed that way, I'll confirm it, thanks for posting the issue.

JiaLiPassion:
close for fixed

PR #539, opened 2016-12-12
fix issue 538, currentZone in all hook methods will not be correct if…

JiaLiPassion:
based on issue #538, all hook methods will pass wrong currentZone if the zonespec has no hook method but the parent Zone spec has.

the issue can be described as following case.

```javascript
 var zone = Zone.current;
      var zoneA = zone.fork({
        name: 'A',
        onInvoke: function(parentDelegate, currentZone, targetZone, callback, applyThis, applyArgs, source) {
          expect(currentZone.name).toEqual('A');
          return parentDelegate.invoke(targetZone, callback, applyThis, applyArgs, source);
        }
      });
      var zoneB = zoneA.fork({
        name: 'B',
        onInvoke: function(parentDelegate, currentZone, targetZone, callback, applyThis, applyArgs, source) {
          expect(currentZone.name).toEqual('B');
          return parentDelegate.invoke(targetZone, callback, applyThis, applyArgs, source);
        }
      });
      var zoneC = zoneB.fork({
        name: 'C'
      });
      zoneC.run(function() {
      });
```

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -715,27 +715,35 @@ const Zone: ZoneType = (function(global: any) {
 
     private _forkDlgt: ZoneDelegate;
     private _forkZS: ZoneSpec;
+    private _forkCurrZone: Zone;
 
     private _interceptDlgt: ZoneDelegate;
     private _interceptZS: ZoneSpec;
+    private _interceptCurrZone: Zone;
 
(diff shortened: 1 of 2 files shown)


mhevery:
Very nice work!

Issue #540, opened 2016-12-13
The "typeof element.oninput" changes with Zone.js (object vs. function)


mkdynamic:
Test case to reproduce:

```html
<!DOCTYPE html>
<html>
  <body>
    <script>
      var test = function() {
        var input = document.createElement("input");
        input.setAttribute("oninput", "return");
        return typeof input.oninput;
      };

      document.write("<div>before zone.js: " + test() + "</div>");
    </script>

    <script src="https://cdnjs.cloudflare.com/ajax/libs/zone.js/0.7.2/zone.min.js"></script>

    <script>
      document.write("<div>after zone.js: " + test() + "</div>");
    </script>
  </body>
</html>
```

This outputs the following:

```
before zone.js: function
after zone.js: object
```

This breaks Modernizr’s feature detection for "oninput" (it assumed a function type):
https://github.com/Modernizr/Modernizr/blob/master/feature-detects/event/oninput.js#L29

JiaLiPassion:
after zone.js, the function become object because it will be null.
I found the same issue in #525, and I found a solution recently, I'll make a PR to fix it.

PR #541, opened 2016-12-13
fix inline event handler issue #525 #540

### LINKED ISSUE #525: Zone.js patch DOM on-event handler will failed with inline event handler.
I will describe the problem with the sample code 

```javascript 
<html>
<head>
</head>
<script src="./zone.js"> </script>
<script>
    function loaded() {
        var btn = document.getElementById('btn');
        console.log('btn click event handler', btn.onclick);
    }

    function btnClick(e) {
        console.log('clicked', e);
    }

    function blur() {
        console.log('blur', e);
    }
</script>
<body onload="loaded();">
    <input id="txt" type="text" onblur="blur();"/>
    <button type="button" id="btn" onclick="btnClick(event);">OK</button>
</body>
</html>
```

after zone.js patched, the button's onclick will be null, if without zone.js patched, the onclick will be btnClick, the problem only occurs when I use inline event handler like above, if I use 

```javascript
    btn.onclick = btnClick;
```
or
```javascript
   btn.addEventListener(btnClick);
 ```

the result will be ok.

It seems not the problem of Zone.js itself but the DOM defineProperty.
https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L69
it seems after definePRoperty, the set will not be called when we use inline event handler.

### PR DESCRIPTION
**Pull request by JiaLiPassion**
fix #525 and #540, inline event handler will not be patched with zone. And when you want to get the inline event handler will return null.

```javascript
<button id="btn1" onclick="func()">ok</button>
```

The func will not be patched, and if you access document.getElementById('btn1').onclick will return null.

The PR will not totally fixed the issue, if you don't access the elem.onclick for example, still no patch(but not null, you will get the original native version handler). you should add the following code to patch if you want to handle inline event handler

```javascript
          var all = document.getElementsByTagName("*");

          for (var i=0, max=all.length; i < max; i++) {
           var elem = all[i];
           for (var j = 0; j < eventNames.length; j ++) {
             var attr = eventNames[j];
             var attrValue = elem['on' + attr];
           }
          }
```

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -54,6 +54,15 @@ export const isBrowser: boolean =
 export function patchProperty(obj, prop) {
   const desc = Object.getOwnPropertyDescriptor(obj, prop) || {enumerable: true, configurable: true};
 
+  const originalDesc = Object.getOwnPropertyDescriptor(obj, 'original' + prop);
+  if (!originalDesc) {
+    Object.defineProperty(obj, 'original' + prop, {
+      enumerable: false,
+      configurable: true,
+      get: desc.get
+    });
+  }
+
   // A property descriptor cannot have getter/setter and be writable
   // deleting the writable and value properties avoids this error:
   //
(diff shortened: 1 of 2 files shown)


JiaLiPassion:
If zone.js don't want to support inline event handler attribute, this PR is not necessary, inline event handler is not a good solution anyway.

Issue #243, opened 2016-02-01
matchMedia events?


justindujardin:
I'm building an angular2 app and I'm trying to do dynamic things when `window.matchMedia` tells me that a media breakpoint has been hit.  

It seems like this is bypassing change detection, and I suspect it might be because zone.js doesn't wrap the matchMedia API.

Does this sound plausible? Would you be open to a PR to work through and resolve the issue?  Right now I have to manually tick the ng2 app when things change, and it feels pretty dirty.

JiaLiPassion:
supported now.

PR #543, opened 2016-12-14
feature: add MatchMedia Zone support

poxrud:
Add MatchMedia (js media queries) Zone support. Patches MediaQueryList.addListener and  MediaQueryList.removeListener. Closes issue #243.

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -8,6 +8,7 @@
 
 import {patchTimer} from '../common/timers';
 import {patchClass, patchMethod, patchPrototype, zoneSymbol} from '../common/utils';
+import {patchMatchMediaAddListener, patchMatchMediaRemoveListener} from './matchmedia';
 
 import {propertyPatch} from './define-property';
 import {eventTargetPatch} from './event-target';
@@ -127,4 +128,10 @@ function patchXHR(window: any) {
 /// GEO_LOCATION
 if (_global['navigator'] && _global['navigator'].geolocation) {
   patchPrototype(_global['navigator'].geolocation, ['getCurrentPosition', 'watchPosition']);
+}
+
+/// MATCH_MEDIA
+if (_global['MediaQueryList']) {
+  patchMatchMediaAddListener(_global['MediaQueryList'].prototype, ['addListener']);
+  patchMatchMediaRemoveListener(_global['MediaQueryList'].prototype, ['removeListener']);
 }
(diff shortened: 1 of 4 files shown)


JiaLiPassion:
In your PR, you patch event listener with bindArgument, it will make the event handler in zone, but will not trigger zoneSpec's scheduleTask/invokeTask/cancelTask callback.

there is already a event listener patch logic in [utils.ts](https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L328)
I suggest you just use it.
so maybe your patch will just look like this. It is much more simple and can be treaded as eventTask.

```javascript
    patchMethod(MatchMedia.prototype, ADD_EVENT_LISTENER, () => zoneAwareAddEventListener);
    patchMethod(MatchMedia.prototype, REMOVE_EVENT_LISTENER, () => zoneAwareRemoveEventListener);
```

poxrud:
@JiaLiPassion Even though MediaQueryList's _addListener_() and _removeListener_() have similar looking names to EventTarget' _addEventListener_() and _removeEventListener_(), they have completely different method signatures and behavior. Therefore you cannot use the existing event listener patch logic in utils.ts.

JiaLiPassion:
Yes, you are right,  but I still believe we should use the common logic, I will make a PR to generalize the zoneAwareAddEventListener part, I just run a little test about MediaQuery with the generalized version, it worked, so with common logic, we can add other listener patch later.

JiaLiPassion:
I just make a PR for a more generic way to patch event target related methods.
so you can patch MediaQueryList with following logic.

```javascript
patchEventTarget(MediaQuerlList.prototype, 'addListener', 'removeListener', (self, args) => {
    return {
     useCapturing: false,
     eventName: 'mediaQuery',
     handler: args[0],
     target: self || _global,
     name: 'mediaQuery',
     invokeAddFunc: function (addFnSymbol: any, delegate: Task | NestedEventListenerOrEventListenerObject) {
         if (delegate && (<Task>delegate).invoke) {
           return this.target[addFnSymbol]((<Task>delegate).invoke);
         } else {
           return this.target[addFnSymbol](delegate);
         }
     },
     invokeRemoveFunc: function (removeFnSymbol:any, delegate: Task | NestedEventListenerOrEventListenerObject) {
       if (delegate && (<Task>delegate).invoke) {
         return this.target[removeFnSymbol]((<Task>delegate).invoke);
       } else {
         return this.target[removeFnSymbol](delegate);
       }
     }
   };
})
```

Is that make sense?

PR #547, opened 2016-12-16
fix #546, should keep error prototype chain correctly

### LINKED ISSUE #546: v0.7+ - custom Error classes do not work as expected
**Zone.js version:** 0.7.2

**Setup:**
```JavaScript
class MyError extends Error {}
let error = new MyError();
console.log(error instanceof MyError);
```

**Expected:** logs `true`
**Actual:** logs `false` - `error` is an instance of `Error`, rather than `MyError` when zone.js is loaded in the app.

**Demo:** https://plnkr.co/edit/kSKVeuhIrMcluK1D9JSl?p=preview 
Look in the console to see the erroneous behaviour, and then comment out the script tag including zone.js; and the correct behaviour will be observed in the console.

Replacing the version in the <script> tag with 0.6.26 in the above Plunker demo demonstrates correct behaviour - so this bug was introduced with v0.7+ it seems.

JiaLiPassion:
Fix #546, should keep error prototype chain correctly, in the original code, the ZoneAwareError will return a nativeError and lost the prototype chain

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -1283,13 +1283,14 @@ const Zone: ZoneType = (function(global: any) {
   function ZoneAwareError() {
     // Create an Error.
     let error: Error = NativeError.apply(this, arguments);
+    this.message = error.message;
 
     // Save original stack trace
-    error.originalStack = error.stack;
+    this.originalStack = error.stack;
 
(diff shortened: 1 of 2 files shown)


chuckjaz:
This change was reverted.

PR #548, opened 2016-12-16
fix(closure): avoid property renaming on globals

alexeagle:
This upstreams some of the local changes from google3.


alexeagle:
We need this for the Ionic demo, and since the changes were already reviewed in google3 I went ahead and merged this.

mhevery:
We don't usually update the `dist` folder in the same PR as the change. We update `dist` on PRs which push to npm only. For next time...

alexeagle:
ack. the dist update was in a different commit in the PR. needed so I can
depend on the latest github snapshot.
We should have a readme address this - I did look for some doc about
whether to update the dist folder.

Issue #536, opened 2016-12-11
Notifications API is not covered by Zone.js


marcovdb:
The [Notifications API](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API) is not covered by Zone.js - meaning, if I initiate a notification from a zone and attach an function to the notification's onclick event, this function is not executed in the current zone.

I originally opened this as an Angular issue (https://github.com/angular/angular/issues/12624) but I was asked to reopen it here, so here goes.

JiaLiPassion:
Sure, several other apis are not in zone now, I would like to make a pr to implement it.

PR #549, opened 2016-12-17
add generic patch eventtarget method

JiaLiPassion:
now util.ts provide a patchEventTargetMethods to patch dom addEventListener and nodejs EventEmitter addEventListener method. 
becasue dom/node addEventListener has the similar signature. 

```javascript
   addEventListener(eventName, handler, ...);
```
It is ok for dom element and node, but in other case, for example in PR #543, if we want to patch 
other eventTarget like MediaQuery , the add listener method signature is different 

```javascript
addListener(MediaQueryListListener listener);
```

So if we want to patch it, we need to write another patch logic like #543, it is not effective and in the future, if we want to patch other event Target like notification API like #536, we will need another same logic, so we need a generic way to patch event target like object.

In this PR, I provide  a ListenerTaskMeta creator to create related information for customize, so 
to fix #543, we may call like this.

```javascript
patchEventTargetMethods(MediaQuerlList.prototype, 'addListener', 'removeListener', (self, args) => {
    return {
     useCapturing: false,
     eventName: 'mediaQuery',
     handler: args[0],
     target: self || _global,
     name: 'mediaQuery',
     invokeAddFunc: function (addFnSymbol: any, delegate: Task | NestedEventListenerOrEventListenerObject) {
         if (delegate && (<Task>delegate).invoke) {
           return this.target[addFnSymbol]((<Task>delegate).invoke);
         } else {
           return this.target[addFnSymbol](delegate);
         }
     },
     invokeRemoveFunc: function (removeFnSymbol:any, delegate: Task | NestedEventListenerOrEventListenerObject) {
       if (delegate && (<Task>delegate).invoke) {
         return this.target[removeFnSymbol]((<Task>delegate).invoke);
       } else {
         return this.target[removeFnSymbol](delegate);
       }
     }
   };
})
```

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -118,14 +118,14 @@ const EVENT_TASKS = zoneSymbol('eventTasks');
 const ADD_EVENT_LISTENER = 'addEventListener';
 const REMOVE_EVENT_LISTENER = 'removeEventListener';
 
-interface NestedEventListener {
+export interface NestedEventListener {
   listener?: EventListenerOrEventListenerObject;
 }
 
-declare type NestedEventListenerOrEventListenerObject =
+export declare type NestedEventListenerOrEventListenerObject =
     NestedEventListener | EventListener | EventListenerObject;
 
-interface ListenerTaskMeta extends TaskData {
+export interface ListenerTaskMeta extends TaskData {
   useCapturing: boolean;
   eventName: string;
   handler: NestedEventListenerOrEventListenerObject;
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -344,18 +344,9 @@ const zoneAwareAddEventListener =
     makeZoneAwareAddListener(ADD_EVENT_LISTENER, REMOVE_EVENT_LISTENER);
 const zoneAwareRemoveEventListener = makeZoneAwareRemoveListener(REMOVE_EVENT_LISTENER);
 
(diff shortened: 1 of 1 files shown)


poxrud:
@JiaLiPassion please add the PR for MatchMedia support using your new patchEventTargetMethods. You can use the tests in my branch to validate.

JiaLiPassion:
@poxrud , ok, thank you, I will make a PR with the generic method and your test code.

JiaLiPassion:
@poxrud , I am testing the code in my branch now, the problem is that a lot of different errors occur in the test codes on IE10/IE11/Safari9/Android, I add some code to check the MediaQuery availability, but still error occurs, it will take some time for me to fix them, and I think the MediaQuery is so experimental feature, maybe at last you should call a API (zone.patchMediaQuery) from app side. I will continue to debug the test code in those env (IE/Android)

poxrud:
Yes you need to check if MatchMedia is supported with something like:

```
if (_global['MediaQueryList']) {
 patchEventTargetMethods...
}
```

If you don't mind pasting your patching code I'd like to test it out myself as well.

JiaLiPassion:
@poxrud , yeah, I have added the detection code, and the patch part is ok, but the test code will throw a lot of error, you can find my commits here, [commits](https://github.com/JiaLiPassion/zone.js/commits/issue-543), and you can check Travis log to see the error details. the code is on my branch issue-543.

Issue #555, opened 2016-12-20
Zone.js Does not support creating Error objects without new oberator


KrauseStefan:
When running protractor with angular2 Dependency injection and therefore using zone.js I get the following error:
```
 Error: TypeError: Cannot set property 'message' of undefined
    at ZoneAwareError (project\node_modules\zone.js\dist\zone-node.js:683:22)
    at captureStackTrace (project\node_modules\selenium-webdriver\lib\promise.js:691:11)
    at new Task (project\node_modules\selenium-webdriver\lib\promise.js:2662:29)
    at ControlFlow.execute (project\node_modules\selenium-webdriver\lib\promise.js:2246:16)
    at Function.createSession (project\node_modules\selenium-webdriver\lib\webdriver.js:329:24)
    at Builder.build (project\node_modules\selenium-webdriver\builder.js:458:24)
    at Local.DriverProvider.getNewDriver (project\node_modules\protractor\built\driverProviders\driverProvider.js:37:33)
    at Runner.createBrowser (project\node_modules\protractor\built\runner.js:197:43)
    at project\node_modules\protractor\built\runner.js:276:30
    at _fulfilled (project\node_modules\q\q.js:834:54)
    at self.promiseDispatch.done (project\node_modules\q\q.js:863:30)
    at Promise.promise.promiseDispatch (project\node_modules\q\q.js:796:13)
    at project\node_modules\q\q.js:604:44
    at runSingle (project\node_modules\q\q.js:137:13)
    at flush (project\node_modules\q\q.js:125:13)
```

The line producing errors in zone.js: https://github.com/angular/zone.js/blob/610649b77f2a4403c941fcccad39a0d8cba89593/lib/zone.ts#L1320

Example of a place where Error is called without new: https://github.com/SeleniumHQ/selenium/blob/ba56ad1ae0b98a1fe1efdd2163c62fc847950178/javascript/node/selenium-webdriver/lib/promise.js#L714

We can agree that it would be best practice to use new, but I think zone.js if it forces wrapping the Error class should work the same way.

JiaLiPassion:
I think it is my mistake, I didn't return this in PR #547, I'll make a PR now.

KrauseStefan:
I was trying to fix this as well, it seems that the issue does not reproduce normally in the test environment.
Apparently `"use strict"` is not applied there.
This defect does not reproduce unless it is there.

[...]

Issue #554, opened 2016-12-20
Unhandled Promise rejection: TypeError: Cannot set property 'name' of undefined


filipesilva:
Using `zone.js@0.7.3`, Angular 2 `routerLink` does not work properly whereas with  `zone.js@0.7.2` it did.

Here is a repro plunker: https://plnkr.co/edit/SYPR2SRi0xayQ693jJAB?p=preview
It is a simple list with a couple of links. Clicking on any link breaks the app with the following console error:
```
zone.js@0.7.3?main=browser:405 Unhandled Promise rejection: TypeError: Cannot set property 'name' of undefined
    at new EmptyError (VM20135 EmptyError.js:21)
    at FirstSubscriber._complete (VM20134 first.js:147)
// ...
```
In `index.html` if you revert `https://unpkg.com/zone.js@0.7.3` to `0.7.2` the app works again.

JiaLiPassion:
@filipesilva , sorry, it is my mistake, I didn't return this in PR #547, I'll make a PR now.

filipesilva:
@JiaLiPassion if I understand your PR correctly, it should only affect errors themselves. This scenario should not result in an error at all though, since it worked with `0.7.2` without throwing errors.

chuckjaz:
This has been fixed with 0.7.4

blueskyfish:
Hi Guys,

Thanks for the solve of the zone.js in version 0.7.4. 👍 :-)

PR #559, opened 2016-12-20
Fix #546, #554, #555, the error should keep prototype chain and can be called without new

### LINKED ISSUE #546: v0.7+ - custom Error classes do not work as expected

JiaLiPassion:
With the commit #547, the #546, error prototype issue was fixed, but a bigger issue occurs, that
Error object can be created with direct call.

let error = Error('message');
And the #547 PR will break it, cause a lot of errors, such as #554, #555

Sorry for the buggy commit in #547.

In this new PR, I checked the this exists or not and checked the this is global or not (non strict mode). So it should fix both #546 and #554, #555.

Please review, Thank you.


mhevery:
This was reverted on re-opened because one can call `Error` without the `new` operator. This PR breaks that.

lib/zone.ts
```diff
@@ -1352,13 +1363,18 @@ const Zone: ZoneType = (function(global: any) {
...
+      }
     }
-    return error;
-  }
+    return (this && this !== global) ? this : error;
+  };
 
   // Copy the prototype so that instanceof operator works as expected
-  ZoneAwareError.prototype = NativeError.prototype;
+  ZoneAwareError.prototype = Object.create(NativeError.prototype);
```
mhevery: Why is `Object.create` necessary here? What is wrong with having it the way it was before?

lib/zone.ts
```diff
@@ -1352,13 +1363,18 @@ const Zone: ZoneType = (function(global: any) {
...
+      if (this && this !== global) {
+        this.stack = this.zoneAwareStack = frames.join('\n');
+      } else {
+        error.stack = error.zoneAwareStack = frames.join('\n');
+        error.originalStack = originalStack;
+      }
     }
-    return error;
-  }
+    return (this && this !== global) ? this : error;
```
mhevery: The fact that sometimes we return `this` and sometimes `errror` concerns me. We have two seperate code path and I fear that things will only work in one or the other code path.

lib/zone.ts
```diff
@@ -1318,13 +1318,24 @@ const Zone: ZoneType = (function(global: any) {
   function ZoneAwareError() {
     // Create an Error.
     let error: Error = NativeError.apply(this, arguments);
-
+    if (this && this !== global) {
```
mhevery: I think a better strategy would be to always assume that we were called without the `new` operator, that way we don't have to decide.

JiaLiPassion:
@mhevery , Thank you for your review, I'll consider it again and reply to you soon.

JiaLiPassion:
And if we use typescript to write code, current logic is still ok, because typescript will compile 'extends' into 

```javascript
var __extends = (this && this.__extends) || function (d, b) {
    for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
    function __() { this.constructor = d; }
    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
};

var MyError = (function (_super) {
    __extends(MyError, _super);
    function MyError() {
        _super.apply(this, arguments);
    }
    return MyError;
}(Error));
```

But in browser, native implementation, I think the 'extends' may like
```javascript
var MyError = (function (_super) {
    __extends(MyError, _super);
    function MyError() {
        this = new _super(...arguments);
    }
    return MyError;
}(Error));
```
So, in browser, when user call var myError = new MyError(); myError(this) will be the nativeError which was returned from ZoneAwareError, but in typescript version, this is MyError.

I'll continue to do more research.

lib/zone.ts
JiaLiPassion: Yes, I understand your concern, you are right. I will change it.

lib/zone.ts
JiaLiPassion: Yes, it is a little risky code, I will modify it.

lib/zone.ts
JiaLiPassion: I just want to create a new object to let ZoneAwareError.prototype be different with NativeError's, so in the future, if someday modify ZoneAwareError's prototype will not impact NativeError's, for this issue #546, we don't need to use Object.create

JiaLiPassion:
@mhevery , I have changed my PR, please review it is correct or not, Thanks a lot.

package.json
```diff
@@ -69,7 +69,7 @@
     "ts-loader": "^0.6.0",
     "tslint": "^3.15.1",
     "tslint-eslint-rules": "^3.1.0",
-    "typescript": "^2.0.2",
```
mhevery: Let's not upgrade to TS 2.1.4 since we have to make sure that the generated `.d.ts` is compatible with angular which is on alder version on TS. The easiest is to wait until angular is ready.

lib/zone.ts
```diff
@@ -1354,6 +1362,12 @@ const Zone: ZoneType = (function(global: any) {
       }
       error.stack = error.zoneAwareStack = frames.join('\n');
     }
+    // if this is valid, means not undefined
+    // and not global (non-strict mode call Error())
+    // and it is an error not an unrelated object
+    if (this && this !== global && this instanceof Error) {
+      error = setPrototypeOf(error, getPrototypeOf(this));
```
mhevery: Calling `setProtypeOf` will make this object slow. Not sure if that is a problem since it is a Error object.

package.json
JiaLiPassion: Ok, I see, I will rollback the typescript version.

lib/zone.ts
JiaLiPassion: Thank you for pointing out this one, I didn't realize that setPrototypeOf will make javascript engine optimization become slow.  how about change the code above to 

```javascript
    // if this is valid, means not undefined
    // and not global (non-strict mode call Error())
    // and it is an error not an unrelated object
    if (this && this !== global && this instanceof Error) {
       Object.assign(error, this);
       return this;
    }
```

-Pro: don't use setPrototypeOf
-Con:if/else will return different object

lib/zone.ts
```diff
@@ -1387,8 +1387,8 @@ const Zone: ZoneType = (function(global: any) {
     // if this is valid, means not undefined
     // and not global (non-strict mode call Error())
     // and it is an error not an unrelated object
-    if (this && this !== global && this instanceof Error
-      && Object.getPrototypeOf(this) !== NativeError.prototype && assignAll(this, error)) {
+    if (this && this !== global && this instanceof Error &&
```
mhevery: I was thinking about this some more, and I think the best way to do this would be to place this guard at the top of the function

```
if (! this instanceof ZoneAwareError) {
  return ZoneAwareError.apply(Object.create(ZoneAwareError), arguments);
}
```

This way we force the `new` and only have one code path. What do you think?

JiaLiPassion: Thank you for the review, in my understanding, your code will guarantee that 'this' is a valid ZoneAwareError, and we can set all properties from error to this and then return this. Is my understanding correct? 

I have modified the PR based on your idea, please review. Thank you.

mhevery:
Thanks for all of your hard work.

Issue #287, opened 2016-03-20
"More tasks executed then were scheduled" exception thrown


mdekrey:
First off, I'll admit to not knowing whether this is an Angular2 or Zone.js issue. Sorry if I'm putting it in the wrong Issues list!

When loading the [Google Maps API asynchronously](http://stackoverflow.com/questions/34931771/how-to-load-google-maps-api-asynchronously-in-angular2),  setTimeout and clearTimeout appear to be called which causes the error `More tasks executed then were scheduled.`

When I load it before the zone.js library, I am able to get further in the process. However, callbacks threw similar errors (eventually) without [further instrumentation](http://stackoverflow.com/questions/31352397/how-to-update-view-after-change-in-angular2-after-google-event-listener-fired). (I'm not sure if this is expected, but I suspect it is. I figured I'd mention it just in case.)

I have not been able to resolve, however, that the error occurs again if I leave the window and return upon moving my mouse over the map again. This event appears to be wholly internal to Google Maps, and I can't capture it.  Unfortunately, due to the `debugger;` line in the code, I'm unable to just ignore it for the time being without altering the codebase.

I'm using Angular 2.0.0-beta.11 and zone.js 0.6.4.

robwormald:
Could you post a reproduction in a plunker? This wouldn't wholly surprise me but I can't reproduce it.

fxck:
happens when using this lib https://github.com/stabzs/Angular2-Toaster as well, after it timeouts and is about to disappear

born2net:
+1 seeing same issues

born2net:
I will add that this seems to have started in 0.10...

born2net:
FYI disabled snippet is a workaround until issue is fixed in zone.ts
```
line: 763 zone.ts
if (next < 0) {
        debugger;
        throw new Error('More tasks executed then were scheduled.');
      }
```

rvalimaki:
Happens with my app as well, all the time. I'm not using anything related to Google Maps, so I guess it's completely irrelevant to the subject. Probably the title should be altered to better match the subject.

@born2net: What do you mean by "disabled snippet"? I there any workaround over this issue without touching Zone.js?

I'm not sure if it's related, but I'm also receiving exceptions about Task cancellation:
"EXCEPTION: Error: Task does not support cancellation, or is already canceled.

[...]

PR #570, opened 2016-12-29
fix #569, request will cause updateTaskCount failed if we call abort …

### LINKED ISSUE #569: XMLHTTPRequest abort multipletimes will cause updateTaskCount incorrect
Based on #287, and the comment from @RicardoVaranda's  [comment](https://github.com/angular/zone.js/issues/287#issuecomment-269608380), the XMLHTTPRequest will cause Uncaught Error: More tasks executed then were scheduled. The error can be described 
as 

```javascript
      req.open('get', '/', true);
      req.send();
      req.addEventListener('readystatechange', function(ev) {
        if (req.readyState >= 2) {
          req.abort();
        }
      });
```

The abort will be called multiple times before the request is done.
So we should add logic to verify whether the request has already been aborted.

JiaLiPassion:
Fix the issue #287, #569.

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -125,7 +125,10 @@ function patchXHR(window: any) {
         const task: Task = findPendingTask(self);
         if (task && typeof task.type == 'string') {
           // If the XHR has already completed, do nothing.
-          if (task.cancelFn == null) {
+          // If the XHR has already been aborted, do nothing.
+          // Fix #569, call abort multiple times before done will cause
+          // macroTask task count be negative number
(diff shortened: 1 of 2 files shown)


vicb:
Could you describe the error case in the commit message / inline comment and squash the commits ?
Thanks

JiaLiPassion:
@vicb, thank you for the review, I have added inline comment and squash the commits, do I need to add the issue link in comment ? please review.

RicardoVaranda:
@vicb can this be merged? Really need this fix so I can get rid of those annoying console messages.

PR #571, opened 2016-12-31
add web-api.ts to patch mediaQuery

JiaLiPassion:
Based on the PR #543 and #549, I add a new file under lib/browser web-api.ts to patch mediaQuery, and other non-standard api later (for instance, notification API).

the patch in web-api will not be loaded by default, because those api are not supported by all browsers and versions, and it is difficult to tell whether the browser support it or not.  So user should load web-api.js by themselves. 

And because it is too difficult to test mediaQuery in jasmine,I only add the basic test case to see the prototype is patched or not.


JiaLiPassion:
@code-tree , yeah, I also considered how to separate the API into different files or modules, I'll do some research about how other API does, if you have some good ideas, please tell me, thank you.

PR #572, opened 2016-12-31
patch fs methods as macrotask, add test cases of fs watcher

JiaLiPassion:
1. patch fs async methods (open, exists, write...) as macrotask, in previous patch, we use wrap, 
in my understanding, patch with wrap vs as macro task, we can both intercept schedule and invoke, 
but patch with macroTask can also have the ability to intercept cancelTask and hasTask.

2. add test cases of fs watcher related methods. Because we already patched EventEmitter, so fs.watch, fs.watchFile, fs.unwatchFile don't need additional patch. 
3. add a generic patch macro task method in util.ts

And in original patch, we patched the fs function with wrap, now we treat them as macrotask, so fs method will make additional callback (invokeTask/scheduleTask etc..) call, it may have some performance impact, I will make some other performance demo to check it.

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -481,3 +481,37 @@ export function patchMethod(
   }
   return delegate;
 }
+
+export interface MacroTaskMeta extends TaskData {
+  name: string;
+  target: any;
+  callbackIndex: number;
+  args: any[];
+}
+
+// TODO: support cancel task later if necessary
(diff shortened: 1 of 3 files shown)


JiaLiPassion:
I have make some performance test to compare the performance, basically the macrotask-patch version run the same performance with non-patch and wrap-patch version.

PR #580, opened 2017-01-02
fix #579, add isOrphan property to zoneSpec

### LINKED ISSUE #579: Implement a orphan zonespec not call parentZoneDelegate's callback
Sometimes we need a zone that run as an orphan without calling parentZoneDelegate, for example, 
in the following example.

```javascript
  const zoneASpec = {
        name: 'A',
        onScheduleTask: (delegate, currentZone, targetZone, task): Task => {
          console.log('log here....');
        }
      };
```

In the zoneASpec, we want to console.log something or do some other stuffs which may
be an async operation. And the new created async operation will call the onScheduleTask again 
and eventually cause dead loop.  (the above sample will cause dead loop in node.js, because 
console.log will call process.nextTick and will trigger onScheduleTask again).

Maybe it is not recommended to do heavy work in zoneSpec's callback, but I think we still need to
have the ability to do that.

So we really need is 

```javascript
  const zoneASpec = {
        name: 'A',
        onScheduleTask: (delegate, currentZone, targetZone, task): Task => {
           Zone.current.fork({name:'orphan', isOrphan: true}).run(() => {
              console.log('log here');
           });
        }
      };
```

the orphanZoneSpec will not call parentZoneDelegate's callback. 

**So I think basically we should use the orphanZoneSpec to do staff that may trigger 
another async operation all the time.**

**Base on the above idea, another approach is by default let all zoneSpec's callback in the orphan zone, so developer don't need to create orphan zone by themselves **

Please help me to review whether the thought is correct or not.

JiaLiPassion:
fix #579, add isOrphan property to zoneSpec, by default is false, so will not impact current behavior, 
if isOrphan is true, then the parent zone's callback will not be called.

- usage

```javascript
  const zoneASpec = {
        name: 'A',
        onScheduleTask: (delegate, currentZone, targetZone, task): Task => {
           Zone.current.fork({name:'orphan', isOrphan: true}).run(() => {
              console.log('log here');
           });
        }
      };
```

So the code run in orphan zone will not trigger it's parent zone zoneA's callback.

I think maybe all code run in zoneSpec's callback which may trigger other async operation 
should run in an orphan zone, otherwise may trigger dead loop error.
for example.

```javascript
const zoneASpec = {
        name: 'A',
        onScheduleTask: (delegate, currentZone, targetZone, task): Task => {
           asyncLog('log here');
        }
      };
```

such code may cause dead loop, because asyncLog will also trigger onScheduleTask 
in zoneA.

with orphanZone, the code may like this.
```javascript
  const zoneASpec = {
        name: 'A',
        onScheduleTask: (delegate, currentZone, targetZone, task): Task => {
           Zone.current.fork({name:'orphan', isOrphan: true}).run(() => {
              asyncLog('log here');
           });
        }
      };
```

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -271,6 +271,12 @@ interface ZoneSpec {
   name: string;
 
   /**
+   * Is the spec is Orphan Spec, by default orphan is false
+   * If Orphan is true, than the spec will not call parent zone delegate's callback
+   */
+  isOrphan?: boolean;
+
+  /**
    * A set of properties to be associated with Zone. Use [Zone.get] to retrive them.
    */
   properties?: {[key: string]: any};
@@ -717,6 +723,8 @@ const Zone: ZoneType = (function(global: any) {
   class ZoneDelegate implements AmbientZoneDelegate {
     public zone: Zone;
 
+    private _isOrphan: boolean = false;
+
     private _taskCounts: {microTask: number,
                           macroTask: number,
                           eventTask: number} = {microTask: 0, macroTask: 0, eventTask: 0};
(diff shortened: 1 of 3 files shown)


mhevery:
Why not just run `asyncLog` in root zone to prevent loop? Not sure about `isOrphan` can you discuss more the motivations?

JiaLiPassion:
@mhevery , Thank you for your review, yes, run asyncLog in root zone can resolve the problem, but sometimes we need to run some async code in current zone instead of root zone, for example, we can't change root zone spec's code because root zone spec is in other library (angular2 etc.), another example is we just want to debug something by using console.log (console.log will cause async operation by calling process.nextTick).

mhevery:
> for example, we can't change root zone spec's code because root zone spec

But you can fork from the root zone, which is effectively very similar to `isOprphan'.

> console.log will cause async operation by calling process.nextTick

`console.log` is sync and it should not cause `process.nextTick`. Did you mean your own `asyncLog` which writes to the server?


What is the benefit of having `isOrphan` since we can just fork a new spec from `RootZone`? What kids of `ZoneSpec` hooks would you like to run, which can't be just forked from rootZone?

mhevery:
Also this is behavior is hard to explain, which tells me that it probably is not a good primitive.

JiaLiPassion:
@mhevery , yeah, I agree with you, fork from root zone is much better way to do that, I'll close this PR, should we provide a util method to get Root Zone ?

PR #601, opened 2017-01-15
add Zone.root to get root zone directly 

JiaLiPassion:
add runOutsideOfCurrentZone method just like runOutsideAngular so we can run under rootZone or fork a new zone under rootZone.

```javascript
Zone.current.fork({name: 'testZone'}).run(function() {
        Zone.runOutsideOfCurrentZone(() => {
          expect(Zone.current.name).toEqual('<root>');
        });
      });
```

The motivation is to run some async code inside the zoneSpec's callback. Without the runOutsideOfCurrentZone method, such async call will cause infinite loop. 
for example:

```javascript
Zone.current.fork({
  name: 'testZone', 
  onScheduleTask: function(delegate, currentZone, targetZone, task) {
     asyncLog(task);
  }
});
```

asyncLog will cause infinite loop, now we can 

```javascript
Zone.current.fork({
  name: 'testZone', 
  onScheduleTask: function(delegate, currentZone, targetZone, task) {
     Zone.runOutsideCurrentZone(() => {
        asyncLog(task);
    });
  }
});
```

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -257,6 +257,11 @@ interface ZoneType {
    * Verify that Zone has been correctly patched. Specifically that Promise is zone aware.
    */
   assertZonePatched();
+
+  /**
+   *  Return the root zone.
+   */
+  root: Zone;
 }
 
 /**
@@ -564,6 +569,13 @@ const Zone: ZoneType = (function(global: any) {
       }
     }
 
+    static get root(): AmbientZone {
+      let zone = Zone.current;
+      while (zone.parent) {
+        zone = zone.parent;
+      }
+      return zone;
+    }
 
     static get current(): AmbientZone {
       return _currentZoneFrame.zone;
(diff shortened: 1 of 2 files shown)


lib/zone.ts
```diff
@@ -564,6 +571,27 @@ const Zone: ZoneType = (function(global: any) {
       }
     }
 
+    static get rootZone(): AmbientZone {
```
mhevery: I am fine with this API, but I would rename it to `root`

lib/zone.ts
```diff
@@ -564,6 +571,27 @@ const Zone: ZoneType = (function(global: any) {
...
+      let zone = Zone.current;
+      while (zone.parent) {
+        zone = zone.parent;
+      }
+      return zone;
+    }
+
+    static runOutsideOfCurrentZone(
```
mhevery: I think this API gives too much complexity to the user. Same can be achieved by:
```
Zone.root.fork(zoneSpec).run(....)
```

So I don't see the need to have it.

JiaLiPassion:
Ok, got it,I will modify it to just provide a root getter API.

PR #608, opened 2017-01-22
fix #533,  make a zone-mix to handle like electron/nw to patch browser and nodejs

### LINKED ISSUE #533: zone-node does not trigger change detection in Angular-CLI projects since version 1.0.0-beta.20
I use Angular 2 in Electron and created [this](https://github.com/DevWurm/electron-explorer) little sample app, with version 1.0.0-beta.19-3 of the Angular-CLI. In this case the change detection of data, which are altered through Node core functions (fs.readdir) work absolutely fine, when using the zone-node polyfill.

When I tried to build the same app with version 1.0.0-beta.20-4 of the CLI the change detection did not work at all.
I also built the app for Electron and included the zone-node polyfill, but in contrast to the first time I wrote the app, the data got not updated in the UI, when they were altered (i.e. when changing the directory in the file explorer). 

Both versions use Angular 2.1.

There is a small difference in the behaviour when using the plain zone polyfill:<br>
When using zone-node, the data get displayed the first time after loading the view  (they get fetched the first time in the OnViewInit Lifecycle event). When using zone, they don't.

I have absolutely no idea, what could cause the problem, so I don't know, which information could be necessary. Please ask for additional information, if you have any idea!

JiaLiPassion:
fix #533,  in electron/nw.js, we may need to patch both browser API and nodejs API, so we need a zone-mix.js to contains both patched API.

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -50,6 +50,10 @@ export const isNode: boolean =
 export const isBrowser: boolean =
     !isNode && !isWebWorker && !!(typeof window !== 'undefined' && window['HTMLElement']);
 
+// we are in electron of nw, so we are both browser and nodejs
+export const isMix: boolean = typeof process !== 'undefined' &&
+    {}.toString.call(process) === '[object process]' && !isWebWorker &&
+    !!(typeof window !== 'undefined' && window['HTMLElement']);
 
 export function patchProperty(obj, prop) {
   const desc = Object.getOwnPropertyDescriptor(obj, prop) || {enumerable: true, configurable: true};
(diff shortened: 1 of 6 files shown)


lib/common/utils.ts
```diff
@@ -50,6 +50,10 @@ export const isNode: boolean =
 export const isBrowser: boolean =
     !isNode && !isWebWorker && !!(typeof window !== 'undefined' && window['HTMLElement']);
 
+// we are in electron of nw, so we are both browser and nodejs
+export const isMix: boolean = typeof process !== 'undefined' &&
```
mhevery: Why do we call it `mix`? Would `electron` be a better name?

JiaLiPassion: I just wonder if nw.js is the same with electron, if we only handle electron, zone-electron is a better name

JiaLiPassion:
@mhevery , I think nw.js should have the same behavior with electron, so I use zone-mix and isMix name here, please review.

Issue #476, opened 2016-10-03
HTTPS security is compromised by null in Edge


RaleS:
SEC7111: HTTPS security is compromised by null
The error is logged in Edge console window when running zone.js 0.6.25

https://jsfiddle.net/g0mzz515/

vicb:
What is your setup ? Does this happen only in jsfiddle ?

RaleS:
@vicb Not sure why do you ask about the setup. Here is another example with zone.js 0.6.25

https://nspeubk.blob.core.windows.net/temp/zonejsedge.html

vicb:
I should have been clearer, what's Edge version do you use ?
Thanks

RaleS:
It's EdgeHTML 14.14393, Edge 38.14393.0.0, but it also fails in preview Edge 39.14936.1000.0, EdgeHTML 14.14936. Can you reproduce it?

born2net:
same issue here...

patrickmichalina:
Happening here too in an Angular 2 app..

JiaLiPassion:
I did a little debug, I think it is related to fetch promise patch, maybe PR #622 will fix this one.

MichaelBuerge:
Author of PR #622 here. I have a little demo Angular app live that includes said PR (+lots of dev output): https://orino.ch/test/zone.js-fetch/
I'm not seeing the security error anymore there, so it seems that #622 indeed addresses this too.

antonyRoberts:
@JiaLiPassion Can we get this pull request merged?

JiaLiPassion:
@antonyRoberts , sorry I don't have the permission to merge pull request, I think the PR is under review.

MichaelBuerge:
#622

Issue #446, opened 2016-09-15
#445 fix causes uncaught promise rejection in chrome


petewalker:
The fix in #445 causes an uncaught promise rejection in Chrome.

**Chrome Version:** 53.0.2785.116 (64-bit)

**Reproduction:**
1. Go to http://plnkr.co/edit/UwHURsu8rchPkTXoveTZ?p=preview
2. Open up Chrome Dev Tools
3. Refresh the page

**Expected:**

Page refreshes without error

**Actual:**

Page refresh is caught by uncaught rejection:

![screen shot 2016-09-15 at 15 36 52](https://cloud.githubusercontent.com/assets/453816/18554136/1feebcee-7b5b-11e6-8c8b-a09ece62d34f.png)

petewalker:
Actually, is this a bug with chrome devtools? The exception does appear to be being caught - but I don't have "Pause On Caught Exceptions" ticked, as per the screenshot.

jimitndiaye:
Having this same problem on Chrome

M4ttsson:
Having the same problem, Chrome 53.0.2785.116 m (64-bit). 

My angular 2 page works fine in Firefox but gets this error in Chrome: 
"Failed to execute 'fetch' on 'Window': 1 argument required, but only 0 present."

wizarrc:
It looks like chrome may have fixed their "bug"?? where they allowed an empty fetch call, because that did work for me before without hitting a promise rejection breakpoint while in debug mode before. I'm not sure which version of chrome I was on, but I am currently on Version 53.0.2785.116 m (64-bit). My previous fix using an empty string still seems to work in both Edge and IE.

`global['fetch']('')`

https://github.com/angular/zone.js/issues/436#issuecomment-246804254

hartsan:
We're having this problem, too, and I can confirm the wizarrc's fix solves the issue.

wesleycho:
We're seeing this problem in only Edge 38.14393.0.0, where it crashes the script load.

petewalker:
The problem with @wizarrc's solution - https://github.com/angular/zone.js/issues/446#issuecomment-249450289 - is that it introduces a side-effect - a GET request is made to the current page:

![sep-29-2016 14-38-07](https://cloud.githubusercontent.com/assets/453816/18956187/64de9386-8652-11e6-93ce-4949e895968b.gif)

wizarrc:
[...]

PR #622, opened 2017-02-01
Make fetch() zone-aware without triggering extra requests or uncatchable errors.

MichaelBuerge:
Previous implementation called fetch() during setup in order to obtain a result promise from it so its constructor could be patched to make it zone-aware.
This had two problems:
- Calling fetch() would trigger an unwanted (and unnecessary) HTTP-request.
- Chrome threw an uncatchable error, which means with open dev tools will halt execution and enter the debugger (which is hugely annoying, as it happens on every page load).

This change replaces fetch() with a wrapper function that lazily ensures the result returned by fetch() is zone-aware.


MichaelBuerge:
CLA signed.

MichaelBuerge:
#446

JiaLiPassion:
@MichaelBuerge , please read [DEVELOPER.md](https://github.com/angular/zone.js/blob/master/DEVELOPER.md), before commit part, so you can check your commit. Currently it seems you build failed with tslint part.

MichaelBuerge:
@JiaLiPassion Apologies for missing the linter. Everything should be fine now (`npm run ci` goes through).

JiaLiPassion:
@MichaelBuerge , you are welcome, could you check my review comment? especially whether the first promise is patched or not one.

MichaelBuerge:
@JiaLiPassion I can't find your review comment. As of now, the pull request page still says "no reviews". (Am I looking in the wrong place?)

However, guessing from your comment here, your question seems to be whether the first time the patched fetch() function is called its result promise is properly patched.
That is indeed the case, as the constructor's prototype is modified before any caller can get a hold of the returned promise. Also, the fetch tests (in `test/common/Promise.spec.ts`) would not pass with an unpatched result promise.
Does this address your concern?

JiaLiPassion:
@MichaelBuerge, sorry I forget to push submit review button...

> That is indeed the case, as the constructor's prototype is modified before any caller can get a hold of the returned promise. Also, the fetch tests (in test/common/Promise.spec.ts) would not pass with an unpatched result promise.
> Does this address your concern?

yeah, you are right! thank you for explaining this, I didn't realize that.

lib/zone.ts
```diff
@@ -1313,38 +1313,47 @@ const Zone: ZoneType = (function(global: any) {
...
-      const nativePromise = this;
-      return new ZoneAwarePromise((resolve, reject) => {
-               NativePromiseThen.call(nativePromise, resolve, reject);
-             })
-          .then(onResolve, onReject);
+
+  const symbolThenPatched = __symbol__('thenPatched');
+
+  function patchThen(Ctor) {
+    let proto = Ctor.prototype;
```
MichaelBuerge: done

lib/zone.ts
```diff
@@ -1313,38 +1313,47 @@ const Zone: ZoneType = (function(global: any) {
...
+    let originalThen = proto.then;
+    // Keep a reference to the original method.
+    proto[symbolThen] = originalThen;
+
+    Ctor.prototype.then = function(onResolve, onReject) {
+      let wrapped = new ZoneAwarePromise(
```
MichaelBuerge: done

MichaelBuerge:
My pleasure!
I assume there's a process in place, but if there's anything left do so this gets merged eventually and the associated issue (https://github.com/angular/zone.js/issues/446) closed please let me know.

MichaelBuerge:
Also addresses #476

PR #629, opened 2017-02-10
feat: allow tasks to be canceled and rescheduled on different zone in…

mhevery:
… a zone delegate

### DIFF
--- a/lib/common/timers.ts
+++ b/lib/common/timers.ts
@@ -71,7 +71,8 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
       patchMethod(window, cancelName, (delegate: Function) => function(self: any, args: any[]) {
         const task: Task = typeof args[0] === 'number' ? tasksByHandleId[args[0]] : args[0];
         if (task && typeof task.type === 'string') {
-          if (task.cancelFn && task.data.isPeriodic || task.runCount === 0) {
+          if (task.state !== 'notScheduled' &&
+              (task.cancelFn && task.data.isPeriodic || task.runCount === 0)) {
             // Do not cancel already canceled functions
             task.zone.cancelTask(task);
           }
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -220,15 +220,54 @@ interface Zone {
    * @returns {*}
    */
   runTask(task: Task, applyThis?: any, applyArgs?: any): any;
+
+  /**
+   * Schedule a MicroTask.
+   *
+   * @param source
+   * @param callback
+   * @param data
+   * @param customSchedule
+   */
   scheduleMicroTask(
       source: string, callback: Function, data?: TaskData,
       customSchedule?: (task: Task) => void): MicroTask;
+
+  /**
+   * Schedule a MacroTask.
+   *
+   * @param source
+   * @param callback
+   * @param data
+   * @param customSchedule
+   * @param customCancel
+   */
   scheduleMacroTask(
       source: string, callback: Function, data: TaskData, customSchedule: (task: Task) => void,
       customCancel: (task: Task) => void): MacroTask;
+
+  /**
+   * Schedule an EventTask.
+   *
+   * @param source
+   * @param callback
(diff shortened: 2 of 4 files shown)


mhevery:
TODO
- [ ] Needs a better commit message
- [ ] Add test for task cancelation in ZoneDelegate, which shows how a Task can be rescheduled on a different zone.

JiaLiPassion:
@mhevery, could you explain a little more about the motivation of this PR, I would like to understand more deeply about zone.js, thank you very much.

mhevery:
@JiaLiPassion sorry for the delayed response. 

The issue can best be described by this [test](https://github.com/angular/zone.js/pull/629/files#diff-29a5b9319349c754058e070be5b0215aR235) 

Imagine that you want to write a ZoneSpec which would redirect the scheduling of a task to a different zone. For example you want to make sure that all `addEventListeners` on `scroll` would run outside of some Zone. What you need to do is to intercept the scheduling of the task and than reschedule it on a different zone. However, that could not have been done, since there was no way to tell the `ZoneDelegate`s that the current task, is not part of the current request. We now have `Task. cancelScheduleRequest()` which does just that.

In addition it was hard to keep track of in which state the task was, so I made the Task state explicit, which helps with the error messages.

karma-base.conf.js
```diff
@@ -39,7 +39,7 @@ module.exports = function (config) {
 
     logLevel: config.LOG_INFO,
 
-    browsers: ['Firefox'],
+    browsers: ['ChromeCanary'],
```
mhevery: revert

lib/common/timers.ts
```diff
@@ -71,8 +71,8 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
       patchMethod(window, cancelName, (delegate: Function) => function(self: any, args: any[]) {
         const task: Task = typeof args[0] === 'number' ? tasksByHandleId[args[0]] : args[0];
         if (task && typeof task.type === 'string') {
-          if (task.cancelFn && task.data.isPeriodic || task.runCount === 0) {
-            // Do not cancel already canceled functions
+          if (task.state !== 'notScheduled' && task.cancelFn ) {
```
mhevery: revert

lib/common/utils.ts
```diff
@@ -467,9 +467,8 @@ export function createNamedFn(name: string, delegate: (self: any, args: any[]) =
 }
 
 export function patchMethod(
-    target: any, name: string,
-    patchFn: (delegate: Function, delegateName: string, name: string) => (self: any, args: any[]) =>
-        any): Function {
+    target: any, name: string, patchFn: (delegate: Function, delegateName: string, name: string) =>
```
mhevery: revert

lib/zone.ts
```diff
@@ -229,6 +229,7 @@ interface Zone {
   scheduleEventTask(
       source: string, callback: Function, data: TaskData, customSchedule: (task: Task) => void,
       customCancel: (task: Task) => void): EventTask;
+  scheduleTask<T extends Task>(task: T): T;
```
mhevery: add docs

lib/zone.ts
```diff
@@ -516,6 +527,8 @@ interface Task {
    * Number of times the task has been executed, or -1 if canceled.
    */
   runCount: number;
+
+  cancelScheduleRequest(): void;
```
mhevery: add docs

lib/zone.ts
```diff
@@ -622,13 +639,13 @@ const Zone: ZoneType = (function(global: any) {
     }
 
     public fork(zoneSpec: ZoneSpec): AmbientZone {
-      if (!zoneSpec) throw new Error('ZoneSpec required!');
+      if (!zoneSpec) throw Error('ZoneSpec required!');
```
mhevery: revert

JiaLiPassion:
@mhevery , Thank you for explaining, I will read the source and learn how it works!

PR #631, opened 2017-02-11
feat(longStackTraceSpec): handled promise rejection can also render longstacktrace

JiaLiPassion:
1. fix typo in longstacktrace
2. zone.js have longStackTraceSpec, it is very powerful, but it can't render handled promise rejection.

```javascript
const longStackTraceZoneSpec = Zone['longStackTraceZoneSpec'];

Zone.current.fork(longStackTraceZoneSpec).run(() => {
  Promise p = new Promise((resolve, reject) => {
    setTimeout(() => {
      reject(new Error('Promise error'));
    }, 10);
  });

  p.catch(error => console.log(error.stack);}
});
```

In this case, the error p.catch got is not a longStackTrace, because it is a handled promise rejection so it will not call longStackTraceSpec's onHandleError callback.

in this PR, I add a helper method in LongStackTraceZoneSpec, so if user want to render longStackTrace in promise.then/catch, user can call the method.

```javascript
const longStackTraceZoneSpec = Zone['longStackTraceZoneSpec'];

Zone.current.fork(longStackTraceZoneSpec).run(() => {
  Promise p = new Promise((resolve, reject) => {
    setTimeout(() => {
      reject(new Error('Promise error'));
    }, 10);
  });

  p.catch(error => console.log(longStackTraceZoneSpec.getLongStackTrace(error));}
});
```

Here, user should call longStackTraceZoneSpec.getLongStackTrace(error) themselves, error.stack
is still the original stack. the reason that not set stack automatically here is because I don't want to break the logic of "handled promise rejection".

### DIFF
--- a/lib/zone-spec/long-stack-trace.ts
+++ b/lib/zone-spec/long-stack-trace.ts
@@ -27,18 +27,18 @@ function getStacktraceWithUncaughtError(): Error {
 function getStacktraceWithCaughtError(): Error {
   try {
     throw getStacktraceWithUncaughtError();
-  } catch (error) {
-    return error;
+  } catch (err) {
+    return err;
   }
 }
 
 // Some implementations of exception handling don't create a stack trace if the exception
 // isn't thrown, however it's faster not to actually throw the exception.
 const error = getStacktraceWithUncaughtError();
-const coughtError = getStacktraceWithCaughtError();
+const caughtError = getStacktraceWithCaughtError();
 const getStacktrace = error.stack ?
     getStacktraceWithUncaughtError :
-    (coughtError.stack ? getStacktraceWithCaughtError : getStacktraceWithUncaughtError);
+    (caughtError.stack ? getStacktraceWithCaughtError : getStacktraceWithUncaughtError);
 
 function getFrames(error: Error): string[] {
   return error.stack ? error.stack.split(NEWLINE) : [];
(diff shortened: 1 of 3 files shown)


lib/zone-spec/long-stack-trace.ts
```diff
@@ -6,6 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {zoneSymbol} from '../common/utils';
```
mhevery: Can't do imports like that, because it would pull in the whole utils file. This file is meant to be stand alone and so no external imports are allowed.

Just use `Zone.__symbol__`

JiaLiPassion: Yes, you are right, I will modify it.

By the way, I just found the same problem exists in my previous PR, for web-apis-mediaquery patch, and web-apis-notification patch, those patches are also be separated, and import function from utils. I will also modify those files  to make sure the standalone files not import from utils.

JiaLiPassion:
@mhevery , I have modified it , please review.

test/zone-spec/long-stack-trace-zone.spec.ts
```diff
@@ -91,6 +92,25 @@ describe('longStackTraceZone', function() {
...
+      setTimeout(function() {
+        setTimeout(function() {
+          let promise = new Promise(function(resolve, reject) {
+            setTimeout(function() {
+              reject(new Error('Hello Promise'));
+            }, 0);
+          });
+          promise.catch(function(error) {
+            expect(longStackTraceZoneSpec.getLongStackTrace(error).split('Elapsed: ').length)
+                .toBe(4);
```
mhevery: What exactly is this asserting? Could we make a test which is more explicit?

test/zone-spec/long-stack-trace-zone.spec.ts
```diff
@@ -91,6 +92,25 @@ describe('longStackTraceZone', function() {
       }, 0);
     });
   });
+
+  it('should produce long stack traces when has handled error in promise', function(done) {
```
mhevery: ...`when handling errors in promises`

JiaLiPassion: modified

test/zone-spec/long-stack-trace-zone.spec.ts
JiaLiPassion: Here the asserting means with the longStackTraceZoneSpec.getLongStackTrace method, error which was catch in handing promise rejection can have long stack trace, so here I just assert the stack trace contains `Elapsed:` string.

JiaLiPassion:
@mhevery , I have updated the comment, please review whether it is ok. thank you.

PR #637, opened 2017-02-14
fix(timers): cleanup task reference when exception

bsergei:
Reference to task is not removed when timer's handler was faulty with exception. This can lead to memory leak.


bsergei:
I signed it!

mhevery:
Could we have a test for this?

Issue #638, opened 2017-02-14
"Paused in debugger" in Chrome with version "0.7.7"


fanshuai:
<img width="1134" alt="2017-02-15 00 56 43" src="https://cloud.githubusercontent.com/assets/1005487/22939460/b5b37676-f319-11e6-8741-2e0f40dc7d68.png">

When I click a button with `routerLink` for change router url it raise, only with version `0.7.7`.

JiaLiPassion:
@fanshuai , thank you for reporting the issue, I will make a PR to remove it.

rsnider19:
I am seeing the same thing when I run code in `this.zone.runOutsideAngular(() => {...});`. It is happening in 0.7.7, but not in 0.7.6. Even if you remove the `debugger`, it is still going to throw that error. Could there be another issue going on here?

JiaLiPassion:
@rsnider19 , can you post a screenshot of the error or the pause on debugger?

rsnider19:
![image](https://cloud.githubusercontent.com/assets/13546759/23103738/274cab32-f68e-11e6-8d02-d5aef7c0b0c4.png)

JiaLiPassion:
@rsnider19 , thank you for the error screenshot, could you post a repository to reproduce so I can debug into it?

rsnider19:
Sure thing! Here is a plunk. Just hover over the sections of the pie chart a couple times and it will trigger.
https://plnkr.co/edit/gar5B50EwAfrIrdU60m5

JiaLiPassion:
@rsnider19, Thank you, I have made a PR to fix this one.

rsnider19:
Awesome! Thanks!

naveedahmed1:
Should this PR also fix the issues reported here https://github.com/angular/angular/issues/14683#issuecomment-282119653

JiaLiPassion:
@naveedahmed1 , yes, the same issue.

naveedahmed1:
@JiaLiPassion btw when should we expect it to be released?

JiaLiPassion:
@naveedahmed1 , I think it is under review. It will take some time, Thank you.

naveedahmed1:
Is it possible to use the version with your submitted pr while its under review?

JiaLiPassion:
@naveedahmed1 , sure ,you can use the attached file to test.
[zone.js.zip](https://github.com/angular/zone.js/files/802489/zone.js.zip)

PR #642, opened 2017-02-21
fix(task): eventTask/Periodical task should not be reset after being cancelled when running

JiaLiPassion:
the issue can be described as the following cases.

```javascript
     const zone = Zone.current.fork({name: 'testZone'});

     const task = zone.scheduleMacroTask('testPeriodTask', () => {
       zone.cancelTask(task);
     }, {isPeriodic: true}, () => {}, () => {});

    task.invoke();
    expect(task.state).toBe('notScheduled');

    
     const task = zone.scheduleEventTask('testEventTask', () => {
       zone.cancelTask(task);
     }, null, () => {}, () => {});

    task.invoke();
    expect(task.state).toBe('notScheduled');
```

When task is cancelled in task's callback(running state), the runTask logic here https://github.com/angular/zone.js/blob/master/lib/zone.ts#L757 will reset the task's state to scheduled, so it will cause issues such as #638, updateTaskCount cause null error, because the task is cancelled but still in scheduled state.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -754,7 +754,11 @@ const Zone: ZoneType = (function(global: any) {
         }
       } finally {
         if (task.type == eventTask || (task.data && task.data.isPeriodic)) {
-          reEntryGuard && (task as ZoneTask)._transitionTo(scheduled, running, notScheduled);
+          // if the task's state is notScheduled, then it has already been cancelled
+          // we should not reset the state to scheduled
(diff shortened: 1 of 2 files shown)


naveedahmed1:
Can this be released anytime soon? Its breaks Angular applications in most of the cases.

naveedahmed1:
Thank you so much @mhevery my Angular 2 app is stable again after this update :)

PR #647, opened 2017-02-24
fix(error): fix #618, ZoneAwareError should copy Error's static properties as well

### LINKED ISSUE #618: Sharepoint 2013 conflict
Hi, I get an exception "Out of stack space" when I try to use sharepoint [controls](http://prntscr.com/e03bem) 
So, controls aren't available after the exception.

Issue is fully described on [stackoverflow](http://stackoverflow.com/questions/40466012/using-angular2-on-sharepoint-2013-masterpage-leads-to-zone-js-error)

Could you provide workaround for this problem ?
zone.js: 0.7.4 
Thanks.

JiaLiPassion:
fix #618, ZoneAwareError copy the Error's prototype properties , but not the properties of Error itself (static one)

for example, if other library define some custom function to Native Error

```javascript
Error.func = function() {}
```

then after zone.js patch Error, the func of Error will be lost.

in this PR, I also copy those properties from NativeError.

### DIFF
--- a/karma-build.conf.js
+++ b/karma-build.conf.js
@@ -9,6 +9,7 @@
 module.exports = function (config) {
   require('./karma-base.conf.js')(config);
   config.files.push('build/test/wtf_mock.js');
+  config.files.push('build/test/custom_error.js');
   config.files.push('build/lib/zone.js');
   config.files.push('build/test/main.js');
 };
--- a/test/custom_error.ts
+++ b/test/custom_error.ts
@@ -1,0 +1,14 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+'use strict';
+(function(global) {
+  const NativeError = global['Error'];
+  NativeError.customProperty = 'customProperty';
+  NativeError.customFunction = function() {};
+})(typeof window === 'object' && window || typeof self === 'object' && self || global);
--- a/test/node_entry_point.ts
+++ b/test/node_entry_point.ts
@@ -8,6 +8,7 @@
 
 // Must be loaded before zone loads, so that zone can detect WTF.
 import './wtf_mock';
+import './custom_error';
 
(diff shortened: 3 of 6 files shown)


karma-build.conf.js
```diff
@@ -9,6 +9,7 @@
 module.exports = function (config) {
   require('./karma-base.conf.js')(config);
   config.files.push('build/test/wtf_mock.js');
+  config.files.push('build/test/custom_error.js');
```
mhevery: Why do we need this?

test/custom_error.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
mhevery: I don't see this file being used.

test/node_entry_point.ts
```diff
@@ -8,6 +8,7 @@
 
 // Must be loaded before zone loads, so that zone can detect WTF.
 import './wtf_mock';
+import './custom_error';
```
mhevery: Why do we need this?

mhevery:
This PR is failing travis, can you get it rebased and green?

karma-build.conf.js
JiaLiPassion: I add this file just to simulate a Error with customized method before patched by zone.js. so I can test the customized method is still valid after zone.js patch Error with ZoneAwareError.

test/custom_error.ts
JiaLiPassion: it is used here.
https://github.com/JiaLiPassion/zone.js/blob/f318c3b1fce222f0d4551555ca5834b37e4868d6/test/common/Error.spec.ts#L167

test/node_entry_point.ts
JiaLiPassion: it is used here.
https://github.com/JiaLiPassion/zone.js/blob/f318c3b1fce222f0d4551555ca5834b37e4868d6/test/common/Error.spec.ts#L167

which the method is defined here

https://github.com/JiaLiPassion/zone.js/blob/f318c3b1fce222f0d4551555ca5834b37e4868d6/test/custom_error.ts#L13

PR #652, opened 2017-02-25
fix(webapi): refactor webapi to not import util.ts directly

JiaLiPassion:
1. webapi mediaquery, notification should not import methods from util.ts because it will pull all util.ts content into webapi-media-query.js and webapis-notification.js under dist
2. export some zone util's method through Zone[symbol(methodName)]
3. add webapis dist/min.js

### DIFF
--- a/lib/browser/webapis-media-query.ts
+++ b/lib/browser/webapis-media-query.ts
@@ -5,8 +5,6 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {NestedEventListenerOrEventListenerObject, patchEventTargetMethods} from '../common/utils';
-
 ((_global: any) => {
   // patch MediaQuery
   patchMediaQuery(_global);
@@ -15,6 +13,7 @@ import {NestedEventListenerOrEventListenerObject, patchEventTargetMethods} from 
     if (!_global['MediaQueryList']) {
       return;
     }
+    const patchEventTargetMethods = Zone[Zone['__symbol__']('patchEventTargetMethods')];
     patchEventTargetMethods(
         _global['MediaQueryList'].prototype, 'addListener', 'removeListener', (self, args) => {
           return {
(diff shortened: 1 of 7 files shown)


lib/browser/webapis-media-query.ts
```diff
@@ -23,16 +22,14 @@ import {NestedEventListenerOrEventListenerObject, patchEventTargetMethods} from
             handler: args[0],
             target: self || _global,
             name: 'mediaQuery',
-            invokeAddFunc: function(
-                addFnSymbol: any, delegate: Task|NestedEventListenerOrEventListenerObject) {
+            invokeAddFunc: function(addFnSymbol: any, delegate) {
```
mhevery: Why are you removing the types? That does not seem right.

lib/common/utils.ts
```diff
@@ -569,3 +569,15 @@ export function findEventTask(target: any, evtName: string): Task[] {
   }
   return result;
 }
+
+Zone[zoneSymbol('patchEventTargetMethods')] = patchEventTargetMethods;
```
mhevery: I feel very uneasy with exposing these APIs even if they are behind a symbol. What is the reasoning?

lib/browser/webapis-media-query.ts
JiaLiPassion: The reason is I don't want to import the type from until.ts which will drag all contents from util.ts when rollup build.

lib/common/utils.ts
JiaLiPassion: Yes,I feel the same,the reason to expose those API in such way is

- for easier patch other non standard or 3rd party API
- don't need to import those methods from util.ts

lib/browser/webapis-media-query.ts
JiaLiPassion: Just like the review here,https://github.com/angular/zone.js/pull/631#pullrequestreview-21679511

lib/common/utils.ts
mhevery: I think we should expose the minimal set. I will remove the rest of the exports and just keep `patchEventTargetMethods `

JiaLiPassion: sure, mediaQuery use patchEventTargetMethods, and Notification use patchOnProperties. I will leave those two and remove others.

PR #655, opened 2017-03-02
feat(bluebird): patch bluebird promise and treat it as microtask

### LINKED ISSUE #455: Zone.js conflicts with Bluebird
I'm including Zone 0.6.21 in a project that also uses [Bluebird](http://bluebirdjs.com) as one of the first steps for using Angular 2. Unfortunately, there does not seem to be a clean way for having the two libraries work together.

Bluebird is more than a "Promise shim", since it replaces the `Promise` implementation entirely, with a more performant one, while also adding many utilities both to the `Promise` constructor and its prototype.
- If Bluebird is included _before_ Zone (which I understand is the preferred way), Zone will overwrite the entire `Promise` constructor, loosing all utilities and the internal implementation of Bluebird. This basically defeats the purpose of including Bluebird in the first place.
- If Bluebird is included _after_ Zone _and is made Zone-aware by replacing its scheduling function_ then everything works: Bluebird is active and zones work correctly. However the `Zone.assertZonePatched` that Angular 2 calls will fail, since the `Promise` constructor is no longer the one from Zone.

Curently I can work around the problem by including Bluebird after Zone and then manually patching each function in the Bluebird API.

``` javascript
Promise.prototype.then = (function(_orig) {
  return function(fulfilledHandler, rejectedHandler) {
    return _orig.call(this, zonify(fulfilledHandler), zonify(rejectedHandler));
  }
})(Promise.prototype.then);

// similar patches for other 25+ functions

// hack to keep Angular happy
Zone.assertZonePatched = function() {};
```

If you are interested, [this Gist](https://gist.github.com/lorenzodallavecchia/d21684d5b7645df7e1dcd45642569004) shows the full code.

As you can see it is pretty hacky and I would better go without it if it was possible.

Is there a particular reason for Zone introducing an entire `Promise` constructor instead of just wrapping `then`, `catch` and other methods?

JiaLiPassion:
fix #455.

patch bluebird promise in zone and treat it as microTask,
add testcases for all bluebird APIs.

-usage, node

```javascript
require('zone-node.js');
const Bluebird = require('bluebird');
require('zone-bluebird.js');
Zone[Zone['__symbol__']('bluebird')](Bluebird);
```

- usage, browser
```javascript
 <script src="zone.js"></script>
  <script src="bluebird.js"></script>
  <script src="zone-bluebird.js"></script>
  <script>
  Zone[Zone['__symbol__']('bluebird')](Promise);
  </script>
```

We can simplify those steps to automatically patch Bluebird Promise, if global.Promise is Bluebird, but Bluebird maybe used just as a library (such as by sequelize), so I think patch it explicitly will be much safer.

### DIFF
--- a/dist/zone-bluebird.min.js
+++ b/dist/zone-bluebird.min.js
@@ -1,0 +1,1 @@
+!function(e,o){"object"==typeof exports&&"undefined"!=typeof module?o():"function"==typeof define&&define.amd?define(o):o()}(this,function(){"use strict";!function(e){var o=Zone.__symbol__;Zone[o("bluebird")]=function(e){e.setScheduler(function(e){Zone.current.scheduleMicroTask("bluebird",e)})}}("object"==typeof window&&window||"object"==typeof self&&self||global)});
--- a/NON-STANDARD-APIS.md
+++ b/NON-STANDARD-APIS.md
@@ -13,6 +13,48 @@ But there are still a lot of non standard APIs are not patched by default, such 
 
 ## Currently supported non standard node APIs
 
+## Currently supported non standard common APIs
+
+* bluebird promise
+
+Browser Usage: 
+
+```
+  <script src="zone.js"></script>
+  <script src="bluebird.js"></script>
+  <script src="zone-bluebird.js"></script>
+  <script>
+  Zone[Zone['__symbol__']('bluebird')](Promise);
+  </script>
+```
+
+After those steps, window.Promise will become a ZoneAware Bluebird Promise.
+
+Node Usage:
+
+```
+require('zone-node.js');
+const Bluebird = require('bluebird');
+require('zone-bluebird.js');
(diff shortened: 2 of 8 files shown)


dist/zone-bluebird.min.js
```diff
@@ -0,0 +1 @@
+!function(e,o){"object"==typeof exports&&"undefined"!=typeof module?o():"function"==typeof define&&define.amd?define(o):o()}(this,function(){"use strict";!function(e){var o=Zone.__symbol__;Zone[o("bluebird")]=function(e){e.setScheduler(function(e){Zone.current.scheduleMicroTask("bluebird",e)})}}("object"==typeof window&&window||"object"==typeof self&&self||global)});
```
JiaLiPassion: OK, got it, I will remove it.

PR #659, opened 2017-03-03
fix(jasmine): support "pending" `it` clauses with no test body

petebacondarwin:


lib/jasmine/jasmine.ts
```diff
@@ -89,10 +89,10 @@
     // The `done` callback is only passed through if the function expects at least one argument.
     // Note we have to make a function with correct number of arguments, otherwise jasmine will
     // think that all functions are sync or async.
-    return (testBody.length == 0) ? function() {
-      return testProxyZone.run(testBody, this);
-    } : function(done) {
+    return (testBody && testBody.length) ? function(done) {
       return testProxyZone.run(testBody, this, [done]);
+    } : function() {
```
gkalpak: If we always return a function, won't it essentially make all specs non-pending (since there will be a callback specified)?
I think this should be:

```ts
return testBody && ((testBody.length == 0) ? function() { ... } : function { ... });
```

petebacondarwin: Good spot

petebacondarwin: Not sure how to test that easily...

petebacondarwin: Done

package.json
```diff
@@ -79,7 +79,7 @@
     "ts-loader": "^0.6.0",
     "tslint": "^4.1.1",
     "tslint-eslint-rules": "^3.1.0",
-    "typescript": "^2.0.2",
+    "typescript": "2.1.x",
```
mhevery: We have to upgrade TS in a separate PR.

smithad15:
Just ran into this issue today. Would love to use it. Any idea on the timeline for the next release cut?

mhevery:
released

smithad15:
Many Thanks!

PR #665, opened 2017-03-08
fix(package): sync dependencies version with angular

JiaLiPassion:
copy all package dependencies' version from angular/angular.


package.json
```diff
@@ -62,24 +62,24 @@
...
     "nodejs-websocket": "^1.2.0",
     "promises-aplus-tests": "^2.1.2",
     "pump": "^1.0.1",
-    "systemjs": "^0.19.37",
+    "systemjs": "0.18.10",
     "ts-loader": "^0.6.0",
     "tslint": "^4.1.1",
     "tslint-eslint-rules": "^3.1.0",
-    "typescript": "2.0.10",
+    "typescript": "^2.1.5",
```
mhevery: can't upgrade typescript just yet. We need to do the upgrade in a separate PR and have separate version of Zones. Can you revert this one line?

JiaLiPassion:
Got it, I will revert it

package.json
JiaLiPassion: BTW, in a separate PR  you mentioned means we need to 

1. update typescript to 2.1.x
2. change some source code which was not compatible with 2.1.x

is that right?

PR #668, opened 2017-03-10
fix(webapi): expose patchOnProperties

JiaLiPassion:
expose patchOnProperties for Notification patch.


mhevery:
This is already merged.

PR #678, opened 2017-03-14
Patching node.js domain.bind to preserve Zone

Laguana:
A number of libraries (such as redis and mysql) include support for domains by invoking `domain.bind(callback)` at some point. With these changes, those libraries will now also preserve the current zone in their callbacks as well.

### DIFF
--- a/lib/node/node.ts
+++ b/lib/node/node.ts
@@ -35,6 +35,8 @@ if (shouldPatchGlobalTimers) {
 patchProcess();
 handleUnhandledPromiseRejection();
 
+patchDomains();
+
 // Crypto
 let crypto: any;
 try {
@@ -92,4 +94,14 @@ function handleUnhandledPromiseRejection() {
 
   (Zone as any)[zoneSymbol('rejectionHandledHandler')] =
       findProcessPromiseRejectionHandler('rejectionHandled');
+}
+
+// Preserve Zones along with domains
+function patchDomains() {
+  if (process.domain && process.domain.bind) {
+    const originalBind = process.domain.bind;
+    process.domain.constructor.prototype.bind = function(cb: Function) {
+      return originalBind.call(this, Zone.current.wrap(cb, 'domain.bind'));
+    };
+  }
 }


JiaLiPassion:
1. I think you should add test case to test the wrapped process.bind
2. if you patch domain.bind, you should also patch domain.intercept
3. if you patch domain.bind, domain will add some additional properties on Error, such as 
error.domain, because Error is important in zone.js, you should also create test cases to test it.
4. domain is Deprecated and it is just like zone.js to group async operations, I am not sure it is a good idea to patch domain with zone.js

Laguana:
I'll add some test cases, but first let me try to address your other concerns:

My intention was not to provide full support for domains in zone.js, but rather to provide zone.js support for libraries which already support domains. For example, redis has code at https://github.com/NodeRedis/node_redis/blob/master/index.js#L874 which tries to preserve domains for callbacks. If someone is using redis and zone.js, then because redis is not zone-aware it may end up using the wrong zone at some point (perhaps because events are batched together). With this change, the callback will preserve the zone as appropriate, and only if there is a current domain will the domain be preserved.

I am aware that domains are deprecated and will be removed once a replacement is worked out, but that won't be for a long time (looks like Node 8.X doesn't have a replacement in mind yet). When a replacement is decided, then Zone.js should definitely integrate with that, but what about backwards compatibility and the nearer term?

JiaLiPassion:
Yes, redis and mysql use domain, so I think we should also keep them in the correct zone, I am not familiar with domain, I just check the example and source code of domain, I think the callback should already be in the zone. I am not sure how to use domain with zone.js together, can you post your sample code usage?

Laguana:
Looks like writing a test is harder than expected: in the test environment, there is no `process.domain` it seems, even though if I just open up a node console it's definitely there.

My intended code usage is just to take redis or mysql as-is, drop in zone.js (with these changes), and have zones persist appropriately. For some context, I was also looking at mongodb, and in that case all callbacks are called in the context of whatever zone you were in when you initialize the database connection. I've been looking at some monkey-patching to wrap callbacks there and preserve the correct zone. When I looked at redis and mysql, I saw that they already had a mechanism to do exactly what I wanted, it just happened to be done via domains rather than zones, so I thought that this patch would solve those two cases, as well as any other libraries trying to do the right thing.

JiaLiPassion:
@Laguana , I think process.domain should exists even in jasmine test.
and in my imagination, such as mogodb calls like this 

```
 let userColl = this.mongo.db.collection('seFile');
        userColl.findOne({...}, function(result) {});
```

the callback need to be patched, because they are async.

But process.domain.bind is not async.it just add the callback in domain.
just like Zone.wrap

https://github.com/nodejs/node/blob/master/lib/domain.js#L293

So if you can post some sample code to describe that  callback not in zone , we can discuss how to patch it better.

Laguana:
I'm not sure why `process.domain` is missing, but I check for its existence in the test and it's definitely not there.

I agree that in the case of mongo, the callbacks need to be patched, and redis and mysql need similar things. However, my point is that these libraries have already recognized that callbacks need to be patched for domains, and essentially the same thing needs to happen for zone.js

For a specific example where redis fails at the moment: If the connection is not yet established then requests get buffered and dealt with later. This snippet demonstrates that:

```javascript
require("zone.js");

var redis = require("redis");

var z1 = Zone.current.fork({name:'new zone'});
var client = redis.createClient('redis://mycache');

var callback = function () {
    console.log(Zone.current.name);
}

z1.run(function () {
    client.incr('requestCount', callback); // I would expect this to print 'new zone' but it prints '<root>'
});
```

JiaLiPassion:
@Laguana , if that is the case, 

```
z1.run(function () {
    client.incr('requestCount', callback); // I would expect this to print 'new zone' but it prints '<root>'
});
```

I think client.incr should be patched , not domain, patch domain will not help here.

Laguana:
At the moment the code that I have to fix it does make changes to redis (not to incr, but to a more common function that things pass through). mysql doesn't have such a convenient location, which is why I thought to try using the pre-existing infrastructure that it had in place for domains.

However, further experimentation has shown that I've misunderstood how domains work. It seems that running node interactively uses a domain by default, but running `node foo.js` directly does not, so my fix here is not as useful as I had first thought.

JiaLiPassion:
Yes,I understand what you want to achieve,domain does not need to be patched anyway

PR #632, opened 2017-02-12
feat(error): remove zone internal stack frames in error.stack

JiaLiPassion:
1.zone.js patches most basic APIs, so when we want to debug, we will see a lot of zone related stack traces, for example, in example/basic.html, the simple event listener stack trace will look like this.

```
Error: aw shucks
    at HTMLButtonElement.throwError (file:///Users/lijia/workspace/zone.js/example/basic.html:43:11) [long-stack-trace]
    at Zone.runTask (file:///Users/lijia/workspace/zone.js/dist/zone.js:162:47) [<root> => long-stack-trace]
    at HTMLButtonElement.ZoneTask.invoke (file:///Users/lijia/workspace/zone.js/dist/zone.js:356:33) [<root>]
  -------------   Elapsed: 2483 ms; At: Sun Feb 12 2017 16:18:19 GMT+0900 (JST)   -------------  
    at getStacktraceWithUncaughtError (file:///Users/lijia/workspace/zone.js/dist/long-stack-trace-zone.js:33:12) [long-stack-trace]
    at new LongStackTrace (file:///Users/lijia/workspace/zone.js/dist/long-stack-trace-zone.js:27:22) [long-stack-trace]
    at Object.onScheduleTask (file:///Users/lijia/workspace/zone.js/dist/long-stack-trace-zone.js:83:18) [long-stack-trace]
    at ZoneDelegate.scheduleTask (file:///Users/lijia/workspace/zone.js/dist/zone.js:263:49) [long-stack-trace]
    at Zone.scheduleEventTask (file:///Users/lijia/workspace/zone.js/dist/zone.js:182:39) [long-stack-trace]
    at zoneAwareAddListener (file:///Users/lijia/workspace/zone.js/dist/zone.js:1318:14) [long-stack-trace]
    at HTMLButtonElement.addEventListener (eval at createNamedFn (file:///Users/lijia/workspace/zone.js/dist/zone.js:1424:17), <anonymous>:3:43) [long-stack-trace]
    at HTMLButtonElement.bindSecondButton (file:///Users/lijia/workspace/zone.js/example/basic.html:38:8) [long-stack-trace]
    at Zone.runTask (file:///Users/lijia/workspace/zone.js/dist/zone.js:162:47) [<root> => long-stack-trace]
    at HTMLButtonElement.ZoneTask.invoke (file:///Users/lijia/workspace/zone.js/dist/zone.js:356:33) [<root>]
[...]

PR #690, opened 2017-03-21
fix(REVERT): remove zone internal stack frames in error.stack (#632)

mhevery:
This feature was reverted because it was causing a performance regression. 
The setup of all of the stack traces added about 20-25ms to the total 
time of the Zone initialization. This was a bit more than 50% of overall
time zone needed to initialize. While this is a useful feature, the 
performance cost does not justify its cost.


mhevery:
@JiaLiPassion FYI

JiaLiPassion:
@mhevery , 

I have following proposal.

1. should we make it as a helper method like error.getSimpleStack(), and when the method was first called, the detect procedure will be run once. 
2. Or we make it a stand alone js just like longstacktrace, so user can decide to load it or not.

what do you think?

mhevery:
Yes we should brainstorm as to what is the best way to provide this functionality.

PR #693, opened 2017-03-22
feat(error): an idea for simple stack

JiaLiPassion:
An idea to only throw one error to detect most zone stack trace frames.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -1664,7 +1664,6 @@ const Zone: ZoneType = (function(global: any) {
             case FrameType.transition:
               if (zoneFrame.parent) {
                 // This is the special frame where zone changed. Print and process it accordingly
-                // frames[i] += ` [${zoneFrame.parent.zone.name} => ${zoneFrame.zone.name}]`;
                 zoneFrame = zoneFrame.parent;
               } else {
                 zoneFrame = null;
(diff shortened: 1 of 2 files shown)


lib/zone.ts
```diff
@@ -1834,17 +1830,60 @@ const Zone: ZoneType = (function(global: any) {
...
+            (t as any)._transitionTo = fakeTransitionTo;
+            t.invoke();
+          },
+          () => {});
     });
-  };
-  // Cause the error to extract the stack frames.
-  detectZone.runTask(detectZone.scheduleMacroTask('detect', detectRunFn, null, () => null, null));
+  });
+  Error.stackTraceLimit = 15;
```
mhevery: we should store and restore, rather than reset to 15

lib/zone.ts
```diff
@@ -1834,17 +1830,60 @@ const Zone: ZoneType = (function(global: any) {
...
+    }
+  });
+
+  Error.stackTraceLimit = 100;
+  childDetectZone.run(() => {
+    childDetectZone.runGuarded(() => {
+      const fakeTransitionTo =
+          (toState: TaskState, fromState1: TaskState, fromState2: TaskState) => {};
+      childDetectZone.scheduleEventTask(
+          'detect',
```
mhevery: to keep the size down, can we reuse existing string such as `blacklistedStackFramesSymbol`?

lib/zone.ts
JiaLiPassion: Sure, will do.

lib/zone.ts
JiaLiPassion: Yes, you mean remove the duplicate frame from blacklistedstackFrames? I will modify the source and let you review.

JiaLiPassion: Sorry, I am not sure I understand your meaning, could you explain more? thank you.

PR #699, opened 2017-03-24
fix(listener): fix #616, webdriver removeEventListener throw permission denied error

### LINKED ISSUE #616: Protractor tests on Firefox fail for Angular apps using Zone.JS > 0.7.2
All protractor tests are failing on Firefox 50+, using Angular v2.4.4, zone.js v0.7.6, Geckodriver 0.13.0, protractor 4.0.13.

The error is the following:
`Permission denied to access property "invoke"`

This seems similar to https://github.com/angular/protractor/issues/2784 and can be solved only by downgrading Zone.js to v0.7.2.

JiaLiPassion:
fix #616, and https://github.com/mozilla/geckodriver/issues/515.

the issue just like https://github.com/angular/protractor/issues/2784.

In cross site contexts (such as WebDriver frameworks like Selenium), 
access eventHandler will throw permission denied error.

and in  https://github.com/angular/protractor/issues/2784, @juliemr fixed addEventListener, 
in #616, firefox will also call removeEventListener , the code can be found here, 
https://reviewboard.mozilla.org/r/98064/diff/12#index_header

so we need to do the same with addEventListener in removeEventListener to avoid 
such issue.

- And I modify a little about addEventListener/removeEventListener

https://github.com/JiaLiPassion/zone.js/blob/3fae05d347fde7e46cc8834158da74e544506877/lib/common/utils.ts#L320

in earlier fix, if data.handler.toString is permission denied, we do nothing, but in FF52+, toString() is not accessible, but we can still call native addEventListener/removeEventListener with the data.handler, so we can try to do that and try/catch to avoid error.

- I also add test cases which include the test cases from https://github.com/juliemr/zone-firefox-driver/blob/master/test.js provided by @juliemr and from https://github.com/mgiambalvo/firefox-zonejs-issue-demo which provided by @mgiambalvo. Now it can be run locally after install the 
Firefox 52+, webdriver-manager 12.0.4+, selenium-webdriver 3.3.0+, and I will try to add it into travis/saucelabs after saucelabs support those newest browser/webdriver versions.

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -164,6 +164,7 @@ export interface ListenerTaskMeta extends TaskData {
   handler: NestedEventListenerOrEventListenerObject;
   target: any;
   name: string;
+  crossContext: boolean;
   invokeAddFunc: (addFnSymbol: any, delegate: Task|NestedEventListenerOrEventListenerObject) => any;
   invokeRemoveFunc:
       (removeFnSymbol: any, delegate: Task|NestedEventListenerOrEventListenerObject) => any;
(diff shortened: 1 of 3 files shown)


jlin412:
Which version, this PR is going to be in?

JiaLiPassion:
@jlin412 , I think it will be released into the next version(0.8.6 maybe).

PR #704, opened 2017-03-29
fix(error): remove throw in Error constructor to improve performance in IE11

### LINKED ISSUE #691: IE10 is very slow on 0.8.4
I have got this problem when upgrade zone.js from 0.7.2 to 0.8.4

IE is very slow through function `attachZoneAndRemoveInternalZoneFrames(error)`

screen shot 2017-03-21 at 3 52 07 pm

### LINKED ISSUE #698: long-stack-trace throws error in InternetExplorer
[Error.stack is sometimes `undefined` in IE](https://docs.microsoft.com/en-us/scripting/javascript/reference/stack-property-error-javascript). Therefor Zone fails to call trim method on it:
https://github.com/angular/zone.js/blob/e11d9ff6c7b0f0bb27d0efd49f03d74e47ea48c9/dist/long-stack-trace-zone.js#L70

JiaLiPassion:
fix #691.

in pervious modification, error.stack will be constructed in Error constructor, this will impact the performance of IE very badly, so in this PR, the auto-throw is removed.

fix #698.

when Error.stackTraceLimit is set to 0, the stack will be undefined, so we don't need to generate longstacktrace when stackTraceLimit = 0, it will improve performance in some test environment (such as some karma-test-shim for IE)

### DIFF
--- a/lib/zone-spec/long-stack-trace.ts
+++ b/lib/zone-spec/long-stack-trace.ts
@@ -58,7 +58,7 @@ function addErrorStack(lines: string[], error: Error): void {
 }
 
 function renderLongStackTrace(frames: LongStackTrace[], stack: string): string {
-  const longTrace: string[] = [stack.trim()];
+  const longTrace: string[] = [stack ? stack.trim() : ''];
 
   if (frames) {
     let timestamp = new Date().getTime();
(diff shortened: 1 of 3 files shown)


andtii:
@mhevery Could we have a review of this soon? Pretty big performance issue

PR #711, opened 2017-04-01
fix(patch): fix #708, modify the canPatchDescriptor logic when browser has no onreadystatechange

### LINKED ISSUE #708: Can't use zones with jsdom
I'm trying to run jsdom to run my angular2 test with karma.  JsDom tries to set onreadystatechange to null and fails.  The problem exist because jsdom doesn't define onreadystatechange using defineProperty() but zones sets the property as readonly in property-descriptor.ts (line 73).

Adding a setter seems to fix the problem
```
  Object.defineProperty(XMLHttpRequest.prototype, 'onreadystatechange', {
    enumerable: true,
    configurable: true,
    get: function() {
      return this.orsc;
    },
    set: function(f) {
      this.orsc = f;
    }
  });
  req.onreadystatechange = function(){}
  const result = !!req.onreadystatechange;
```

JiaLiPassion:
Fix #708,

in jsdom, XMLHttpRequest has no onreadystatechange, 
in zone.js canPatchDescriptor logic, zone.js use a fake onreadystatechange descriptor which only have a getter, and XMLHttpRequest will use this one, and prevent set onreadystatechange property.

so we need to use a workable onreadystatechange descriptor in this case.

### DIFF
--- a/lib/browser/property-descriptor.ts
+++ b/lib/browser/property-descriptor.ts
@@ -67,19 +67,40 @@ function canPatchViaPropertyDescriptor() {
   // by default XMLHttpRequest.prototype.onreadystatechange is undefined
   // without adding enumerable and configurable will cause onreadystatechange
   // non-configurable
-  Object.defineProperty(XMLHttpRequest.prototype, 'onreadystatechange', {
-    enumerable: true,
-    configurable: true,
-    get: function() {
-      return true;
-    }
-  });
-  const req = new XMLHttpRequest();
-  const result = !!req.onreadystatechange;
-  // restore original desc
-  Object.defineProperty(XMLHttpRequest.prototype, 'onreadystatechange', xhrDesc || {});
-  return result;
-}
+  // and if XMLHttpRequest.prototype.onreadystatechange is undefined,
+  // we should set a real desc instead a fake one
+  if (xhrDesc) {
+    Object.defineProperty(XMLHttpRequest.prototype, 'onreadystatechange', {
+      enumerable: true,
+      configurable: true,
+      get: function() {
+        return true;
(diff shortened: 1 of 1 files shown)


lib/browser/property-descriptor.ts
```diff
@@ -67,18 +67,39 @@ function canPatchViaPropertyDescriptor() {
...
-    }
-  });
-  const req = new XMLHttpRequest();
-  const result = !!req.onreadystatechange;
-  // restore original desc
-  Object.defineProperty(XMLHttpRequest.prototype, 'onreadystatechange', xhrDesc || {});
-  return result;
+  // and if XMLHttpRequest.prototype.onreadystatechange is undefined,
+  // we should set a real desc instead a fake one
+  if (xhrDesc) {
```
DaveMBush: It would probably make sense to just immediately return false if xhrDesc is falsy...

```
if (!xhrDesc) return;
```
this reduces nesting

JiaLiPassion: @DaveMBush , the function canPatchDescriptor will see whether zone.js can patch global object, if return false, zone.js will not patch DOM,XHR, so everything will not in zone. So we have to pick up a global object to try to patch and see the result.

DaveMBush: OK.. I see the else now.

mhevery:
could you rebase so that we can merge it in?

JiaLiPassion:
@mhevery , I have rebased the PR, please review.

abner:
great! @JiaLiPassion   I need this for tests with jest+jsdom.

For now i'm using a patch with your changes

Issue #719, opened 2017-04-04
PhantomJS 2.5 and zone.js 0.8.5: TypeError: undefined is not an object


CSchulz:
I have encountered an issue with PhantomJS 2.5-beta and zone.js 0.8.5:

```
PhantomJS 2.5.0 (Linux 0.0.0) ERROR
  TypeError: undefined is not an object (evaluating '_prop')
  at webpack:///~/zone.js/dist/zone.js:1193:0 <- src/polyfills.ts:9781
```

Version 0.8.4. works fine.

JiaLiPassion:
@CSchulz , can you post a reproduce repo? Or if you can debug, can you check 

```
webpack:///~/zone.js/dist/zone.js:1193:0 <- src/polyfills.ts:9781
```

here, that the `prop` and  `eventName` and `obj` is?

CSchulz:
- eventName: message
- obj: Window
- prop: onmessage

called from Immediate.js:

```
    ImmediateDefinition.prototype.canUsePostMessage = function () {
        var root = this.root;
        // The test against `importScripts` prevents this implementation from being installed inside a web worker,
        // where `root.postMessage` means something completely different and can't be used for this purpose.
        if (root.postMessage && !root.importScripts) {
            var postMessageIsAsynchronous_1 = true;
            var oldOnMessage = root.onmessage;
            root.onmessage = function () {
                postMessageIsAsynchronous_1 = false;
            };
            root.postMessage('', '*');
            root.onmessage = oldOnMessage;
            return postMessageIsAsynchronous_1;
        }
        return false;
    };
```

PR #721, opened 2017-04-04
fix: make getter safe when patching a property

mlaval:
I'm not sure about how to test this case, but it is happening in Safari 10.0 and iOS 10.2 when running the Angular main test campaign:
- iOS 10: https://travis-ci.org/angular/angular/jobs/218405097
- Safari 10: https://travis-ci.org/angular/angular/jobs/218405098

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -106,6 +106,9 @@ export function patchProperty(obj: any, prop: string) {
   // The getter would return undefined for unassigned properties but the default value of an
   // unassigned property is null
   desc.get = function() {
+    if (this == null) {
+      return this;
+    }
     let r = this[_prop] || null;
     // result will be null when use inline event attribute,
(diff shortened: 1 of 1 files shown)


JiaLiPassion:
@mlaval , I think it may related to #719, because now the window object are also patched, and maybe we should return the original get to it, I will try to reproduce it and add test cases.

mlaval:
For information, I can only see this in Safari 10.0 and iOS 10.2 browser.
It doesn't happen in the latest Safari 10.1, and not in any other browser.

JiaLiPassion:
@mlaval, got it, I will try to reproduce it.

JiaLiPassion:
@mlaval , I have made a PR #723 to fix this one, but I can't make angular travis build run now, it will broke when install node.

mlaval:
Closing in favor of #723

PR #722, opened 2017-04-04
fix: resolve errors with closure

vikerman:
Also don't patch all "on" properties if a list of properties is being
passed in. Otherwise various global variables in window get clobbered.

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -128,24 +128,25 @@ export function patchProperty(obj: any, prop: string) {
   };
 
   Object.defineProperty(obj, prop, desc);
-};
+}
 
 export function patchOnProperties(obj: any, properties: string[]) {
-  const onProperties = [];
-  for (const prop in obj) {
-    if (prop.substr(0, 2) == 'on') {
-      onProperties.push(prop);
-    }
-  }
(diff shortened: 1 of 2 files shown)


mhevery:
Could you get this green?

lib/common/utils.ts
```diff
@@ -128,24 +128,25 @@ export function patchProperty(obj: any, prop: string) {
...
-    }
-  }
-  for (let j = 0; j < onProperties.length; j++) {
-    patchProperty(obj, onProperties[j]);
-  }
   if (properties) {
     for (let i = 0; i < properties.length; i++) {
       patchProperty(obj, 'on' + properties[i]);
     }
+  } else {
```
JiaLiPassion: @vikerman , I think we should merge the `properties` array and `prop in obj`  and then remove duplicate ones, because if we use `if/else` here, some properties may not be patched.

so I think the logic should be.
```
 if (properties) {
   for (let i = 0; i < properties.length; i++) {
      const onProp = 'on' + properties[i];
      if (onProperties.filter(op => op === onProp)).length > 0) {
          continue;
      } 
      patchProperty(obj, onProp);
    }
  }
```

PR #728, opened 2017-04-07
fix(core): fix #405, refine the error message when zone.js already loaded

### LINKED ISSUE #405: Uncaught Error: Zone already loaded
I don't really know how it happened but I got the "`Uncaught Error: Zone already loaded`" error in my console after doing some POST petitions. The functionality it's ok, but the look of the app it changes. 

screen shot 22

JiaLiPassion:
fix #405.

We often see that 'Zone.js already loaded' error, when we wrongly load zone.js twice, or load zone.js in some polyfill by incidentally. 

In this PR, we will try to throw the error message as 

```javascript
Zone already loaded. The loaded Zone.js script file seems to be path/to/zone.js(or some vender.bundle.js...)
```

So it will be easy to find out what is wrong, but it require the both zone.js to be upgrade to newer version (maybe 0.8.6)

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -608,7 +608,19 @@ type AmbientZoneDelegate = ZoneDelegate;
 
 const Zone: ZoneType = (function(global: any) {
   if (global['Zone']) {
-    throw new Error('Zone already loaded.');
+    let zoneAlreadyLoadedMessage: string = 'Zone already loaded.';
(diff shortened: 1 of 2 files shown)


mhevery:
@JiaLiPassion Sorry I had to revert this one as it broke the build. Can you reopen this as a new PR?

JiaLiPassion:
@mhevery,got it,I will make a new PR!

PR #730, opened 2017-04-10
fix(core): fix #405, refine the error message when zone.js already loaded.

### LINKED ISSUE #405: Uncaught Error: Zone already loaded

JiaLiPassion:
fix #405 

Remake the PR to pass the CI test.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -609,7 +609,19 @@ type AmbientZoneDelegate = ZoneDelegate;
 
 const Zone: ZoneType = (function(global: any) {
   if (global['Zone']) {
-    throw new Error('Zone already loaded.');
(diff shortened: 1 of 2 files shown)


mhevery:
Hi @JiaLiPassion, We have an issue that zone takes long to load on mobile devices, and so we are looking for ways to make it smaller. I am hesitant to add more code which is not strictly necessary until we can make it smaller/faster to load.

JiaLiPassion:
@mhevery, got it, I will also do some research to find out how to make zone boot faster.

PR #731, opened 2017-04-11
feat(closure): fix #727, add zone_externs.js for closure compiler

### LINKED ISSUE #727: Distribute closure externs for Zone
Since Zone.js is a "prollyfill" and since multiple angular apps might be on a single page, we don't want to bundle it together with the app.
This means closure compiler needs an externs file to avoid renaming Zone's symbols.
I currently have such an externs here:
https://github.com/alexeagle/closure-compiler-angular-bundling/blob/master/vendor/zone_externs.js
but it should be distributed with Zone.js so that users don't have to fetch it themselves.

JiaLiPassion:
1. add closure compiler externs file in dist/zone_externs.js
2. add test script in travis.
now we have 2 cases.

- normal case, can compile and keep the names defined in extern
- fail case, will throw error if the extern file is broken (definition error).

### DIFF
--- a/dist/zone_externs.js
+++ b/dist/zone_externs.js
@@ -1,0 +1,442 @@
+/**
+* @license
+* Copyright Google Inc. All Rights Reserved.
+*
+* Use of this source code is governed by an MIT-style license that can be
+* found in the LICENSE file at https://angular.io/license
+*/
+
+/**
+ * @fileoverview Externs for zone.js
+ * @see https://github.com/angular/zone.js
+ * @externs
+ */
+
+/**
+ * @interface
+ */
+var Zone = function(){};
+/**
+ * @type {!Zone} The parent Zone.
+ */
+Zone.prototype.parent;
+/**
+ * @type {!string} The Zone name (useful for debugging)
+ */
+Zone.prototype.name;
+
+Zone.assertZonePatched = function(){};
+
+/**
+ * @type {!Zone} Returns the current [Zone]. Returns the current zone. The only way to change
+ * the current zone is by invoking a run() method, which will update the current zone for the
+ * duration of the run method callback.
+ */
+Zone.current;
+
+/**
+ * @type {Task} The task associated with the current execution.
+ */
+Zone.currentTask;
+
+/**
+ *  @type {!Zone} Return the root zone.
+ */
+Zone.root;
+
+/**
+ * Returns a value associated with the `key`.
+ *
+ * If the current zone does not have a key, the request is delegated to the parent zone. Use
+ * [ZoneSpec.properties] to configure the set of properties asseciated with the current zone.
+ *
+ * @param {!string} key The key to retrieve.
+ * @returns {?} The value for the key, or `undefined` if not found.
+ */
+Zone.prototype.get = function(key){};
+
+/**
+ * Returns a Zone which defines a `key`.
+ *
+ * Recursively search the parent Zone until a Zone which has a property `key` is found.
+ *
+ * @param {!string} key The key to use for identification of the returned zone.
+ * @returns {?Zone} The Zone which defines the `key`, `null` if not found.
+ */
+Zone.prototype.getZoneWith = function(key){};
+
+/**
+ * Used to create a child zone.
+ *
+ * @param {!ZoneSpec} zoneSpec A set of rules which the child zone should follow.
+ * @returns {!Zone} A new child zone.
+ */
+Zone.prototype.fork = function(zoneSpec){};
+
+/**
+ * Wraps a callback function in a new function which will properly restore the current zone upon
+ * invocation.
+ *
+ * The wrapped function will properly forward `this` as well as `arguments` to the `callback`.
+ *
+ * Before the function is wrapped the zone can intercept the `callback` by declaring
+ * [ZoneSpec.onIntercept].
+ *
+ * @param {!Function} callback the function which will be wrapped in the zone.
+ * @param {!string=} source A unique debug location of the API being wrapped.
+ * @returns {!Function} A function which will invoke the `callback` through [Zone.runGuarded].
+ */
+Zone.prototype.wrap = function(callback, source) {};
+
+/**
+ * Invokes a function in a given zone.
+ *
+ * The invocation of `callback` can be intercepted be declaring [ZoneSpec.onInvoke].
+ *
+ * @param {!Function} callback The function to invoke.
+ * @param {?Object=} applyThis
+ * @param {?Array=} applyArgs
+ * @param {?string=} source A unique debug location of the API being invoked.
+ * @returns {*} Value from the `callback` function.
+ */
+Zone.prototype.run = function(callback, applyThis, applyArgs, source) {};
+
+/**
+ * Invokes a function in a given zone and catches any exceptions.
+ *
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -43,7 +43,9 @@
   "bugs": {
     "url": "https://github.com/angular/zone.js/issues"
   },
-  "dependencies": {},
+  "dependencies": {
```
alexeagle: should be a devDependency right?

lib/zone.ts
```diff
@@ -136,11 +136,11 @@
 interface Zone {
   /**
    *
-   * @returns {Zone} The parent Zone.
+   * @returns {?Zone} The parent Zone.
```
alexeagle: what is the reasoning for updating all the inline types?

If the externs file is independent of the .ts file,
1) we'll have to maintain two different copies of the same info, and
2) nothing enforces the JSDoc annotations in the .ts file are correct, so we will probably fail to maintain them

angular/tsickle can generate externs from the TS types, so we could just remove the JSDoc annotations if we were using the extern generator.

I know you already did a lot of work here, sorry to have to ask this!

scripts/closure/closure_compiler.sh
```diff
@@ -0,0 +1,57 @@
...
+
+OPTS_FAILED=(
+  "--compilation_level=ADVANCED_OPTIMIZATIONS"
+  "--js_output_file=build/closure/closure-bundle.js"
+  "--rewrite_polyfills=false"
+
+  # Flags to simplify debugging.
+  "--formatting=PRETTY_PRINT"
+
+  "--externs=test/closure/zone_externs_failed.js"
```
alexeagle: nit: I'd have one options block with the options that are in common (including the zone.closure.js input file)
then when you call compiler.jar below, give the options that differ between the two cases

scripts/closure/closure_compiler.sh
```diff
@@ -0,0 +1,57 @@
...
+closureFlags=$(mktemp)
+echo ${OPTS[*]} > $closureFlags
+
+closureFailedFlags=$(mktemp)
+echo ${OPTS_FAILED[*]} > $closureFailedFlags
+
+# compile closure test source file
+tsc -p .
+# Run the Google Closure compiler java runnable.
+java -jar node_modules/google-closure-compiler/compiler.jar --flagfile $closureFlags 'build/test/closure/zone.closure.js'
```
alexeagle: nit: I prefer to have the flagfile be an actual file in this directory, then you don't have to use the fancy bash array construct that few people understand

test/closure/zone_externs_failed.js
```diff
@@ -0,0 +1,445 @@
+/**
+* @license
+* Copyright Google Inc. All Rights Reserved.
+*
+* Use of this source code is governed by an MIT-style license that can be
+* found in the LICENSE file at https://angular.io/license
+*/
+
+/**
+ * @fileoverview Externs for zone.js
```
alexeagle: this file also looks like a lot of maintenance work. What is different between this and the other one? Could we get the same test coverage by just calling closure with no externs at all?

package.json
JiaLiPassion: yes, you are right, I will change it.

lib/zone.ts
JiaLiPassion: Thank for the review, I am new to closure compiler, so I just check the document and add the type I can find.

So I just revert all those inline type changes from zone.ts, is that correct?

scripts/closure/closure_compiler.sh
JiaLiPassion: @alexeagle , yes, you are right, I will make it change that way.

test/closure/zone_externs_failed.js
JiaLiPassion: Yeah, you are right , I will change it that way.

JiaLiPassion:
@alexeagle , thank you the review, I will make the changes of the comments you write above, and about the dist/zone_externs.js, is that ok? It can keep the name unchanged, I am not sure the `@type @interface @record` definitions are correct or not, please help me to check it , thank you.

scripts/closure/closure_compiler.sh
JiaLiPassion: yes, I just copied from material, I also prefer use flag directly, I will modify it.

dist/zone_externs.js
```diff
@@ -0,0 +1,442 @@
+/**
+* @license
+* Copyright Google Inc. All Rights Reserved.
+*
+* Use of this source code is governed by an MIT-style license that can be
+* found in the LICENSE file at https://angular.io/license
+*/
+
+/**
+ * @fileoverview Externs for zone.js
```
alexeagle: we can verify that these work in a real project, like overwrite https://github.com/alexeagle/closure-compiler-angular-bundling/blob/master/vendor/zone_externs.js with yours and re-build

package.json
alexeagle: and you can remove the dependencies{} block too

package.json
```diff
@@ -52,6 +53,7 @@
     "concurrently": "^2.2.0",
     "conventional-changelog": "^1.1.0",
     "es6-promise": "^3.0.2",
+    "google-closure-compiler": "^20170218.0.0",
```
alexeagle: a new April release just came out, may as well upgrade this now

scripts/closure/closure_compiler.sh
```diff
@@ -0,0 +1,29 @@
...
+
+if [ $? -eq 0 ]
+then
+  echo "Successfully pass closure compiler with zone externs"
+else
+  echo "failed to pass closure compiler with zone externs" 
+  exit 1
+fi
+
+# Run the Google Closure compiler java runnable with error zone externs.
```
alexeagle: update comment - "without externs"

dist/zone_externs.js
JiaLiPassion: sure, I will check it too

package.json
JiaLiPassion: Ok, got it.

scripts/closure/closure_compiler.sh
JiaLiPassion: Ok, got it.

JiaLiPassion:
@alexeagle, I have modified the comment and update package.json, please review.

scripts/closure/closure_compiler.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Run the Google Closure compiler java runnable without zone externs.
+java -jar node_modules/google-closure-compiler/compiler.jar --flagfile 'scripts/closure/closure_flagfile'
+
+# Check whether the name of Zone exposed API is lost correctly or not
+node build/closure/closure-bundle.js
+
+if [ $? -eq 1 ]
+then
+  echo "Successfully detect closure compiler error with wrong zone externs"
```
alexeagle: here too, wrong -> missing

scripts/closure/closure_compiler.sh
```diff
@@ -0,0 +1,31 @@
...
+else
+  echo "failed to detect closure compiler error with wrong zone externs"
```
alexeagle: and here

JiaLiPassion: @alexeagle , thanks, I will modify it now.

PR #734, opened 2017-04-12
feat(patch): fix #499, let promise instance toString active like native

### LINKED ISSUE #499: ZoneAwarePromise does not validate other Promise polyfills tests
## Version
0.6.25

## Problem

Many external libraries already include a Promise polyfill in their source code, which tests support for Promise before adding their own implementation. The problem is that the test are very specific and ZoneAwarePromise does not validate all of them.

For example, I have this very popular Promise polyfill in one of my lazyloaded dependencies : https://github.com/stefanpenner/es6-promise

The test used to detect Promise support is: 
```
 let P = local.Promise;

  if (P) {
    var promiseToString = null;
    try {
        promiseToString = Object.prototype.toString.call(P.resolve());
    } catch(e) {
        // silently ignored
    }

    if (promiseToString === '[object Promise]' && !P.cast){
        return;
    }
  }
```

With ZoneAwarePromise, `Object.prototype.toString.call(P.resolve());` returns `[object Object]` which makes the test failing. So ZoneAwarePromise is replaced by the polyfill implementation, and change detection does not work anymore after promise resolution.

Is there anything you can do to make this test pass ?

## Workaround

 Currently the only ugly workaround I found is to monkey-patch `Object.prototype.toString` : 
```
const ZoneAwarePromise = self.Promise;
const originalObjectPrototypeToString = Object.prototype.toString;

// TEMP: monkey patch Object.prototype.toString
Object.prototype.toString = function() {
	if(typeof this === "object" && this instanceof ZoneAwarePromise) {
		return "[object Promise]";
	}
	return originalObjectPrototypeToString.apply(this, arguments);
};
```

JiaLiPassion:
fix #499, let 
```
object.prototype.toString(Promise.resolve())
```

active as native to output

```
[object Promise]
```

instead of current

```
[object Object]
```

in ES6, we can use Symbol.toStringTag, in ES5, I didn't find a better way to do this.

### DIFF
--- a/lib/common/to-string.ts
+++ b/lib/common/to-string.ts
@@ -34,3 +34,13 @@ export function patchFuncToString() {
     return originalFunctionToString.apply(this, arguments);
   };
 }
+
+export function patchObjectToString() {
+  const originalObjectToString = Object.prototype.toString;
+  Object.prototype.toString = function() {
+    if (this instanceof Promise) {
+      return '[object Promise]';
+    }
+    return originalObjectToString.apply(this, arguments);
+  };
+}
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -7,7 +7,7 @@
  */
 
 import {patchTimer} from '../common/timers';
-import {patchFuncToString} from '../common/to-string';
+import {patchFuncToString, patchObjectToString} from '../common/to-string';
 import {findEventTask, patchClass, patchEventTargetMethods, patchMethod, patchPrototype, zoneSymbol} from '../common/utils';
 
 import {propertyPatch} from './define-property';
@@ -154,6 +154,8 @@ if (_global['navigator'] && _global['navigator'].geolocation) {
 
 // patch Func.prototype.toString to let them look like native
 patchFuncToString();
+// patch Object.prototype.toString to let them look like native
+patchObjectToString();
 
 // handle unhandled promise rejection
 function findPromiseRejectionHandler(evtName: string) {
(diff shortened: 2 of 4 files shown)


kevindqc:
Thanks! I think this also fixed a problem I had with popper.js in my Angular app. 

https://github.com/FezVrasta/popper.js/blob/master/src/popper/utils/debounce.js
https://github.com/FezVrasta/popper.js/blob/master/src/popper/utils/isNative.js

This is checking if MutationObserver is native or not, and with my testing, prior to zone.js 0.8.6, it ended up using `taskDebounce` instead `microtaskDebounce` because zone.js was overwriting MutationObserver so MutationObserver.toString() was not returning native code. That meant I would see my tooltip in a wrong position for one paint, and it was positioned correctly only on the next paint.

lib/common/to-string.ts
```diff
@@ -34,3 +34,13 @@ export function patchFuncToString() {
     return originalFunctionToString.apply(this, arguments);
   };
 }
+
+export function patchObjectToString() {
+  const originalObjectToString = Object.prototype.toString;
+  Object.prototype.toString = function() {
+    if (this instanceof Promise) {
```
sod: Isn't the reason why Promise shims are in use because Promise is not available in all common browsers today? So wouldn't this line throw
```log
ReferenceError: Can't find variable: Promise
```

for projects & browsers that haven't window.Promise (neither native nor shim)?

JiaLiPassion: `zone.js` has provided `Promise` shim, so it will be ok even some browser or other environment doesn't provide native `Promise`

PR #747, opened 2017-04-22
fix(patch): fix #746, check desc get is null and only patch window.resize additionally

### LINKED ISSUE #741: window.onresize is not intercepted by zones
Repro snippet:

```js
window.onresize = () => console.trace('stack from within onresize handler');
```

Zone check from within the onresize handler:
```
> window.Zone.current
Zone {_properties: Object, _parent: null, _name: "<root>", _zoneDelegate: ZoneDelegate}
```

We came across this via angular/angular#16143

### LINKED ISSUE #746: Cannot read property 'apply' of undefined
- Create a new empty project with @angular/cli 1.0.0
- Update the packages in the package.json to the latest versions ( "zone.js": "^0.8.8" and "@angular/core": "^4.0.3")
- npm install and npm start
- check your browser console (I used Chrome)
untitled

JiaLiPassion:
fix #746,fix #741 

1. #722 break #741, not patch onresize(maybe others), so I add it back in this PR
2. fix #746, check originalDescGet is null or not.

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -126,7 +126,7 @@ export function patchProperty(obj: any, prop: string) {
     }
     if (target.hasOwnProperty(_prop)) {
       return target[_prop];
-    } else {
+    } else if (originalDescGet) {
       // result will be null when use inline event attribute,
       // such as <button onclick="func();">OK</button>
       // because the onclick function is internal raw uncompiled handler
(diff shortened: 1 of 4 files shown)


lib/common/utils.ts
```diff
@@ -126,41 +126,50 @@ export function patchProperty(obj: any, prop: string) {
...
     for (let i = 0; i < properties.length; i++) {
-      patchProperty(obj, 'on' + properties[i]);
-    }
-  } else {
-    const onProperties = [];
-    for (const prop in obj) {
-      if (prop.substr(0, 2) == 'on') {
-        onProperties.push(prop);
-      }
+      const prop = 'on' + properties[i];
```
mhevery: can we have an explicit list rather than iterating over everything? The issue is that closure compiler creates lots of properties on the window with mangled names, and it is quite easy for them to start with `on`

JiaLiPassion: Got it, I will fist add on resize and figure out what others are needed.

JiaLiPassion:
@mhevery , I have updated my PR.
1. only patch specified properties
2. window patch resize property
3. check originalDesc is null or not

please review. thank you !

smorhaim:
I tried the patch, it still gives the same error. Same steps to reproduce. Brand new project, ng serve, error.

JiaLiPassion:
@smorhaim , you can try the attached zone.js, the error still exists because the dist/ folder is not updated.
[dist.zip](https://github.com/angular/zone.js/files/949368/dist.zip)

smorhaim:
@JiaLiPassion it works! Thanks.

sangecz:
is it fixed? which version? i tried downgrade from 0.8.8 to 0.8.5 and 0.8.4 but same issue as described in #6063

dkontorovskyy:
@JiaLiPassion, do you know when /dist folder is going to be updated?

Thanks!

JiaLiPassion:
@dkontorovskyy , it will be the next version release, please wait for a while.

PR #748, opened 2017-04-22
feat: make codebase more modular so that only parts of it can be loaded

mhevery:


mhevery:
@JiaLiPassion I am working on speeding up the loading of the zone as well as making the loading more modular. This is a start, would love to have your feedback.

JiaLiPassion:
@mhevery , got it, thank you for involving me, I will look into it.

lib/zone.ts
```diff
@@ -587,43 +611,22 @@ interface EventTask extends Task {
...
-   */
-  originalStack?: string;
-}
-
 /** @internal */
 type AmbientZone = Zone;
 /** @internal */
 type AmbientZoneDelegate = ZoneDelegate;
 
 const Zone: ZoneType = (function(global: any) {
```
JiaLiPassion: @mhevery , Should we separate the Zone interface and implementation?

lib/zone.ts
```diff
@@ -1273,613 +1238,51 @@ const Zone: ZoneType = (function(global: any) {
...
-      return function() {
-        if (wasCalled) {
-          return;
-        }
-        wasCalled = true;
-        wrappedFunction.apply(null, arguments);
-      };
-    };
+  const patches: {[key: string]: any} = {};
+  const _api: _ZonePrivate = {
```
JiaLiPassion: In my understandings, the ZonePrivate is the API will be used internal, and can be override by patches( such as promise and other modules), I am a little concerned that we expose Private API in this way will require that the module being loaded in certain order.

Such as if we have 3 modules, 

- A
- B
- Promise

And Promise override consoleError, 
And if in the future module A will also override consoleError, it will be problem.

So should we make it like observers?

lib/browser/browser.ts
```diff
@@ -15,168 +15,181 @@ import {eventTargetPatch} from './event-target';
...
-  target: any;
-  args: any[];
-  aborted: boolean;
-}
-
-function patchXHR(window: any) {
-  function findPendingTask(target: any) {
-    const pendingTask: Task = target[XHR_TASK];
-    return pendingTask;
+Zone.__load_patch('timers', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
```
JiaLiPassion: I like this one!!!, this is much more clear!!!
We should also patch other non-standard libraries in this way, such as 

- Notification
- MediaQuery
- 3rd party promise 
...

lib/common/utils.ts
```diff
@@ -133,7 +133,7 @@ export function patchProperty(obj: any, prop: string) {
       // the onclick will be evaluated when first time event was triggered or
       // the property is accessed, https://github.com/angular/zone.js/issues/525
       // so we should use original native get to retrieve the handler
-      let value = originalDescGet.apply(this);
+      let value = originalDescGet && originalDescGet.apply(this);
       value = desc.set.apply(this, [value]);
```
JiaLiPassion: Just for remind, this may need to use the #747 's modification

lib/zone.ts
```diff
@@ -648,6 +651,17 @@ const Zone: ZoneType = (function(global: any) {
       return _currentTask;
     };
 
+    static __load_patch(name: string, fn: _PatchFn): void {
```
JiaLiPassion: should we allow 3rd party add their own patch?
 If so, we may need to consider which Zone Private API should we expose.

lib/zone.ts
```diff
@@ -1229,40 +1223,11 @@ const Zone: ZoneType = (function(global: any) {
...
-      if (handler && typeof handler === 'function') {
-        handler.apply(this, [e]);
-      }
-    } catch (err) {
-    }
+    task && _microTaskQueue.push(task);
   }
 
   function drainMicroTaskQueue() {
+    const showError: boolean = !(Zone as any)[__symbol__('ignoreConsoleErrorUncaughtError')];
```
JiaLiPassion: Do we need to organize all __symbol__ to some const definition block or separate file?
It is a little hard to find what are defined or used...

lib/zone.ts
```diff
@@ -1273,613 +1238,51 @@ const Zone: ZoneType = (function(global: any) {
...
-          try {
-            uncaughtPromiseError.zone.runGuarded(() => {
-              throw uncaughtPromiseError;
-            });
-          } catch (error) {
-            handleUnhandledRejection(error);
+            if ((Zone as any)[__symbol__('ignoreConsoleErrorUncaughtError')]) {
+              return;
+            }
+            _api.consoleError(showError, error);
```
JiaLiPassion: should we define this API as 
```
processUnhandleMicrotaskError(...)
```
or something like that?

lib/zone.ts
mhevery: We could. What would be the advantage?

lib/zone.ts
mhevery: 1) I want to keep in hidden so that you can only get to it from the callback.

2) We could do that, but since we only have one place which needs it so far, I would prefer to have a use case before we complicate things.

lib/browser/browser.ts
mhevery: Yes, that is my plan, just have not gotten to it all.

lib/common/utils.ts
mhevery: #747 is merged, so it should pick it up on rebase.

lib/zone.ts
mhevery: I would prefer to keep this private to zone for now. Would love to see some use cases before we expose it.

lib/zone.ts
mhevery: Yes that sounds like a good idea. Perhaps we could do that as a second pass after this?

lib/zone.ts
mhevery: good, idea changed.

lib/zone.ts
JiaLiPassion: Just to clarify the interface and implementation and reduce the zone.ts size, it is a little big, user may easily check the API from the interface file.

But we have the zone.js.d.ts file, so I think it is also ok in current way.

lib/zone.ts
JiaLiPassion: Got it. thanks.

lib/browser/browser.ts
JiaLiPassion: Got it!!!

lib/common/utils.ts
JiaLiPassion: Sure!

lib/zone.ts
JiaLiPassion: Yes, I agree.

lib/zone.ts
JiaLiPassion: Yeah, sure!

marinho:
I'm experiencing an issue when trying to upgrade to newer Angular/Zone versions and just found out this change caused it. I will try to describe below with details as possible:

in my job, we have to shit apps wrapped by an internal platform that patches `XMLHttpRequest` before Zone.js is loaded. Such library is patching `addEventListener` and other methods and preventing delegation. So, that makes Zone break, as it relies on delegate to let `XMLHttpRequest` be patched via `EventTarget`.

That's not under my team's feasability to change, so, we solved that with this:

```
require('zone.js/lib/zone');
const patchEventTargetMethods = require('zone.js/lib/common/utils').patchEventTargetMethods;
patchEventTargetMethods(XMLHttpRequest.prototype);
require('zone.js/lib/browser/browser');
```

that means we patch `XMLHttpRequest` directly before `browser` is loaded, so when `patchEventTargetMethods` runs on `EventTarget`, it works as expected.

That used to work until 0.8.9.

Between 0.8.10 and 0.8.12, that's no longer possible, as `zone.js/lib/zone.ts` will complain that `Promise` is already patched:

```
Error: Zone.js has detected that ZoneAwarePromise `(window|global).Promise` has been overwritten.
```

Since 0.8.13, `patchEventTargetMethods` was moved into `lib/common/events.ts`, so I face importing errors, when executing `require('zone.js/lib/zone’)` - before my call on `patchEventTargetMethods`:

```
node_modules/zone.js/lib/common/events.ts:53:26
    Cannot find name '_global'.
```

So, as I understand the main purpose of this commit was about making Zone more modular, I guess there might be another way to reach `patchEventTargetMethods` and call it directly for `XMLHttpRequest` before I run into that issue. Or maybe importing a couple of modules instead of just relying on `dist/zone.js`.

Is there any light that could help me solve this issue? Hopefully I’d like to stick to latest Zone versions, so I’m able to upgrade to Angular 5+.

Thanks.

JiaLiPassion:
@marinho , could you describe more about your issue with some code?
```javascript
require('zone.js/lib/zone');
const patchEventTargetMethods = require('zone.js/lib/common/utils').patchEventTargetMethods;
patchEventTargetMethods(XMLHttpRequest.prototype);
require('zone.js/lib/browser/browser');
```

I am not sure why you did this  before `zone.js 0.8.9`, you didn't add your own patch logic. so could you describe when you did `your own internal platform patch`?

PR #752, opened 2017-04-24
fix: remove eval-based createNamedFn

### LINKED ISSUE #749: Remove CSP incompatible Function evals
The offending code is [here](https://github.com/angular/zone.js/blob/7b43b51239d21753f50df04eff3dac6f89e5328c/lib/common/utils.ts#L554-L561).

First, there's no way to turn off this violating implementation, which was what `ngCsp` did in AngularJS 1. Browsers report it, regardless if you're catching the exception or not.

Second, I don't see an clear performance benefit from doing a `Function`. This is basically JS engine details, and unlike the AngularJS 1 this is not a so complicated logic. If it does run faster, I would like some reference for it.

ishitatsuyuki:
Close #749

- there was no way to turn off the bad implementation, causing a CSP violation every time.
- try..catch is quite expensive, defeating the point of using an "optimized" version.

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -551,14 +551,9 @@ export function patchClass(className: string) {
 }
 
 export function createNamedFn(name: string, delegate: (self: any, args: any[]) => any): Function {
-  try {
-    return (Function('f', `return function ${name}(){return f(this, arguments)}`))(delegate);
-  } catch (error) {
-    // if we fail, we must be CSP, just return delegate.
-    return function() {
-      return delegate(this, <any>arguments);
-    };
-  }
+  return function() {
+    return delegate(this, <any>arguments);
+  };
 }
 
 export function patchMethod(


ishitatsuyuki:
Looks like this is harder than I expect.

There's an ES6 way to create dynamically named function on SO: https://stackoverflow.com/questions/9479046/is-there-any-non-eval-way-to-create-a-function-with-a-runtime-determined-name

However, CSP existed from years before, and this isn't a complete fix.

We can also completely throw away the patched name, sacrificing debugging experience.

JiaLiPassion:
This has been fixed in https://github.com/angular/zone.js/pull/748/files#diff-f5e084bcb7c43d977d56c3b791bbca2dR570,I think that is better solution.

ishitatsuyuki:
Looks super cool, feel free to close mine if it resolves the problem.

JiaLiPassion:
@ishitatsuyuki , thanks for posting the issue, you may close this PR and leave the issue #752.

PR #761, opened 2017-04-26
feat: turn more code into Zone modules

mhevery:


lib/common/to-string.ts
```diff
@@ -9,7 +9,8 @@ import {zoneSymbol} from './utils';
 
 // override Function.prototype.toString to make zone.js patched function
 // look like native function
-export function patchFuncToString() {
+Zone.__load_patch('toString', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
+  // patch Func.prototype.toString to let them look like native
   const originalFunctionToString = Function.prototype.toString;
   const g: any =
```
JiaLiPassion: I think we should use global from parameter here.

mhevery: yes, let me fix that.

PR #768, opened 2017-05-01
fix(closure): patchOnProperty with exact eventNames as possible

JiaLiPassion:
related to #722.

We should try to patch onProperties with exact eventNames as possible.

in this PR, I separate the eventNames into several categories.

1. globalEventHandlerEventNames, eventNames in [GlobalEventHandlers](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers)
2. documentEventNames, eventNames in [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document)
3. window, eventNames in [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window)
4. element, eventNames in [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element)
5. webgl, eventNames in [WebGlContextEvent](https://developer.mozilla.org/en-US/docs/Web/API/WebGLContextEvent)
6. form, eventNames in [FormElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement)
7. details, eventNames in [DetailsElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailListElement)
8. XMLHttpRequest, eventNames in [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest)
9. IDBIndex, eventNames in [IDBIndex](https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex)
10. WebSocket, eventNames in [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket)

So we can patch each target with exact eventNames.

For example, we patch HTMLElement with 
`globalEventHandlers + element + webgl + form + details`, 
and we patch Window with
`globalEventHandlers + window`.

### DIFF
--- a/lib/browser/property-descriptor.ts
+++ b/lib/browser/property-descriptor.ts
@@ -10,9 +10,30 @@ import {isBrowser, isMix, isNode, patchClass, patchOnProperties, zoneSymbol} fro
 
 import * as webSocketPatch from './websocket';
 
-const eventNames =
-    'copy cut paste abort blur focus canplay canplaythrough change click contextmenu dblclick drag dragend dragenter dragleave dragover dragstart drop durationchange emptied ended input invalid keydown keypress keyup load loadeddata loadedmetadata loadstart message mousedown mouseenter mouseleave mousemove mouseout mouseover mouseup pause play playing progress ratechange reset scroll seeked seeking select show stalled submit suspend timeupdate volumechange waiting mozfullscreenchange mozfullscreenerror mozpointerlockchange mozpointerlockerror error webglcontextrestored webglcontextlost webglcontextcreationerror'
+const globalEventHandlersEventNames =
+    'abort animationcancel animationend animationiteration auxclick blur cancel canplay canplaythrough change click close contextmenu curechange dblclick drag dragend dragenter dragexit dragleave dragover drop durationchange emptied ended error focus gotpointercapture input invalid keydown keypress keyup load loadstart loadeddata loadedmetadata lostpointercapture mousedown mouseenter mouseleave mousemove mouseout mouseover mouseup mousewheel pause play playing pointercancel pointerdown pointerenter pointerleave pointerlockchange mozpointerlockchange webkitpointerlockerchange pointerlockerror mozpointerlockerror webkitpointerlockerror pointermove pointout pointerover pointerup progress ratechange reset resize scroll seeked seeking select selectionchange selectstart show sort stalled submit suspend timeupdate volumechange touchcancel touchmove touchstart transitioncancel transitionend waiting'
         .split(' ');
+const documentEventNames =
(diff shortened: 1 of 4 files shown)


JiaLiPassion:
@mhevery , just as #722, we may need to patch exact event names as possible instead of use `patchOnProperties(XXX.Prototype, null)`, please review whether the idea is OK or not, I will continue to check the event names.

lib/common/utils.ts
```diff
@@ -148,10 +148,45 @@ export function patchProperty(obj: any, prop: string) {
...
+    }
+  }
+  return validEventNames;
+}
+
 export function patchOnProperties(obj: any, properties: string[]) {
   if (properties) {
-    for (let i = 0; i < properties.length; i++) {
-      patchProperty(obj, 'on' + properties[i]);
+    const validEventNames = getEventNames(obj, properties);
```
mhevery: Could we have a fixed set rather than performing this complex operation on each startup?

JiaLiPassion: @mhevery, yes, in my first commit, my goal is to make a fixed set, but it seems sometimes patch non exist property may cause errors, I will do more research and describe it more clear here.

lib/common/utils.ts
```diff
@@ -60,6 +60,7 @@ export const isMix: boolean = typeof process !== 'undefined' &&
     !!(typeof window !== 'undefined' && (window as any)['HTMLElement']);
 
 export function patchProperty(obj: any, prop: string) {
+  // TODO: @JiaLiPassion, if the desc not exists in obj, should we still patch the property?
```
mhevery: I don't think so. 

The way I imagine this to work is that we have a list of all properties, and than we only patch those properties, but only if the property already exist. But we don't look for extra ones.

JiaLiPassion: @mhevery , got it, thanks! I also think we should do like that, I will modify it and do some tests.

lib/browser/property-descriptor.ts
```diff
@@ -23,24 +230,32 @@ export function propertyDescriptorPatch(_global: any) {
   if (canPatchViaPropertyDescriptor()) {
     // for browsers that we can patch the descriptor:  Chrome & Firefox
     if (isBrowser) {
-      patchOnProperties(window, eventNames.concat(['resize']));
+      patchOnProperties(window, eventNames, Object.getPrototypeOf(window));
```
mhevery: The fact that we are getting all of the properties of the window, does not seem right. Why can't we have a fixed list here as well?

JiaLiPassion: @mhevery , the reason is in IE/Edge, the `onProp` don't exist in `window` but in `WindowPrototype`.

The code below describe the issue.

```javascript
const desc = Object.getOwnPropertyDescriptor(window, 'onload');
// in IE
console.log(desc); // will be undefined.
// in chrome/firefox
console.log(desc); // will be {configurable:true,get/set....}
```

So in new `patchProperty` method, we only patch the prop which exist in the target object, so I pass the prototype of `window` for the existency check.

PR #770, opened 2017-05-02
fix: Package Error stack rewriting as a separate bundle.

mhevery:


example/profiling.html
```diff
@@ -4,6 +4,14 @@
...
   <title>Zones Profiling</title>
   <link rel="stylesheet" href="css/style.css">
+  <script>
+    __Zone_disable_Error = true;
+    __Zone_disable_on_property = true;
+    __Zone_disable_geolocation = true;
+    __Zone_disable_toString = true;
+    __Zone_disable_blocking = true;
+    __Zone_disable_PromiseRejectionEvent = true;
+  </script>
```
JiaLiPassion: @mhevery , will we use those flags to decide which modules to load?

gulpfile.js
```diff
@@ -179,6 +187,8 @@ gulp.task('build', [
   'build/zone.js',
   'build/zone.js.d.ts',
   'build/zone.min.js',
+  'build/zone-error.js',
```
JiaLiPassion: @mhevery , got it, I agree that to separate the error-rewrite to a stand alone module!
I will modify the document to decribe how to load zone.js and other stand alone modules(now we have `error-rewrite` and other non-standard api modules).  Just like the flags in add in `profiling.html`, we may have other stand alone modules (toString, geolocation...).

If we have too many small modules, should we provide some predefined module collections for users to load? 
such as `standard browser set`, `standard node set`, `standard electron set`, `basic set` and so on...

example/profiling.html
mhevery: That is the idea for now. Do you have thoughts or concerns?

gulpfile.js
mhevery: Yes that is my thinking, but i am not sure we should do this just yet. Perhaps disabling things may be good enough for now?

example/profiling.html
JiaLiPassion: @mhevery , only the names, should we named those flags as __zone_symbol__disable_toString?

gulpfile.js
JiaLiPassion: @mhevery , got it!

PR #779, opened 2017-05-15
ensure zoneDelegates is not null before updateTask.

aitboudad:


aitboudad:
somehow it getting null when I use https://github.com/matthieua/WOW in angular also zoneDelegates can be nullable looks at https://github.com/angular/zone.js/blob/master/lib/zone.ts#L879

JiaLiPassion:
@aitboudad , can you provide a reproduce repo? I think just check null will not really resolve the issue.

aitboudad:
ok I'll try

aitboudad:
can't reproduce it anymore since the last release

PR #789, opened 2017-05-21
feat(patch): fix #758, patch cordova.exec success/error with zone.wrap

### LINKED ISSUE #758: Cordova seems to need to be patched
Hey guys, 

So we're doing some heavy Angular work with Cordova and we're running into a few cases where we lose the Zone (and therefore lose change detection) when talking to Cordova. There are two main issues I've found so far:

1. Need to patch `window.cordova.exec()` - success and error callbacks are not run in the correct zone.
2. Need to patch `document.addEventListener()` specifically for the following Cordova events: `pause`, `resume`, `activated`, `documentready` (and possibly more - these are at least the ones for iOS). These events are specifically intercepted by Cordova before reaching the underlying (already-patched) APIs, and also do not run in the correct zone.

I have a patch for the first one which I can add to the thread later, but still working on the second. 

Thoughts on patching Cordova in general though?

Thanks,
Greg

JiaLiPassion:
1. patch cordova.exec success/error callback with Zone.wrap, and cordova.exec may not be async so I didn't patch it as a macroTask.
2. merge fake polyfill (`custom_error`, `fake_cordova`) into a single `test_fake_polyfill` file.

### DIFF
--- a/NON-STANDARD-APIS.md
+++ b/NON-STANDARD-APIS.md
@@ -67,6 +67,24 @@ remove the comment of the following line
 //import './extra/bluebird.spec';
 ```
 
+## Others
+
+* Cordova 
+
+patch `cordova.exec` API
+
+`cordova.exec(success, error, service, action, args);` 
+
+`success` and `error` will be patched with `Zone.wrap`.
+
+to load the patch, you should load in the following order.
+
+```
+<script src="zone.js"></script>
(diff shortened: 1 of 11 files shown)


JiaLiPassion:
@mhevery, sorry, I will update it now.

JiaLiPassion:
@mhevery , I have update the commit ,please review.

PR #795, opened 2017-05-28
[in progress] feat(async): fix #740, use async hooks to wrap nodejs async api

### LINKED ISSUE #740: angular with tsconfig target ES2017 async/await will not work with zone.js
this issue is similar with #715, if we use chrome v8 async/await and compile angular with tsconfig target 'ES2017', then typescript will not generate __awaiter code and use native async/await.
and the following logic will fail

```javascript
(click) = test();
const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
const test = async () => {
      console.log(Zone.current.name) // will output 'angular'
      await delay(100);
      console.log(Zone.current.name) // will output 'root'
    }
```

unlike typescript  transpiler, native async/await will first yield from test, and then call promise.then 
for continuation when await something. So Zone currentFrame will become root.
The sequence of above logic when call `await delay(100)` will look like 

1.  delay function return a ZoneAwarePromise, Zone.current is angular
2.  test function return native Promise which generate from chrome v8 by await, and test() execution is finished.
3. So Zone.currentZone is transite from angular->root
4. ZoneAwarePromise which generated from step1 was chained by called by `Promise.prototype.then`
5. delay timeout is executed, and ZoneAwarePromise resolved
6. the chained Promise is resolved , but the Zone.current is root.

JiaLiPassion:
fix #740, use new  [async_hooks]( https://github.com/nodejs/node/blob/master/lib/async_hooks.js) to wrap nodejs instead of monkey-patch all native APIs.

The basic idea is in `async_hooks`, there are 4 hook apis.

```
init
before
after
destroy
```

in this PR, the API will do those work.

1.  init 

`init` will call `Zone.scheduleTask` to schedule a ZoneTask

2. before 

`before` will call `Zone.beforeRunTask` to create a new ZoneFrame and other work just like `Zone.runTask`

3. after

`after` will call `Zone.afterRunTask` to go back to `parent` ZoneFrame and do other work just like `Zone.runTask`

4. destroy

try to `cancelTask` if not cancelled.

this PR still need a lot of work to do, so I post the PR here just ask for review.

@mhevery , could you review that `node_asynchooks.ts` the whether the idea is ok or not? Thank you!

### DIFF
--- a/lib/node/node_asynchooks.ts
+++ b/lib/node/node_asynchooks.ts
@@ -1,0 +1,91 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * patch nodejs async operations (timer, promise, net...) with
+ * nodejs async_hooks
+ */
+Zone.__load_patch('node_async_hooks', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
+  let async_hooks;
+  const BEFORE_RUN_TASK_STATUS = 'BEFORE_RUN_TASK_STATUS';
+
+  async_hooks = require('async_hooks');
+
+  const idTaskMap: {[key: number]: Task} = (Zone as any)[Zone.__symbol__('nodeTasks')] = {};
+
+  const noop = function() {};
+
+  function init(id: number, provider: string, parentId: number, parentHandle: any) {
+    // @JiaLiPassion, check which tasks are microTask or macroTask
+    //(process as any)._rawDebug('init hook', id , provider);
+    if (provider === 'TIMERWRAP') {
+      return;
+    }
+    // create microTask if 'PROMISE'
+    if (provider === 'PROMISE') {
+      const task = idTaskMap[id] = Zone.current.scheduleMicroTask(provider, noop, null, noop);
+      //(process as any)._rawDebug('after init', id, 'status', task.state);
+      return;
+    }
+    // create macroTask in other cases
+    if (provider === 'Timeout' || provider === 'Immediate' || provider === 'FSREQWRAP') {
+      idTaskMap[id] = Zone.current.scheduleMacroTask(provider, noop, null, noop, noop);
+    }
+  }
+
+  function before(id: number) {
+    //(process as any)._rawDebug('before hook', id);
+    // call Zone.beforeRunTask 
+    const task: Task = idTaskMap[id];
+    if (!task) {
+      return;
+    }
+    (task as any)[Zone.__symbol__(BEFORE_RUN_TASK_STATUS)] = api.beforeRunTask(task.zone, task);
+  }
+
+  function after(id: number) {
+    //(process as any)._rawDebug('after hook', id);
+    const task: Task = idTaskMap[id];
+    if (!task) {
+      return;
+    }
+    const beforeRunTask: BeforeRunTaskStatus = (task as any)[Zone.__symbol__(BEFORE_RUN_TASK_STATUS)]; 
+    if (beforeRunTask) {
+      return;
+    }
+    (task as any)[Zone.__symbol__(BEFORE_RUN_TASK_STATUS)] = null;
+    api.afterRunTask(task.zone, beforeRunTask, task);
+  }
+
+  function destroy(id: number) {
+    // try to cancel the task if is not canceled
+    const task: Task = idTaskMap[id];
+    if (task && task.state === 'scheduled') {
+      task.zone.cancelTask(task);
+    }
+    idTaskMap[id] = null;
+  }
+
+  process.on('uncaughtException', (err: any) => {
+    const task = Zone.currentTask; 
+    if (task) {
+      const beforeRunTask: BeforeRunTaskStatus = (task as any)[Zone.__symbol__(BEFORE_RUN_TASK_STATUS)]; 
+      if (beforeRunTask) {
+        if ((task.zone as any)._zoneDelegate.handleError(Zone.current, err)) {
+          throw err;
+        }
+      }
+    }
+  });
+
+  global[Zone.__symbol__('setTimeout')] = global.setTimeout;
+  global[Zone.__symbol__('setInterval')] = global.setInterval;
+  global[Zone.__symbol__('setImmediate')] = global.setImmediate;
+
+  async_hooks.createHook({ init, before, after, destroy }).enable();
+});
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -313,6 +313,12 @@ interface ZoneType {
 type _PatchFn = (global: Window, Zone: ZoneType, api: _ZonePrivate) => void;
 
 /** @internal */
+interface BeforeRunTaskStatus {
+  reEntryGuard: boolean;
+  previousTask: Task;
+}
+
+/** @internal */
 interface _ZonePrivate {
   currentZoneFrame: () => _ZoneFrame;
   symbol: (name: string) => string;
@@ -323,6 +329,8 @@ interface _ZonePrivate {
   patchEventTargetMethods:
       (obj: any, addFnName?: string, removeFnName?: string, metaCreator?: any) => boolean;
   patchOnProperties: (obj: any, properties: string[]) => void;
+  beforeRunTask: (zone: Zone, task: Task) => BeforeRunTaskStatus;
+  afterRunTask: (zone: Zone, beforeRunTaskStatus: BeforeRunTaskStatus, task: Task) => void;
 }
 
 /** @internal */
@@ -757,8 +765,7 @@ const Zone: ZoneType = (function(global: any) {
       }
     }
 
-
-    runTask(task: Task, applyThis?: any, applyArgs?: any): any {
+    beforeRunTask(task: Task) {
       if (task.zone != this) {
         throw new Error(
             'A task can only be run in the zone of creation! (Creation: ' +
@@ -772,7 +779,7 @@ const Zone: ZoneType = (function(global: any) {
       // typescript compiler will complain below
       const isNotScheduled = task.state === notScheduled;
       if (isNotScheduled && task.type === eventTask) {
-        return;
+        return null;
       }
 
       const reEntryGuard = task.state != running;
@@ -781,6 +788,33 @@ const Zone: ZoneType = (function(global: any) {
       const previousTask = _currentTask;
       _currentTask = task;
       _currentZoneFrame = {parent: _currentZoneFrame, zone: this};
+      //(process as any)._rawDebug('currentFrame increase ', _currentZoneFrame && _currentZoneFrame.zone.name, task.source);
+      return {
+        reEntryGuard: reEntryGuard,
+        previousTask: previousTask
+      }
+    }
+
+    afterRunTask(beforeRunTaskStatus: BeforeRunTaskStatus, task: Task) {
+      // if the task's state is notScheduled or unknown, then it has already been cancelled
+      // we should not reset the state to scheduled
+      if (task.state !== notScheduled && task.state !== unknown) {
+        if (task.type == eventTask || (task.data && task.data.isPeriodic)) {
+          beforeRunTaskStatus.reEntryGuard && (task as ZoneTask<any>)._transitionTo(scheduled, running);
+        } else {
+          task.runCount = 0;
+          this._updateTaskCount(task as ZoneTask<any>, -1);
+          beforeRunTaskStatus.reEntryGuard &&
+            (task as ZoneTask<any>)._transitionTo(notScheduled, running, notScheduled);
+        }
+      }
+      _currentZoneFrame = _currentZoneFrame.parent;
+      //(process as any)._rawDebug('currentFrame decrease ', _currentZoneFrame && _currentZoneFrame.zone.name, task.source);
+      _currentTask = beforeRunTaskStatus.previousTask;
+    }
+
+    runTask(task: Task, applyThis?: any, applyArgs?: any): any {
+      const beforeRunTaskStatus = this.beforeRunTask(task);
       try {
         if (task.type == macroTask && task.data && !task.data.isPeriodic) {
           task.cancelFn = null;
@@ -793,20 +827,7 @@ const Zone: ZoneType = (function(global: any) {
           }
         }
       } finally {
-        // if the task's state is notScheduled or unknown, then it has already been cancelled
-        // we should not reset the state to scheduled
-        if (task.state !== notScheduled && task.state !== unknown) {
-          if (task.type == eventTask || (task.data && task.data.isPeriodic)) {
-            reEntryGuard && (task as ZoneTask<any>)._transitionTo(scheduled, running);
-          } else {
-            task.runCount = 0;
-            this._updateTaskCount(task as ZoneTask<any>, -1);
-            reEntryGuard &&
-                (task as ZoneTask<any>)._transitionTo(notScheduled, running, notScheduled);
-          }
-        }
-        _currentZoneFrame = _currentZoneFrame.parent;
-        _currentTask = previousTask;
+        this.afterRunTask(beforeRunTaskStatus, task);
       }
     }
 
@@ -1241,10 +1262,13 @@ const Zone: ZoneType = (function(global: any) {
     // we must bootstrap the initial task creation by manually scheduling the drain
     if (_numberOfNestedTaskFrames === 0 && _microTaskQueue.length === 0) {
       // We are not running in Task, so we need to kickstart the microtask queue.
+      // @JiaLiPassion, use native promise if async_hooks is available
       if (global[symbolPromise]) {
         global[symbolPromise].resolve(0)[symbolThen](drainMicroTaskQueue);
-      } else {
+      } else if (global[symbolSetTimeout]) {
         global[symbolSetTimeout](drainMicroTaskQueue, 0);
+      } else {
+        Promise.resolve(0).then(drainMicroTaskQueue);
       }
     }
     task && _microTaskQueue.push(task);
@@ -1294,7 +1318,13 @@ const Zone: ZoneType = (function(global: any) {
     scheduleMicroTask: scheduleMicroTask,
     showUncaughtError: () => !(Zone as any)[__symbol__('ignoreConsoleErrorUncaughtError')],
     patchEventTargetMethods: () => false,
-    patchOnProperties: noop
+    patchOnProperties: noop,
+    beforeRunTask: (zone: Zone, task: Task) => {
+      return zone.beforeRunTask(task);
+    },
+    afterRunTask: (zone: Zone, beforeRunTaskStatus: BeforeRunTaskStatus, task: Task) => {
+      return zone.afterRunTask(beforeRunTaskStatus, task);
+    }
   };
   let _currentZoneFrame: _ZoneFrame = {parent: null, zone: new Zone(null, null)};
   let _currentTask: Task = null;
@@ -1305,7 +1335,6 @@ const Zone: ZoneType = (function(global: any) {
   function __symbol__(name: string) {
     return '__zone_symbol__' + name;
   }
-
 
   performanceMeasure('Zone', 'Zone');
   return global['Zone'] = Zone;
--- a/test/common/Promise.spec.ts
+++ b/test/common/Promise.spec.ts
@@ -9,12 +9,22 @@
 import {ifEnvSupports} from '../test-util';
 declare const global: any;
 
+let useZoneAwarePromise: boolean = true; 
+try {
+  Zone.assertZonePatched();
+} catch (error) {
+  useZoneAwarePromise = false;
+}
+
 class MicroTaskQueueZoneSpec implements ZoneSpec {
   name: string = 'MicroTaskQueue';
   queue: MicroTask[] = [];
   properties = {queue: this.queue, flush: this.flush.bind(this)};
 
   flush() {
+    if (!useZoneAwarePromise) {
+      return;
+    }
     while (this.queue.length) {
       const task = this.queue.shift();
       task.invoke();
@@ -109,7 +119,7 @@ describe(
         });
       });
 
-      it('should allow sync resolution of promises', () => {
+      it('should allow sync resolution of promises', (done) => {
         queueZone.run(() => {
           const flush = Zone.current.get('flush');
           const queue = Zone.current.get('queue');
@@ -126,11 +136,14 @@ describe(
           expect(queue.length).toEqual(1);
           expect(log).toEqual([]);
           flush();
-          expect(log).toEqual(['RValue', 'second value']);
+          setTimeout(() => {
+            expect(log).toEqual(['RValue', 'second value']);
+            done();
+          }, 0);
         });
       });
 
-      it('should allow sync resolution of promises returning promises', () => {
+      it('should allow sync resolution of promises returning promises', (done) => {
         queueZone.run(() => {
           const flush = Zone.current.get('flush');
           const queue = Zone.current.get('queue');
@@ -147,7 +160,10 @@ describe(
           expect(queue.length).toEqual(1);
           expect(log).toEqual([]);
           flush();
-          expect(log).toEqual(['RValue', 'second value']);
+          setTimeout(() => {
+            expect(log).toEqual(['RValue', 'second value']);
+            done();
+          }, 0);
         });
       });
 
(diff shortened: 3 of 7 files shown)


test/common/Promise.spec.ts
```diff
@@ -9,12 +9,22 @@
...
+  useZoneAwarePromise = false;
+}
+
 class MicroTaskQueueZoneSpec implements ZoneSpec {
   name: string = 'MicroTaskQueue';
   queue: MicroTask[] = [];
   properties = {queue: this.queue, flush: this.flush.bind(this)};
 
   flush() {
+    if (!useZoneAwarePromise) {
```
mhevery: Why do we need to disable the tests? I would imagine that the tests should still work even with async_hook?

test/common/Promise.spec.ts
```diff
@@ -126,11 +136,14 @@ describe(
           expect(queue.length).toEqual(1);
           expect(log).toEqual([]);
           flush();
-          expect(log).toEqual(['RValue', 'second value']);
+          setTimeout(() => {
```
mhevery: Why are these needed? Is it because the `flush()` no longer works?

lib/zone.ts
```diff
@@ -323,6 +329,8 @@ interface _ZonePrivate {
   patchEventTargetMethods:
       (obj: any, addFnName?: string, removeFnName?: string, metaCreator?: any) => boolean;
   patchOnProperties: (obj: any, properties: string[]) => void;
+  beforeRunTask: (zone: Zone, task: Task) => BeforeRunTaskStatus;
```
mhevery: I am not a fan of `beforeRunTask` and `afterRunTask`, but I do understand why you need them. But I think it would mean that we should also add them to `ZoneDelegate`? Or are you planning on making them private?

How do we communicate to the developer that this particular Task can't be wrapped? It never has `runTask` executed on it. It can never be in `fakeAsync`zone.

lib/node/node_asynchooks.ts
```diff
@@ -0,0 +1,91 @@
...
+      const beforeRunTask: BeforeRunTaskStatus = (task as any)[Zone.__symbol__(BEFORE_RUN_TASK_STATUS)]; 
+      if (beforeRunTask) {
+        if ((task.zone as any)._zoneDelegate.handleError(Zone.current, err)) {
+          throw err;
+        }
+      }
+    }
+  });
+
+  global[Zone.__symbol__('setTimeout')] = global.setTimeout;
```
mhevery: Why are these needed here? Seems out of place

lib/zone.ts
```diff
@@ -1294,7 +1318,13 @@ const Zone: ZoneType = (function(global: any) {
     scheduleMicroTask: scheduleMicroTask,
     showUncaughtError: () => !(Zone as any)[__symbol__('ignoreConsoleErrorUncaughtError')],
     patchEventTargetMethods: () => false,
-    patchOnProperties: noop
+    patchOnProperties: noop,
+    beforeRunTask: (zone: Zone, task: Task) => {
```
mhevery: I see you are making them private, but than we should somehow hide them from the public API.

lib/node/node_asynchooks.ts
```diff
@@ -0,0 +1,91 @@
...
+    (task as any)[Zone.__symbol__(BEFORE_RUN_TASK_STATUS)] = api.beforeRunTask(task.zone, task);
+  }
+
+  function after(id: number) {
+    //(process as any)._rawDebug('after hook', id);
+    const task: Task = idTaskMap[id];
+    if (!task) {
+      return;
+    }
+    const beforeRunTask: BeforeRunTaskStatus = (task as any)[Zone.__symbol__(BEFORE_RUN_TASK_STATUS)];
```
mhevery: `Task` has `data`. We should put `BeforeRunTaskStatus` there, rather than monkey patching it on `Task` directly.

test/common/Promise.spec.ts
JiaLiPassion: @mhevery , sorry, I am still debug those test codes, all tests should be able to run with async_hook, but some tests need to be modified.

test/common/Promise.spec.ts
JiaLiPassion: yes, just like the above one, I am just began to debug those tests, I will finish modifying those tests and let you review.

lib/node/node_asynchooks.ts
JiaLiPassion: yes, those will be deleted finally, I just want to make some test pass, and I can debug them easily. I will remove them later.

lib/zone.ts
JiaLiPassion: yes, those APIs are private.

And I will check `fakeAsync` , I think the `runTask` will be called because in `fakeAsyncZone`, the callback is `task.invoke` and it will finally call `task.zone.runTask`. I will check it.

lib/zone.ts
JiaLiPassion: Do you mean we should move the `beforeRunTask` and `afterRunTask` from class Zone so they will not be accessible from outside such as `zone.beforeRunTask`?

lib/node/node_asynchooks.ts
JiaLiPassion: Got it, I will change it.

JiaLiPassion:
waiting for `nodejs` `async_hooks` to provide parent promise information.

https://github.com/nodejs/node/issues/13302

matthewloring:
@JiaLiPassion We are also interested in getting Zones working on top of async hooks avoiding monkey patching if possible and would be happy to help. It looks like some zone APIs require a handle to the callback associated with an async task (onIntercept, onInvoke, etc). Did you have thoughts on how these callbacks could be observed using async hooks without monkey patching?

JiaLiPassion:
@mhevery , @matthewloring , other things that may change when use `async_hooks` instead of `monkey-patch`.

1. `Promise`, onScheduleTask timing.

```
const zone = Zone.current.fork({
  name: 'promise',
  onScheduleTask: (...) => {
    console.log('scheduleTask', task.source);
   }
});

zone.run(() => {
  const p1 = new Promise(resolve => { setTimeout(() => {
    console.log('resolve promise p1');
    resolve();
  }, 100)});

  const p2 = p1.then(v => return v);
});

```

in the current `monkey-patch ZoneAwarePromise` version.
the output will be 

```
resolve promise p1
scheduleTask promise // this is p2
```

and if we use `async_hooks`, the output will be 

```
scheduleTask promise // this is p2
resolve promise p1
```

Because in `ZoneAwarePromise`, a microTask will be scheduled when it's `parent promise` resolved/rejected, and in `async_hooks`, a microTask will be scheduled when a `child promise` was created(when promise.then was called).

2. onHandleError will be a little risky.

just like @matthewloring  said, we have no `callback handle`, so we can't 

```
try {
  task.callback.apply(target, args);
} catch (err) {
  zone.handleError(err);
}
```

we have to use 

```
  process.on('uncaughtException', (err: any) => {

     // get current zone
     Zone.current.handleError(err);
  }
```

JiaLiPassion:
@matthewloring , continue to discuss about the `callback handle`, in my understanding, `async_hooks` not like `monkey-patch`,  it is only `notification`, it will not allow the `hook listener` to control the `lifecycle` of the `async callback`.

so in `async_hooks`,

```javascript
setTimeout(() => {
    console.log('timeout callback')'
});
```

will only trigger `before` and `after`, and the `callback` will be invoked anyway.

but in `zone.js` monkey-patch, 

```
onInvokeTask(...) {
  // zone can decide invoke callback or not
}
```

So I think from the design, `async_hooks` will not support pass `callback handle`, even it does, for example pass the `callback handle` in `before` hook, the hook listener can only reference the `callback handle` information, but can't control the `callback` will be invoked or not, is that correct?

matthewloring:
That is correct, async hooks allows for observation of async events while zones relies on being able to reschedule async work. This was my primary concern with refactoring zones to use async hooks. One possibility would be to have two "modes" supported by zones. A fast, observation-only mode based on async hooks that supported async/await and a slower, monkey-patching based mode which allows rescheduling async tasks but does not have async/await support. This might be too invasive a change (@mhevery what are your thoughts?) but would that address your use case @JiaLiPassion?

JiaLiPassion:
@matthewloring 

> One possibility would be to have two "modes" supported by zones. A fast, observation-only mode based on async hooks that supported async/await and a slower, monkey-patching based mode which allows rescheduling async tasks but does not have async/await support.

Yeah, your suggestion could be a good solution.

So it will be `zone.js` provide 2 modes, and let application select by some configurations.

- fast observation-only mode, only provide `before`, `after` callback, it will resolve the most use cases, but application can't not control the `callback handle`.

- original `monkey-patch` mode without `native async/await` support.

And about the original `monkey-patch` mode, I think maybe we can use `promise async_hooks` to support `native async/await`. so it will be a mix-mode. most API will be `monkey-patched` and promise will be `monkey-patched` and use`async_hooks` to resolve `native async/await` issue.

mhevery:
I am on vacation till end of month. But your discussion sounds reasonable to me.

ofrobots:
If it has two 'modes', would it make sense to split the library into two?

JiaLiPassion:
@ofrobots, yeah, maybe build a `zone.js` and `zone-async-hooks` maybe more clear,but I am not sure about this one...

mhevery:
> @ofrobots, yeah, maybe build a zone.js and zone-async-hooks maybe more clear,but I am not sure about this one...

@JiaLiPassion we could already do this with `Zone.__load_patch` api.

JiaLiPassion:
@mhevery 

`@JiaLiPassion we could already do this with Zone.__load_patch api.`

So you mean we load 

```javascript
Zone.__load_patch('node'); 
Zone.__load_patch('async_hooks`);
```

and add some checks to prevent both them are loaded, is that correct?

mhevery:
@JiaLiPassion that is what I had in mind

JiaLiPassion:
@mhevery , got it, I will continue to work on it after the performance one is done.

matthewloring:
@mhevery Is it expected behavior for zones to only support a subset of its API depending on the set of patches loaded using `Zone.__load_patch`?

JiaLiPassion:
@matthewloring , 

> Is it expected behavior for zones to only support a subset of its API depending on the set of patches loaded using Zone.__load_patch?

In my understanding, it is, you can find all the `modules` which are loaded by `zone.js` with `__load_patch` [here](https://github.com/angular/zone.js/blob/master/MODULE.md), you can disable any of them.

matthewloring:
@JiaLiPassion Looking at the list of patches, it seems like they make zones aware of more async constructs but still support the full zones API. That is, the set of tasks that zones are aware of increases but if I attach an onIntercept function it will still behave correctly for all tasks. If an async hooks plugin were added and a user attempts to reschedule all task callbacks using onIntercept then all tasks corresponding to async hook events will ignore the results of the onIntercept callback if I understand correctly.

JiaLiPassion:
@matthewloring , I want to make sure I understand your meaning correctly.

1. current zone.js `onIntercept` callback will not fire for `tasks`. It will only fire when you use `zone.wrap` like this.


```javascript
const foo = function() {
  console.log('foo');
};

const fooInZone = Zone.current.fork({
  name: 'test',
  onIntercept: function (..., zone, callback) {
     console.log('callback intercepted');
     return callback;
   }
}).wrap(foo);

fooInZone();
```

And in this code, `onIntercept` will be triggered.
But for other `ZoneTasks`, such as `setTimeout`, `Promise`, `XMLHttpRequest`, the `onIntercept` will not be invoked.

```javascript
const fooInZone = Zone.current.fork({
  name: 'test',
  onIntercept: function (..., zone, callback) {
     console.log('callback intercepted');
     return callback;
   }
}).run(() => {
  setTimeout(() => {
}, 1000);
});
```

In this code, `onIntercept` will not be triggered.

2.

> a user attempts to reschedule all task callbacks

About reschedule, if you talk about `onIntercept`, do you mean to return a new callback other than the original one?

```javascript
const foo = function() {
  console.log('foo');
};

const fooInZone = Zone.current.fork({
  name: 'test',
  onIntercept: function (..., zone, callback) {
     console.log('callback intercepted');

    // here return a new callback
     return function() {
       console.log('new foo');
     };
   }
}).wrap(foo);

fooInZone();
```

do you mean some use case like the above code?

And for `ZoneTask` (setTimeout, promise...), `zone.js` currently provide a way to reschedule in new zone. For example, 

```javascript
const barZone = Zone.current.fork({
  name: 'bar'
});
const fooInZone = Zone.current.fork({
  name: 'foo',
  onScheduleTask: function (zoneDelegate, currentZone, targetZone, task) {
      if (task.source === 'something') {
       // I want the task to be scheduled in bar zone
       task.cancelScheduleRequest();
       return barZone.scheduleTask(task);
      }
   }
}).wrap(foo);

fooInZone();
```

And I would like to confirm 

> If an async hooks plugin were added and a user attempts to reschedule all task callbacks using onIntercept then all tasks corresponding to async hook events will ignore the results of the onIntercept callback if I understand correctly.

what is the `ignore the results` you mean?

JiaLiPassion:
@matthewloring , I just read the `PromiseHooks` again, https://docs.google.com/document/d/1rda3yKGHimKIhg5YeoAmCOtyURgsbTH_qaYR79FELlk/edit#, and could you also confirm this one? https://github.com/nodejs/node/issues/14038, the issue is,

## case1: promise's parent is an resolved promise.
```javascript
const p1 = Promise.resolve();

const p2 = p1.then(r => r);
```

In this case, current `async_hooks` API is ok,  I can just `schedule a microTask` in `zone.js` in the `init` callback.

```javascript
function init(id: number, provider: string, triggerId: number, parentHandle: any) {
    if (provider === PROMISE_PROVIDER) {
      if (!parentHandle) {
        return;
      }
      const parentId = parentHandle.parentId;
      if (!parentId) {
        return;
      }

      const task = Zone.current.scheduleMicroTask(PROMISE_SOURCE, noop, null, noop);
      promiseTasks[id] = task;
    }
  }
```

## case2: the promise's parent is an unresolved promise.

```javascript
const p1 = new Promise((resolve, reject) => {
    setTimeout(resolve, 3000);
});

const p2 = p1.then(r => r);
```

In this case, the `microTask` should not be scheduled in `init`, when the `p1.then` is called, because `p1` is still `unresolved`, the `microtask` should be scheduled when `p1.resolve` is 
called, so I wonder is that possible to expose the `PromiseHook.resolve` callback through `async_hooks`?  

And only expose `PromiseHook.resolve` seems can't resolve this issue, I also want to know what `state` promise in to check the following case.

```
const p1 = new Promise(() =>{});
const p2 = new Promise((resolve) => setTimeout(resolve, 1000, p1)); // resolve p2 with p1 after 1s
```

in this case, `resolve (p2)` hook will be called, but `p2` is still in `pending` state, so what I really want is `when promise is really resolved/rejected` through some hook API.

Thank you very much!

JiaLiPassion:
@mhevery , @matthewloring 

I will summarize the known issues here.

1. native async/await issue
2. async_hooks only support readonly notification mode
3. async_hooks onHandleError issue

## Native async/await issue. 

Now with `async_hooks` API, we don't know when to `scheduleMicroTask`. Even `async_hooks` expose the `PromiseHooks` API, https://docs.google.com/document/d/1rda3yKGHimKIhg5YeoAmCOtyURgsbTH_qaYR79FELlk, we still don't know when.

The issue can be described as 

### case1: promise's parent is an resolved promise.
```javascript
const p1 = Promise.resolve();

const p2 = p1.then(r => r);
```

In this case, current `async_hooks` API is ok,  I can just `schedule a microTask` in `zone.js` in the `init` callback.

```javascript
function init(id: number, provider: string, triggerId: number, parentHandle: any) {
    if (provider === PROMISE_PROVIDER) {
      if (!parentHandle) {
        return;
      }
      const parentId = parentHandle.parentId;
      if (!parentId) {
        return;
      }

      const task = Zone.current.scheduleMicroTask(PROMISE_SOURCE, noop, null, noop);
      promiseTasks[id] = task;
    }
  }
```

### case2: the promise's parent is an unresolved promise.

```javascript
const p1 = new Promise((resolve, reject) => {
    setTimeout(resolve, 3000);
});

const p2 = p1.then(r => r);
```

In this case, the `microTask` should not be scheduled in `init`, when the `p1.then` is called, because `p1` is still `unresolved`, the `microtask` should be scheduled when `p1.resolve` is 
called, so I wonder is that possible to expose the `PromiseHook.resolve` callback through `async_hooks`?  

And only expose `PromiseHook.resolve` seems can't resolve this issue, I also want to know what `state` promise in to check the following case.

```
const p1 = new Promise(() =>{});
const p2 = new Promise((resolve) => setTimeout(resolve, 1000, p1)); // resolve p2 with p1 after 1s
```

in this case, `resolve (p2)` hook will be called, but `p2` is still in `pending` state, so what I really want is `when promise is really resolved/rejected` through some hook API.

## Async hooks only support notification mode.

As we discussed, async hooks can't get the control of `callback`, so we need to implement a notification version of zone.js.

## onHandleError 

 Because we can't get the control of callback, so we are not be able to do  monkey-patch like 
the code below.

```javascript
try {
  task.callback.apply(target, args);
} catch (err) {
  zone.handleError(err);
}
```

we have to use

```javascript
  process.on('uncaughtException', (err: any) => {

     // get current zone
     Zone.current.handleError(err);
  }
```

It will become a little risky.

AndreasMadsen:
>  so I wonder is that possible to expose the PromiseHook.resolve callback through async_hooks?

Hello from `async_hooks`. This is something we are discussing, please see https://github.com/nodejs/node/issues/13437.

JiaLiPassion:
@AndreasMadsen , thank you very much, and to implement `zone.js` with `promiseHooks`, the current `PromiseHooks` API seems not enough, not only I want to know `when resolve is called`, but also I need to know `when the promise is really resolved/rejected` so I can schedule microTasks for all the `chained promises` (promise.then).

```javascript
const p1 = new Promise(() => {}); // This will never be resolved
const p2 = new Promise((resolve) => setTimeout(resolve, 1000, p1)); // resolve p2 with p1 after 1s
```

In this case, the `microTask` should not be scheduled in `resolve hook`, when the `resolve(p1)` is called, because `p1` is still `unresolved`, the `microtask` should be scheduled when `p1.resolve` is called(in this case, p1 is never resolved).

in my understanding the `v8 promise.then and resolve` implementation looks like.
```javascript
function then(onFulfilled, onRejected) {
  const chainedPromise = new Promise();
  if (this.state == UNRESOLVED) {
    // current promise not resolved, keep callback into a list
    this[thenList].push(chainPromise, onFulfilled, onRejected);
  } else {
    // current promise is resolved, we can enqueue the chained promise to microTask queue
    // here we really schedule an async microTask
   // in zone.js I need to know when promise get here
    enqueue(chainedPromise);
   }
}

function resolvePromise(promise, value) {
    // here the resolve hook will be triggered
   resolveHook(promise);
   if (isPromise(value) && value.state == UNRESOLVED) {
     // promise still unresolved
     value.then(...);
     return;
   } 
   
   // value is thenable, so promise is not really resolved
   if (isThenable(value)) {
     value.then(...);
     return;
   }

   chainedPromiseList = getThenList(promise);

   // promise was really resolved, we enqueue to start a microTask for chained promise
   // in zone.js I need to know when promise get here
   chainedPromiseList.forEach(chainedPromise => enqueue(chainedPromise))
}
```

 **so I wonder is that possible to 

- expose the `PromiseHook.resolve` callback through `async_hooks` and in the hook we can get the `state of promise` (unresolved, resolved, rejected)**

AndreasMadsen:
I think you are using a technical vocabulary that is specific to zone.js, so I don't really understand what you are saying about `microTask` and `resolve hook`. I will try and comment on the specific question:

> so I wonder is that possible to expose the `PromiseHook.resolve` callback through `async_hooks`

yes, this is what nodejs/node#13437 is about.

> and in the hook we can get the state of promise (unresolved, resolved, rejected).

The `resource.promise` value contains the `promise` object itself. Unfortunately, there is no public API for checking the state of that `promise` object. However, it is possible:

```js
const binding = process.binding('util');
const state2string = new Map([
  [binding.kPending, 'pending'],
  [binding.kFulfilled, 'fulfilled'],
  [binding.kRejected, 'rejected']
]);

const [state, result] = binding.getPromiseDetails(resource.promise);
const readableState = state2string.get(state);
console.log(readableState); // fulfilled
```

I think this was explained in https://github.com/nodejs/node/issues/14038#issuecomment-312490623

I have no idea if `binding.getPromiseDetails` is something we want to make public, it is not really my area. But if we do it won't be though `async_hooks` as it is too promise specific.

JiaLiPassion:
@AndreasMadsen , thank you very much!

> I think you are using a technical vocabulary that is specific to zone.js, so I don't really understand what you are saying about microTask and resolve hook. I will try and comment on the specific questions:

sorry for poor explanation.

```javascript
const p1 = new Promise(() => {}); // This will never be resolved
const p2 = new Promise((resolve) => setTimeout(resolve, 1000, p1)); // resolve p2 with p1 after 1s
```

in this case, `resolve p2 with p1 after 1s` will trigger `PromiseResolve hooks`, but p1 and p2 are still unresolved. for `zone.js`, I need a trigger to tell that the promise is really resolved, not just a trigger to tell that `resolvePromise` function was called.

And about get state of promise.

```javascript
const binding = process.binding('util');
const state2string = new Map([
  [binding.kPending, 'pending'],
  [binding.kFulfilled, 'fulfilled'],
  [binding.kRejected, 'rejected']
]);

const [state, result] = binding.getPromiseDetails(resource.promise);
const readableState = state2string.get(state);
console.log(readableState); // fulfilled

```

Thank you for the information, 

- will this API be gone in `nodejs future version` ?
- will this get state call have performance impact?

AndreasMadsen:
>  I need a trigger to tell that the promise is really resolved, not just a trigger to tell that resolvePromise function was called.

I see, have you experimented with using `resource.promise.then()`?

> will this API be gone in nodejs future version ?

You should assume so. If you have a good use case, you should open an issue in node.js where you request this feature is made public.

> will this get state call have performance impact?

Most C++ calls have, but you will have to benchmark it to be sure. The internal function is used in `util.inspect`, which is not something we are heavily optimizing as it is mostly used for debugging.

JiaLiPassion:
@AndreasMadsen 

> I see, have you experimented with using resource.promise.then()?

the timing of promise.then() is late, and a really want is `after parent promise resolved` and `before enqueue the chained promise to micro tasks queue`.

in my understandings, the code look like this.

```javascript
resolvePromise(promise, value) {
   promise.state = RESOLVED;
   // here I want a trigger.
   trigger(PROMISE_RESOLVED);
   // push the chained promise to microTask (process.tick?) queue.
   enqueueToMicroTask(getChainedPromises(promise));
}
```

And `promise.then(callback)`, the callback will be executed in microTask queue (process.tick?), so I really want the trigger before the then callback was pushed into the microTask queue.

 

> will this API be gone in nodejs future version ?
> You should assume so. If you have a good use case, you should open an issue in node.js where you request this feature is made public.

Got it, I will try to make some cases to describe my requirements.

> 
> will this get state call have performance impact?
> Most C++ calls have, but you will have to benchmark it to be sure. The internal function is used in util.inspect, which is not something we are heavily optimizing as it is mostly used for debugging.

Got it, thank you, I will make benchmark to test it.

jpsfs:
Any news on this?

JiaLiPassion:
@jpsfs, currently I am waiting for https://bugs.chromium.org/p/v8/issues/detail?id=6862 to resolve the `promise` issue.

JiaLiPassion:
## Update on 2018/01/04

### Support es2017 async/await

Because currently `asynchooks promiseResolve` hooks can't tell whether the `promise` is really resolved or not.  I can't use `promiseResolve` to implement `native async/await`, I use another approach. which is override `native promise.then` and use `process.nextTick` to `schedule` another microTask.

the idea can be described as the following code.

```javascript
init(id, triggerId, parentHandle) {
   const promise = parentHandle.promise;
   idPromises[id] = Zone.current; // save id and zone mapping
    promise.then = function(onResolve: any, onReject: any) {
      const task = zone.scheduleMicroTask(PROMISE_PROVIDER, noop, null, noop);
      isPromiseTick = true;
      process.nextTick(() => {
        task.zone.runTask(task, null, null);
        originalThen.call(this, onResolve, onReject);
      });
    };
}

before(id) {
  const zone = getZoneById(); 
  setCurrentZoneFrame(zone); // retrieve the zone by id
}

after(id) {
  setCurrentZoneFrame(null); // leave zone
}
```

and consider the following use case.

```javascript
    const zone = Zone.current.fork({name: 'promise'});
    async function asyncOutside() {
      return 'asyncOutside';
    }

    zone.run(async() => {
      const outside = await asyncOutside();  
      expect(outside).toEqual('asyncOutside');
      expect(Zone.current.name).toEqual(zone.name);
  })
```

with `es2017` target compile and the new `asynchooks` module, the invoke sequence will be.
1. zone.run 
2. await asyncOutside();
3. a new promise was initialized 
4. asynchooks.init was invoked, the aysncId for example `100` and Zone.current (promise) will be kept in memory.
5. zone.run finished, Zone.current become `<root>`
6. await is about to finished, asynchooks.before was invoked with id `100`. so we recover the zone with `promise`.
7. expect(outside).toEqual('asyncOutside'); expect(Zone.current.name).toEqual('promise') will be executed in `promise` zone.
8. asynchooks.after is invoked with id `100`, then we exit `promise` zone and go back to it's parent zone `<root>`.

JiaLiPassion:
## Update on 2018/01/04

### Support `asynchooks`.

Use `asynchooks` instead of `monkey-patch`,for example `fs`, `setTimeout` will use `native asynchooks`, but not all APIs, `EventEmitter`, `console` still need to be patched.

in this PR, 

```shell
  gulp test/node/asynchooks
```

will run all tests for node in `asynchooks` mode.

AndreasMadsen:
@JiaLiPassion Be aware that the V8 team would like to remove `parentHandle.promise`, but I will try and push for `promiseResolve` showing the "resolved" value as a requirement for that (it is much easier if you are actually using `parentHandle.promise`).

The tracking won't be as easy as getting a specific notification, but I think it is much easier for the V8 team to implement and I would like to add another promise-specific event to `async_hooks`. With the "resolved" value, you can simply check if it is a promise. Since you are already tracking that promise, you can simply see if and how that was resolved.

JiaLiPassion:
@AndreasMadsen , thank you for the information.  If a `promiseResolve` call with a `value` parameter, it will solve my problem, I don't need to schedule an additional `microTask` to simulate a `micro async` operation.

And `asynchooks` are not designed to handle `EventEmitter`, is that correct? because in `zone.js`, `EventEmitter.on` is also some kind of `async` operation.

AndreasMadsen:
> And asynchooks are not designed to handle EventEmitter, is that correct? because in zone.js, EventEmitter.on is also some kind of async operation.

That is correct. `EventEmitter`'s are just an interface they don't actually task an asynchronous operation themselves. They are also pretty easy to monkey-patch as the API is very fixed. We actually do it for `domain` in nodecore: https://github.com/nodejs/node/blob/master/lib/domain.js#L400

JiaLiPassion:
@AndreasMadsen , got it, now `zone.js` also `monkey-patch` the `EventEmitter` api to `memorize` zone information, I will keep it that way.

PR #800, opened 2017-06-03
fix(xhr): inner onreadystatechange should not use zoneAwareListener

JiaLiPassion:
fix https://github.com/angular/angular/issues/17167,
zone.js will patch xhr, and add onreadystatechange internally,
this onreadystatechange just for invoke the xhr MacroTask, so it should
not trigger ZoneSpec's callback.

in this commit, use original native addEventListener/removeEventListener
to add internal onreadystatechange event listener.

```javascript
if (listener) {
  // don't use patched removeEventListener
  oriRemoveListener.apply(data.target, ['readystatechange', listener]);
}
const newListener = data.target[XHR_LISTENER] = () => {
    if (data.target.readyState === data.target.DONE) {
    // sometimes on some browsers XMLHttpRequest will fire onreadystatechange with
    // readyState=4 multiple times, so we need to check task state here
    if (!data.aborted && (XMLHttpRequest as any)[XHR_SCHEDULED] &&
        task.state === 'scheduled') {
      task.invoke();   
    }
  }
};      
// don't use patched addEventListener
oriAddListener.apply(data.target, ['readystatechange', newListener]);
```

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -91,8 +91,11 @@ Zone.__load_patch('XHR', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
       const data = <XHROptions>task.data;
       // remove existing event listener
       const listener = data.target[XHR_LISTENER];
+      const oriAddListener = data.target[zoneSymbol('addEventListener')];
+      const oriRemoveListener = data.target[zoneSymbol('removeEventListener')];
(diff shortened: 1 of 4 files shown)


MattJeanes:
This seems to have broken IE support (at least from my testing), raised an issue here: #816

PR #805, opened 2017-06-07
fix(spec): fakeAsyncTestSpec should handle requestAnimationFrame

### LINKED ISSUE #804: requestAnimationFrame is not handled by FakeAsyncZoneSpec
FakeAsyncZoneSpec should handle all async task scheduling without passing it on to the parent zone delegate.

However for requestAnimationFrame it is passing the task to the parent zone.

We should just simulate requestAnimationFrame in fake async zone using setTimeout of 16ms(~60 frames per second).

vikerman:
Fixes #804.

### DIFF
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -297,8 +297,15 @@
               break;
             case 'XMLHttpRequest.send':
               throw new Error('Cannot make XHRs from within a fake async test.');
+            case 'requestAnimationFrame':
+            case 'webkitRequestAnimationFrame':
+            case 'mozRequestAnimationFrame':
+              // Simulate a requestAnimationFrame by using a setTimeout with 16 ms.
+              // (60 frames per second)
+              task.data['handleId'] = this._setTimeout(task.invoke, 16, (task.data as any)['args']);
+              break;
             default:
-              task = delegate.scheduleTask(target, task);
+              throw new Error('Unknown macroTask scheduled in fake async test: ' + task.source);
           }
           break;
         case 'eventTask':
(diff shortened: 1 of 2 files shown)


lib/zone-spec/fake-async-test.ts
```diff
@@ -297,8 +297,15 @@
...
+            case 'requestAnimationFrame':
+            case 'webkitRequestAnimationFrame':
+            case 'mozRequestAnimationFrame':
+              // Simulate a requestAnimationFrame by using a setTimeout with 16 ms.
+              // (60 frames per second)
+              task.data['handleId'] = this._setTimeout(task.invoke, 16, (task.data as any)['args']);
+              break;
             default:
-              task = delegate.scheduleTask(target, task);
+              throw new Error('Unknown macroTask scheduled in fake async test: ' + task.source);
```
juliemr: Could you mention this fix in the commit description as well?

JiaLiPassion: @vikerman , @juliemr, I have one question that if we throw Error, `unknown macroTask` will not be able to be run in `fakeAsyncZone`, currently beside `setTimeout/Interval` and `requestAnimationFrame`, other zone patched macroTasks include `setImmediate` and `nodejs fs/crypto` methods, should we also make them available in the `fakeAsyncTest`? 

thank you!

mhevery:
LGTM, @vikerman can you address @JiaLiPassion concern?

lib/zone-spec/fake-async-test.ts
juliemr: I'd say that we should consider whether it's meaningful to make them available - but in the meantime, throwing an error is much more useful and clearer than silently passing through.

JiaLiPassion: @juliemr, thank you for reply.

> I'd say that we should consider whether it's meaningful to make them available - but in the meantime, throwing an error is much more useful and clearer than silently passing through.

Got it! I agree, if currently we don't need to make those API available, throw error is more making sense.

PR #812, opened 2017-06-11
feat(eventListener): fix #798, improve EventTarget.addEventListener performance

### LINKED ISSUE #798: Slow addEventListener performance.
# Issue

Currently calling `addEventListener` and `removeEventListener` is significantly slower than native version. This has a negative impact on Angular framework performance.

# Root Cause

Given
```
function userCallback() {....}
div.addEventListener('click', userCallback);
div.removeEventListener('click', userCallback);
```

In order to make the zone propagate the `userCallback` has to be wrapped in `zoneAwareUserCallback` like so:

```
var zoneAwareUserCallback = makeZoneAware(userCallback);
div.addEventListener('click', zoneAwareUserCallback);
div.removeEventListener('click', zoneAwareUserCallback);
```

Notice that when the `removeEventListener` gets called it needs to get called with `zoneAwareUserCallback` rather than `userCallback`.

The issue is that the `addEventListener` has to do the translation from `userCallback` to `zoneAwareUserCallback` internally (its monkey patched). And the same translation has to happen in `removeEventListener`. This translation has to be memoized for this is to work. 

Currently the cost of creating the delegate and book keeping is the source of the slowdown.
The current bookkeeping involves creating an array which is patched on the `EventTarget` instance which contains all of the `userCallback`s. For each one there is a corresponding `zoneAwareUserCallback`. When subsequent call is made to the `removeEventListener` the array is first checked to see if it contains the callback, and if so we reuse its zone-aware version.

## Naive Approach

You may ask why is this not sufficient.
```
function userCallback() {....}
userCallback.zoneAware = makeZoneAware(userCallback);
```
This would simply patch the zone-aware function on the function itself and it would greatly simplify the bookkeeping. The problem is that the zone-aware function is bound to a specific Zone. This means that it can't be reused for different zones. 

This code would break:
```
function userCallback() {....}
zoneA.run(() => div.addEventListener('click', userCallback);
zoneB.run(() => div.addEventListener('click', userCallback);
```
The reason is that zone-aware callback is bound to zone, hence it would restore the wrong zone.

# Suggested fix: Modification of Naive Fix.

Since the callback is zone specific, instead of patching to a well known name, we could patch to a name which is zone specific. (This assumes that each zone would have a private unique symbol)

```
function userCallback() {....}
zoneA.run(() => {
  userCallback[Zone.current.__private_symbol]= makeZoneAware(userCallback);
});
zoneB.run(() => {
  userCallback[Zone.current.__private_symbol]= makeZoneAware(userCallback);
});
```

This would greatly simplify the book-keeping and memory pressure which is the source of the performance slowdown.

Internally `addEventListener`/`removeEventListener` can easily determine how to translate from user to zone-aware callback.

### Related Code
- https://github.com/angular/zone.js/blob/master/lib/browser/event-target.ts

JiaLiPassion:
fix #798.

1. improve `EventTarget.addEventListener` performance with @mhevery 's idea which described in #798, we will not save a `eventTask` list on `event target`, and will not go through this `eventTask` list to find whether we have already registered the event or not. 

instead we save the information to `callback` itself, so 

```
div.addEventListener('click', () => {}); 
```

will not slow down

```
div.addEventListener('mouseover', () => {}); 
```

2. this solution only work in browser, in `nodejs` we still use the original way to save `taskList`, 
because in the `eventListener` callback, there is no `eventName` information, and `nodejs` support to add the identical listener multiple times.


mhevery:
I added a performance test here: https://github.com/angular/zone.js/commit/3a017d3fb10a9bcd6fb08e9e9527b9c5d0d53402
Just run `gulp build` and than open the file in the browser to see how fast it is. In DevTools go to `Performance` tab and record the click.  Do `bottom-up` view to see what is going on.

screen shot 2017-06-12 at 2 53 40 am

Right away two things jump out. 
1. `Function.toString` should not be called (I left a comment where it is coming from.)
2. I see try-catch which needs to go (left a comment)
3. Way to much GC

The GC problem means that we are still creating too many objects. Can we do an audit and see if every one of these objects is needed?

mhevery:
The issue is coming from `zoneAwareAddListener`

```TypeScript
    try {
      // In cross site contexts (such as WebDriver frameworks like Selenium),
      // accessing the handler object here will cause an exception to be thrown which
      // will fail tests prematurely.
      validZoneHandler = data.handler && data.handler.toString() === '[object FunctionWrapper]';
    } catch (error) {
      // we can still try to add the data.handler even we are in cross site context
      data.crossContext = true;
      return data.invokeAddFunc(addFnSymbol, data.handler);
    }
```

1. we have to remove the `try-catch` since that is preventing the code from optimizing.
2. Inside we do a check by calling data.handler.toString, this is expensive and has to go.

JiaLiPassion:
@mhevery , thank you for review!

I tried several approaches and did some test.

`native add/remove same callback` performance: 140ms

## Approach 1. current PR version.

`add/remove same callback` performance: `1400~1600ms (10x slower than native).`

## Approach 2. comment the `try/catch` and `toString` code below. 

```javascript
  try {
      // In cross site contexts (such as WebDriver frameworks like Selenium),
      // accessing the handler object here will cause an exception to be thrown which
      // will fail tests prematurely.
      validZoneHandler = data.handler && data.handler.toString() === '[object FunctionWrapper]';
    } catch (error) {
      // we can still try to add the data.handler even we are in cross site context
      data.crossContext = true;
      return data.invokeAddFunc(addFnSymbol, data.handler);
    }
```

`add/remove same callback` performance: `1300ms (9x slower than native)`.
And need to consider the `cross context` issue.

## Approach 3. remove `generic code`, make `zoneAwareCallback` global.

1.  comment the `try/catch` and `toString` code.
2. and make `zoneAwareCallback` be global. (all event target will share one zoneAwareCallback). so we will not create a `zoneAwareCallback function` for each `addEventListener`.
3. remove all `generic` code such as `patchMethod`, `ListenerMeta` and so on...

```javascript
// all event target will share this callback.
const globalZoneAwareCallback = function(event: Event) {
  const target = this || _global;
  const task = target['__zone_symbol__' + event.type][FALSE_STR];
  if (task) {
    return task.invoke.apply(target, [event]);
  }
  
}

// all event target will share this callback with capture
const globalZoneAwareCaptureCallback = function(event: Event) {
  const target = this || _global;
  const task = target['__zone_symbol__' + event.type][TRUE_STR];
  if (task) {
    return task.invoke.apply(target, [event]);
  }
 
}

// define some global string
const TRUE_STR = 'true';
const FALSE_STR = 'false';

// define native delegate
let nativeAddEventListener = (EventTarget.prototype as any)['__zone_symbol__addEventListener'] = EventTarget.prototype.addEventListener;
let nativeRemoveEventListener = (EventTarget.prototype as any)['__zone_symbol__removeEventListener'] = EventTarget.prototype.removeEventListener;


const customSchedule = function(task: Task) {
  let data: any = task.data as any;
  return nativeAddEventListener.apply(data.target, [data.event, data.zoneAwareCallback, data.options]);
}

const customCancel = function(task: Task) {
  let data: any = task.data as any;
  return nativeRemoveEventListener.apply(data.target, [data.event, data.zoneAwareCallback, data.options]);
}

export function patchEventTargetMethodsOptimized() {
  EventTarget.prototype.addEventListener = function() {
    const target = this || _global;
    let capture = arguments.length > 2 ? (typeof arguments[2] === 'boolean' ? arguments[2] : !!arguments[2].capture) : false;
    let events = target['__zone_symbol__' + arguments[0]];
    if (events) {
      const task = events[capture ? TRUE_STR : FALSE_STR];
      if (task) {
        return nativeAddEventListener.apply(target, [arguments[0], capture ? globalZoneAwareCaptureCallback : globalZoneAwareCallback, arguments.length > 2 ? arguments[2] : undefined]);
      }
    } else {
      target['__zone_symbol__' + arguments[0]] = events = {};
    }  
    events[capture ? TRUE_STR : FALSE_STR] = Zone.current.scheduleEventTask(eventName, delegate, {event, capture, zoneAwareCallback}, customSchedule, customCancel); 
  };

  EventTarget.prototype.removeEventListener = function() {
    const target = this || _global;
    let capture = arguments.length > 2 ? (typeof arguments[2] === 'boolean' ? arguments[2] : !!arguments[2].capture) : false;
    const task = target['__zone_symbol__' + arguments[0]][capture ? TRUE_STR : FALSE_STR];
    if (task) {
      target['__zone_symbol__' + arguments[0]][capture ? TRUE_STR : FALSE_STR] = undefined;
      task.zone.cancelTask(task);
    } else {
      nativeRemoveEventListener.apply(target, arguments);
    }
    
  }
}

```

So the performance will become from `500 ~ 700 ms`. (4x slower than native)

## Approach 4. don't schedule task at all.

base on the  above approach , I want to see how much performance can boost by `remove task`, so the code will just like.

```javascript
// global zoneAwareCallback.
const globalZoneAwareCallback = function(event: Event) {
  const target = this || _global;
 
  const zoneWithCallback = target['__zone_symbol__' + event.type][FALSE_STR];
  if (zoneWithCallback) {
    zoneWithCallback.zone.runGuarded(zoneWithCallback.callback, this || _global, [event.type, zoneWithCallback.callback, zoneWithCallback.options]);
  }
}

EventTarget.prototype.addEventListener = function() {
    const target = this || _global;
    let capture = arguments.length > 2 ? (typeof arguments[2] === 'boolean' ? arguments[2] : !!arguments[2].capture) : false;
    let events = target['__zone_symbol__' + arguments[0]];
    if (!events) {   
      target['__zone_symbol__' + arguments[0]] = events = {};
    }
    events[capture ? TRUE_STR : FALSE_STR] = {zone: Zone.current, callback: arguments[1], options: arguments.length > 2 ? arguments[2] : undefined};
    
    return nativeAddEventListener.apply(target, [eventName, zoneAwareCallback, options]);
  };
```

So the performance will become `260~350ms` (2x slower than native).

## Approach 5. schedule fake `eventTask`

based on `Approach 4`, I think maybe we can create a `fake eventTask` which can also be shared globally, and we just call `scheduleTask`, `invokeTask`, `cancelTask` callback of `zoneSpec` directly in `patched method`. Because `eventTask` will not affect `taskCount/hasTask`,and `task status` is also not that useful, so if we can create a pure fake `eventTask` which only for `trigger zoneSpec's callback`, maybe we can get the same performance as  Approach 4.

Please review.

mhevery:
@JiaLiPassion WOW, very impressive analysis! Kudos for such amazing work and insight!

Yes approach 4/5 seems a way to go. It is our best hope to decrease the memory pressure and increases registration speed.

In approach 5 how would task cancelation/rescheduling work? We have to make sure that this would continue to work.

```TypeScript
const someZone = Zone.current.fork({
  name: 'someZone',
  onScheduleTask:
      (delegate: ZoneDelegate, current: Zone, target: Zone, task: Task):
          Task => {
            // Blacklist scroll, mouse, and request animation frame events.
            if ((task.type === 'eventTask' || task.type === 'macroTask') &&
                isBlacklistedEvent(task.source)) {
              task.cancelScheduleRequest();

              // Schedule task in root zone, note Zone.root != target,
              // "target" Zone is Angular. Scheduling a task within Zone.root
              // will prevent the infinite digest cycle from appearing.
              return Zone.root.scheduleTask(task);
            } else {
              return delegate.scheduleTask(target, task);
            }
          }
});```

**Comment by JiaLiPassion** [comment]
@mhevery , got it, I will try to find out Approach 5 work or not.

**Comment by JiaLiPassion** [comment]
Progress.

Use Approach 5. still schedule event task, now the performance data will be 

`350~450` ms, it will be 3x `native version`, and I will continue to add some optimization for `zone.scheduleEventTask` to remove some not needed `try/catch` or `new Function`

**Comment by mhevery** [comment]
awesome progress...

**Comment by JiaLiPassion** [comment]
@mhevery , thank you! currently all the test has passed, please review.

And I will continue to check other tasks such as `macroTask` or `microTask` can improve performance or not in another PR.

File: lib/browser/event-target.ts [hunk 1]
```diff
@@ -31,6 +33,326 @@ export function eventTargetPatch(_global: any) {
...
+
+const TRUE_STR = 'true';
+const FALSE_STR = 'false';
+// an identifier to tell ZoneTask do not create a new invoke closure
+const OPTIMIZED_ZONE_EVENT_TASK = zoneSymbol('optimizedZoneEventTask');
+
+const zoneSymbolEventNames: any = {};
+const globalSources: any = {};
+
+const _global: any =
```
mhevery: Could we get this from the patch function global? It would be better if we checked for global in exactly one place.

lib/browser/event-target.ts
```diff
@@ -31,6 +33,326 @@ export function eventTargetPatch(_global: any) {
...
+    typeof window !== 'undefined' && window || typeof self !== 'undefined' && self || global;
+
+const isDisableIECheck = _global['__Zone_disable_IE_check'] || false;
+const isEnableCrossContextCheck = _global['__Zone_enable_cross_context_check'] || false;
+const ieOrEdge = isIEOrEdge();
+const FUNCTION_WRAPPER = '[object FunctionWrapper]';
+const BROWSER_TOOLS = 'function __BROWSERTOOLS_CONSOLE_SAFEFUNC() { [native code] }';
+
+//  predefine all __zone_symbol__ + eventName + true/false string
+eventNames.forEach((eventName: string) => {
```
mhevery: could we use `for` loop for perf reasons?

lib/browser/event-target.ts
```diff
@@ -31,6 +33,326 @@ export function eventTargetPatch(_global: any) {
...
+  const trueEventName = eventName + TRUE_STR;
+  const symbol = '__zone_symbol__' + falseEventName;
+  const symbolCapture = '__zone_symbol__' + trueEventName;
+  zoneSymbolEventNames[eventName] = {};
+  zoneSymbolEventNames[eventName][FALSE_STR] = symbol;
+  zoneSymbolEventNames[eventName][TRUE_STR] = symbolCapture;
+});
+
+//  predefine all task.source string
+WTF_ISSUE_555.split(',').forEach((target: string) => {
```
mhevery: could we use `for` loop for perf reasons?

mhevery:
also thank you for the awesome work on investigating the perf and getting it down.

lib/browser/event-target.ts
JiaLiPassion: yeah, I will move this code into `__load_patch`

lib/browser/event-target.ts
JiaLiPassion: yes, will do.

lib/browser/event-target.ts
JiaLiPassion: yes, will do

JiaLiPassion:
@mhevery , I have updated the code.

> Minor changes requested
> It seems like we have added a lot of new code but have not deleted existing code. What is going?
> Could we somehow get this better packaged into __load_patch so that it is cleaner?

- the changes you requested have been implemented.
- the `old code` still exists because it will needed by `nodejs EventEmitter`, nodejs EventEmitter currently can't use this `optimized` way, because in the `event listener`, there is not `eventName`.

```javascript
target.addListener('msg', function() {
   // here we can not get the event type which should be `msg`.
});
``` 

so we can't use `globalCallback`.  We have to create a `closure callback` for every `addListener` to save the `eventName` information.

But we can still benefit from other improvements, so I will move this PR's new `addEventListener/removeEventListener` into `common` and remove the old code.

- And I have moved the code into __load_patch.

I will continue to make this code also work for `nodejs`.

JiaLiPassion:
@mhevery , I have updated the source.

1. remove old `addEventListener/removeEventListener` code.
2. make nodejs work in this new method.
And nodejs also get a 40% performance up.

- current version. 

new callback: 700ms 
reuse callback : 550ms

- improved version.

new callback: 440ms
reuse callback: 300ms

3. add new API for `browser`, now `EventTarget` support `eventListenerList` method to get all listeners for all event name or for specified event name

4. add new API for `browser` to support `removeAllListeners`

Please review, Thank you.

Issue #234, opened 2016-01-25
Zone.js is incompatible with FirefoxDriver


juliemr:
FireFox's webdriver implementation uses a firefox add-on which [adds event listeners](https://github.com/SeleniumHQ/selenium/blob/ab1e647d0fc8fc39e6b00ae94321ab228b6728f2/javascript/firefox-driver/extension/content/server.js#L36) to the page. Because the functions in these event listeners come from the add-on content script, they are not accessible by scripts from the webpage (see https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/Content_Scripts/Interacting_with_page_scripts). 

Since zone patches `window.addEventListener` and tries to process the handler, this causes any test to fail with "Permission denied to access property toString".

I have a repo which shows this error at https://github.com/juliemr/zone-firefox-driver

A potential fix would be something like [this](https://github.com/juliemr/zone-firefox-driver/commit/c6563e95a1bddb1399f75572a70396aa200b8682). I could prepare a PR if that looks reasonable.

valorkin:
+1 this is blocking feature for e2e angular2 tests

spinfooser:
I can verify that this is an issue as well. I've been trying to use FluentLenium (which uses the WebDriver Selenium framework) to automate testing of Angular2 websites. This seems to be a Firefox-specific because of the browser design itself like juliemr has said.

Switching to chrome or another browser web driver is the current work-around for the issue. I made a quick pull request to formalize the changes that I would personally make, but I'm not particularly attached to them.

https://github.com/angular/zone.js/pull/264

uglow:
Switching to Chrome is not an option for those projects running Protractor on TravisCI. We really need a fix for Firefox.

spinfooser:
Yes, it needs to be fixed. That's really unfortunate that switching to something else isn't an option. Lets hope these pull requests are investigated soon! (It seems like there is a log jam)

uglow:
@spinfooser - and you can believe me when I tell you I've tried my utmost to get Chrome to work on TravisCI. Something to do with their version of Linux means that the latest version of Chromium that they can install (or is installed) is 2(?) years old. Firefox was the recommended approach, or using SauceLabs for Protractor testing.

rinokadijk:
+1

spinfooser:
Currently I have in a passing pull request after the big 0.6 Zone.js update. I think this will fix the issue and have tested it locally.

https://github.com/angular/zone.js/pull/309

[...]

PR #819, opened 2017-06-17
feat(saucelab): integrate saucelab webdriver test.

### LINKED ISSUE #817: Microsoft EdgeHTML 15 Browser Tools hooking into zone.js
Originally Fixed in pull request #155 but now with EdgeHTML 15, the browser tools are hooking in at a different callback function.

https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L368

The latest version of Edge, `data.handler.toString()` does not return `[object FunctionWrapper]` but `function __BROWSERTOOLS_CONSOLE_SAFEFUNC() { [native code] }` instead. I made changes locally on my machine to zone.js/utils.ts to see if the browser events would go away in my performance tab, and sure enough they did. This seems to unnecessarily hook into zone.js, causing a lot of calls when testing performance.

Can anyone confirm this issue is back in EdgeHTML 15? There should be a better way than just check a string, because they have seemed to changed their internal function. Anyways, below is my local changes and works on my machine.

```
var testString = data.handler.toString();
validZoneHandler = data.handler && (testString === '[object FunctionWrapper]' || testString == 'function __BROWSERTOOLS_CONSOLE_SAFEFUNC() { [native code] }');
```

**Update**
Actually this code looks more correct as it handles the null check properly.
```
if(data.handler) {
  var testString = data.handler.toString();
  validZoneHandler = testString === '[object FunctionWrapper]' || testString == 'function __BROWSERTOOLS_CONSOLE_SAFEFUNC() { [native code] }';
}
```

JiaLiPassion:
I have removed the #817 fix and move it into #812. so this PR will only deal with `integration with webdriver in saucelabs`.

add `webdriver` test integration with `saucelabs` in `travis` to test `cross context` case.
And I will add `MSEdge 15` after `saucelabs` support it.

The webdriver test can also run in local.
the steps have been added to [DEVELOPER.md](url)

1. run locally

```
npm run webdriver-start
npm run webdriver-http
npm run webdriver-test
```

2. run locally with saucelabs

```
// export SAUCE_USERNAME and SAUCE_ACCESS_KEY
export SAUCE_USERNAME=XXXX
export SAUCE_ACCESS_KEY=XXX

sc -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY
npm run webdriver-http
npm run webdriver-sauce-test
```

@juliemr , @heathkit, could you please help me to check the `webdriver` code is ok or not?
I basically use the test code you wrote in other issues.

1. https://github.com/angular/zone.js/issues/234
2. https://github.com/angular/zone.js/issues/616

And add some `addEventListener to check Zone` test code.

## Updated

**I have confirmed the test code can detect errors such as in #616 `permission denied to access property invoke` by following steps.**

- test locally:
   I revert Firefox to 52, and I can reproduce the issue in #616 `permission denied to access property invoke` even with the newest `geckodriver` after I revert all `toString()` check. 

- test with saucelabs,  comment out the `toString` check  https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L364 and https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L416, I can reproduce the error in #616 `permission denied to access property invoke` in `saucelabs` webdriver test.

And currently it seems only `Firefox 52` has this issue, `Firefox 53/54`  is OK

Please review,

Thank you very much!


JiaLiPassion:
@mhevery , please review this one, I have moved the `Edge BROWSER_TOOLS` check into #812, so this PR is only for add webdriver test work in saucelabs.  Thank you!

mhevery:
@juliemr do you have a comment before I merge this in?

PR #840, opened 2017-07-14
reduce duplicate code for the zone aware callbacks

kgentes:
abstract the base functionality and pass in a capture variable instead of having a separate method for each value of capture.

### DIFF
--- a/lib/common/events.ts
+++ b/lib/common/events.ts
@@ -54,12 +54,12 @@ export function patchEventTarget(
     // invoke static task.invoke
     task.invoke(task, target, [event]);
   };
-
-  // global shared zoneAwareCallback to handle all event callback with capture = false
-  const globalZoneAwareCallback = function(event: Event) {
+  
+  // global shared zoneAwareCallback to handle all event callback with possible capture
+  const globalZoneAwareCallbackBase = function(event: Event, capture: string) {
     const target = this || _global;
 
-    const tasks = target[zoneSymbolEventNames[event.type][FALSE_STR]];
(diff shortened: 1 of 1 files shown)


lib/common/events.ts
```diff
@@ -54,12 +54,12 @@ export function patchEventTarget(
     // invoke static task.invoke
     task.invoke(task, target, [event]);
   };
-
-  // global shared zoneAwareCallback to handle all event callback with capture = false
-  const globalZoneAwareCallback = function(event: Event) {
+  
+  // global shared zoneAwareCallback to handle all event callback with possible capture
+  const globalZoneAwareCallbackBase = function(event: Event, capture: string) {
```
JiaLiPassion: Thanks for the PR, but if we changed the code like this, there will be an additional function call for all callbacks, it will have a little impact for performance.

So I copy the code on purpose, just want to have some better performance.

kgentes:
I signed cla

lib/common/events.ts
kgentes: I see. so performance is more essential than abstraction and maintenance in this case for zone.js?

JiaLiPassion: @kgentes , yes, the original version of `zone.js` has a much better abstraction and maintenance, no duplicate code, everything have a much better structure.

but the performance become the main issue, so in this case, performance is the essential priority. I think we have to do some trade off, for this `events.ts`, I think we should just focus on performance.

kgentes: ok, understood. Thanks for the explanation.

PR #843, opened 2017-07-16
feat(rxjs): fix #830, monkey patch rxjs to make rxjs run in correct zone

### LINKED ISSUE #830: RXjs and zone.js issue
Hello,

I'm developing an Outlook Add-in which essentially is an instance of a Safari browser, this points to an angular 4.2.5 app.

I'm heavily using NGRX Store and RX but I'm hitting issues when I run it inside Outlook rather than developing on Chrome.

This is all using the Angular CLI and JIT not AoT.

```
Angular version: 4.2.5
Zone.js version: 0.8.12
RX version: 5.4.1
Node version: 8.1
```

It seems #671 is close to what I'm seeing. But seemingly randomly. I essentially have:

Component:

```typescript
this.things$ = this.store.select((state) => state.things).distinctUntilChanged();
```

View:

```html
<div *ngFor="let thing of (things$ | async)">
    <p>Hello {{ thing.name }}!</p>
</div>
```

Sometimes it won't show the list at all, other times it'll do it.

If I swap back to arrays:

Component:

```typescript
this.things = [ ];

this.store
     .select((state) => state.things)
     .distinctUntilChanged()
     .map((things) => this.things = things)
     .subscribe();
```

View:

```html
<div *ngFor="let thing of things">
    <p>Hello {{ thing.name }}!</p>
</div>
```

This doesn't work at all!

I have to do this in the component:

```typescript
this.store
     .select((state) => state.things)
     .distinctUntilChanged()
     .map((things) => {
         this.ngZone.run(() => {
             this.things = things;
        });
     })
     .subscribe();
```

This works every time I reload the page or add-in.

Which to me hints at the RX is somehow escaping the zone on occasion?

JiaLiPassion:
copy logic from @mhevery 's PR on `rxjs` repository, https://github.com/ReactiveX/rxjs/pull/2266, try to monkey patch `rxjs` to make 

1. Observable construct
2. Subscription
3. Operator
4. unsubscribe 

run in correct zone.

now the test can pass in Firefox 54 and node.

@mhevery , please review the idea is workable or not. 

currently the risky part is 
https://github.com/angular/zone.js/pull/843/files#diff-7dba82bf33d9e5b20f59836ce1601b82R25
To patch `Observable` constructor. 

```javascript
const Observable = rxjs.Observable;

rxjs.Observable = function() {
  Observable.apply(this, arguments);
  this._zone = Zone.current;
  return this;
};
```

I am not sure patch like this is ok or not. I will continue to find some other way to make sure `new Observable()` will use this patched version.

If the idea is ok, I will add other test cases and make it work for other browsers.
Thank you!


lib/rxjs/rxjs.ts
```diff
@@ -0,0 +1,140 @@
...
+                                                         _subscribe.apply(this, arguments);
+          if (tearDownLogic && typeof tearDownLogic === 'function') {
+            const patchedTeadDownLogic = function() {
+              if (_zone && _zone !== Zone.current) {
+                return _zone.run(tearDownLogic, this, arguments);
+              } else {
+                return tearDownLogic.apply(this, arguments);
+              }
+            };
+            return patchedTeadDownLogic;
```
mleibman: Typo in 'patchedTeadDownLogic'.

JiaLiPassion: @mleibman, thanks!

lib/rxjs/rxjs.ts
```diff
@@ -0,0 +1,140 @@
...
+  let rxjs;
+
+  const subscribeSource = 'rxjs.subscribe';
+  const nextSource = 'rxjs.Subscriber.next';
+  const errorSource = 'rxjs.Subscriber.error';
+  const completeSource = 'rxjs.Subscriber.complete';
+  const unsubscribeSource = 'rxjs.Subscriber.unsubscribe';
+
+  try {
+    rxjs = require('rxjs');
```
mhevery: `require` statement will not work in browser.

So far the `.js` files have been stand alone code which can execute in global context, this means that we can't rely on ES6 `import` to get a hold of `Rx`. 

There is also the matter that we have to run the code at a specific time. We have to load `Zone.js` first than `Rx` and only than we can do `Rx` patching but before the other code gets it. 

I think the solution here is to package this file as `umd` that way it can be used in both cases.
It would mean that at some point one would have to import it as: `zone.js/rxjs-patch`

lib/rxjs/rxjs.ts
```diff
@@ -0,0 +1,140 @@
...
+  Observable.prototype.subscribe = function() {
+    const _zone = this._zone;
+    const currentZone = Zone.current;
+
+    if (this._subscribe && typeof this._subscribe === 'function') {
+      this._subscribe._zone = this._zone;
+      const _subscribe = this._subscribe;
+      if (_zone) {
+        this._subscribe = function() {
+          const subscriber = arguments.length > 0 ? arguments[0] : undefined;
```
mhevery: referring to `arguments` is slower. Can we find another way to do this?

lib/rxjs/rxjs.ts
```diff
@@ -0,0 +1,140 @@
...
+          }
+          return tearDownLogic;
+        };
+      }
+    }
+
+    if (this.operator && _zone && _zone !== currentZone) {
+      const call = this.operator.call;
+      this.operator.call = function() {
+        const subscriber = arguments.length > 0 ? arguments[0] : undefined;
```
mhevery: don't refer to `arguments`

gulpfile.js
```diff
@@ -199,6 +199,14 @@ gulp.task('build/sync-test.js', ['compile-esm'], function(cb) {
   return generateScript('./lib/zone-spec/sync-test.ts', 'sync-test.js', false, cb);
 });
 
+gulp.task('build/rxjs.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/rxjs/rxjs.ts', 'zone-rxjs.js', false, cb);
```
JiaLiPassion: Got it, I will change it.

lib/rxjs/rxjs.ts
JiaLiPassion: yeah, I will use an array to get the object.

lib/rxjs/rxjs.ts
JiaLiPassion: yeah, I will use an array to get the object.

lib/rxjs/rxjs.ts
JiaLiPassion: got it, I will try to organize the file into `umd` way.

JiaLiPassion:
@mhevery , I have modified the code based on your comment, and the travis build will be passed after #844 was merged, please review.

lib/rxjs/rxjs.ts
```diff
@@ -0,0 +1,244 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+declare let define: any;
+(function(root: any, factory: (Rx: any) => any) {
```
mhevery: Instead of doing it this way we should use rollup umd function to build the UMD.

This should be replaced with simple `import {RX} from 'rxjs'`

See: https://rollupjs.org/repl?version=0.45.2&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QlJ4JTdEJTIwZnJvbSUyMCdyeGpzJTJGUlgnJTVDbiU1Q24lMkYlMkYlMjBkbyUyMHNvbWV0aGluZyUyMCUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJ1bWQlMjIlMkMlMjJtb2R1bGVOYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTJDJTIybW9kdWxlSWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==

lib/rxjs/rxjs.ts
```diff
@@ -0,0 +1,244 @@
...
+    // AMD
+    define(['rxjs'], factory);
+  } else {
+    factory(root.Rx);
+  }
+}(typeof window !== 'undefined' && window || typeof self !== 'undefined' && self || global,
+  (Rx: any) => {
+    'use strict';
+    Zone.__load_patch('rxjs', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
+      const subscribeSource = 'rxjs.subscribe';
```
mhevery: Out of these `const`s only `nextSource` seems to be used.

test/rxjs/rxjs.spec.ts
```diff
@@ -0,0 +1,242 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+let rxjs: any;
```
JiaLiPassion: Ok, I will try it.

lib/rxjs/rxjs.ts
JiaLiPassion: Got it, I will change it.

lib/rxjs/rxjs.ts
JiaLiPassion: yeah, I used the wrong const, I will modify this one.

JiaLiPassion:
@misko, I just updated the `import/UMD` part, please review!

misko:
@mhevery

PR #854, opened 2017-07-24
fix(spec): do not count requestAnimationFrame as a pending timer

vikerman:
We added a new fature to FakeAsyncTestSpec to treat a
requestAnimationFrame as a timer with 16ms. However this breaks existing
Angular project tests because the RAF is reported as a pending timer at
the end of the test.

This change makes sure that existing Angular tests are not broken by not
counting RAF-s as pending timers at the end of a test.

### DIFF
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -14,6 +14,7 @@
     args: any[];
     delay: number;
     isPeriodic: boolean;
+    isRequestAnimationFrame: boolean;
   }
 
   interface MicroTaskScheduledFunction {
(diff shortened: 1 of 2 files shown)


lib/zone-spec/fake-async-test.ts
```diff
@@ -35,7 +36,7 @@
 
     scheduleFunction(
         cb: Function, delay: number, args: any[] = [], isPeriodic: boolean = false,
-        id: number = -1): number {
+        isRequestAnimationFrame: boolean = false, id: number = -1): number {
```
JiaLiPassion: @vikerman , should we also add an option flag in `FakeAsyncTestSpec` to control consider `RAF` as timer or not?

```javascript
constructor(namePrefix: string, isRequestAnimationFrame: boolean = false) {
      this.name = 'fakeAsyncTestZone for ' + namePrefix;
      this.isRequestAnimationFrame = isRequestAnimationFrame;
    }
```

vikerman: Done

PR #855, opened 2017-07-25
fix(console): console.log in nodejs should run in root Zone

JiaLiPassion:
fix https://github.com/angular/angular/issues/17180

in nodejs, `console.log` will use `process.nextTick` and schedule a microTask which will cause 
infinite loop, so we need to make sure it run in `Zone.root`.


vikerman:
Thanks! LGTM

PR #877, opened 2017-08-16
feat(onProperty): fix #875, can disable patch specified onProperties

### LINKED ISSUE #875: [API] __Zone_disable_RequestAnimationFrame
We use  webgl with zone.js.
```requestAnimationFrame```  monkey patching makes perf issue from zone.js.
We need api such as ```__Zone_disable_RequestAnimationFrame = true;```.

I found some flag on/off api from the modules page here.
Someone has any idea at all? ;)

### Modules
https://github.com/angular/zone.js/blob/b3a76d3f8d374be8edf009f7a077e51eefec6a65/MODULE.md

### PR DESCRIPTION
**Pull request by JiaLiPassion**
fix #875.

can disable patch specified onProperties,for example,
if you want to disable `window.onmessage` and `HTMLElement.prototype.onclick` from zone.js patching,
you can do like this.

```
 <script>
    __Zone_ignore_on_properties = [
      {
        target: window,
        ignoreProperties: ['message']
      }, {
        target: HTMLElement.prototype,
        ignoreProperties: ['click']
      }
    ];
  </script>
  <script src="../dist/zone.js"></script>
```

### DIFF
--- a/lib/browser/property-descriptor.ts
+++ b/lib/browser/property-descriptor.ts
@@ -230,7 +230,7 @@ export const eventNames = globalEventHandlersEventNames.concat(
     webglEventNames, formEventNames, detailEventNames, documentEventNames, windowEventNames,
     htmlElementEventNames, ieElementEventNames);
 
-interface IgnoreProperty {
+export interface IgnoreProperty {
   target: any;
   ignoreProperties: string[];
 }
--- a/lib/browser/property-descriptor.ts
+++ b/lib/browser/property-descriptor.ts
@@ -230,6 +230,32 @@ export const eventNames = globalEventHandlersEventNames.concat(
     webglEventNames, formEventNames, detailEventNames, documentEventNames, windowEventNames,
     htmlElementEventNames, ieElementEventNames);
 
+interface IgnoreProperty {
+  target: any;
+  ignoreProperties: string[];
+}
+
+function filterProperties(
+    target: any, onProperties: string[], ignoreProperties: IgnoreProperty[]): string[] {
+  if (!ignoreProperties) {
+    return onProperties;
+  }
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/browser/property-descriptor.ts [hunk 1]
```diff
@@ -230,53 +230,86 @@ export const eventNames = globalEventHandlersEventNames.concat(
...
+}
+
 export function propertyDescriptorPatch(api: _ZonePrivate, _global: any) {
   if (isNode && !isMix) {
     return;
   }
 
   const supportsWebSocket = typeof WebSocket !== 'undefined';
   if (canPatchViaPropertyDescriptor()) {
+    const ignoreProperties: IgnoreProperty[] = _global.__Zone_ignore_on_properties;
```
mhevery: could we turn this into our own module so that it is consistent with the rest of the system?

JiaLiPassion: @mhevery , currently `propertyDescriptorPatch` is only be called from here, it is already in our module system.

```javascript
Zone.__load_patch('on_property', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
  propertyDescriptorPatch(api, global);
  propertyPatch();
  registerElementPatch(global);
});
```

so do you mean we should make the `ignoreProperties` as module like this?

```javascript
Zone.__load_patch('ignoreOnProperties', (global, Zone, api) => {
  Zone[symbol('ignoreProperties')] = [
      {
        target: window,
        ignoreProperties: ['message']
      }, {
        target: HTMLElement.prototype,
        ignoreProperties: ['click']
      }
    ];
});

```

is that right?

mhevery: So what is the point of having this check?

PR #878, opened 2017-08-16
fix(task): fix closure compatibility issue with ZoneDelegate._updateT…

qiyigg:
…askCount

onHasTask wasn't getting triggerred for the Angular Zone for setTimeout
when the application was closure compiled. This was causing errors with
Protractor not being able to wait for a closure compiled application using setTimeout.


qiyigg:
Joined Google github org

PR #884, opened 2017-08-22
fix(rxjs): asap should runGuarded to let error inZone

JiaLiPassion:
fix https://github.com/angular/angular/issues/14316.

```javascript
    const constructorZone: Zone = Zone.root.fork({
    name: 'Constructor Zone',
    onHandleError: (delegate: ZoneDelegate, currentZone: Zone, targetZone: Zone, error: Error) => {
      log.push('error' + error.message);
      const result = delegate.handleError(targetZone, error);
      errorCallback && errorCallback();
      return false;
    }
  });
     let observable: any;
       constructorZone.run(() => {
         observable = Rx.Observable.of(1, 2, 3).observeOn(Rx.Scheduler.asap);
       });

       errorCallback = () => {
         expect(log).toEqual(['erroroops']);
         setTimeout(done, 0);
       };

       Zone.root.run(() => {
         observable
             .map((value: number) => {
               if (value === 3) {
                 throw new Error('oops');
               }
               return value;
             })
             .subscribe((value: number) => {});
       });
```

the case like this will reproduce the error, error will not in constructor zone if error throws with asap.
we should use `zone.runGuarded` in patched Immediate.

### DIFF
--- a/NON-STANDARD-APIS.md
+++ b/NON-STANDARD-APIS.md
@@ -103,3 +103,61 @@ for example, if you want to add MediaQuery patch, you should do like this.
   <script src="path/zone.js"></script> 
   <script src="path/webapis-media-query.js"></script> 
 ```  
+
+* rxjs
+
+`zone.js` also provide a `rxjs` patch to make sure rxjs Observable/Subscription/Operator run in correct zone.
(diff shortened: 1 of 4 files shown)


mhevery:
@JiaLiPassion  Can you look into why the build is red?

JiaLiPassion:
@mhevery , sure , I will check it.

PR #887, opened 2017-08-25
fix(patch): fix #886, #885, patchOnProperties getter should return original listener.

### LINKED ISSUE #885: ZoneJs 8.17 break call of event listeners
var wrapFn = function (event) {
    var eventNameSymbol = zoneSymbolEventNames[event.type];
    if (!eventNameSymbol) {
        eventNameSymbol = zoneSymbolEventNames[event.type] = zoneSymbol('ON_PROPERTY' + event.type);
    }
If some calls function without parameter, if event is undefined zone crashes
In my case window.onbeforeunload is called from parent frame but zone wrapFn does not check for event parameter and trys to get it from map, and since event is undefined it crashes

### LINKED ISSUE #886: Zone.js 0.8.17 and Angular-cli / rangeerror maximum call stack size exceeded
I thought that might be helpful If I open an issue in "zone.js" repo instead of "angular-cli". 

Please refer to this comment https://github.com/angular/angular-cli/issues/7502#issuecomment-324852946,

JiaLiPassion:
fix #886. 
fix #885.
fix https://github.com/angular/angular-cli/issues/7502.
patchOnProperties getter should return original listener.

the issue can be described as following code.

```javascript
         req.open('get', '/', true);
         const listener = req.onreadystatechange = function() {
           if (req.readyState === 4) {
             done();
           }
         };
         expect(req.onreadystatechange).toBe(listener);
         req.onreadystatechange = function() {
           return listener.apply(this, arguments);
         };
         req.send();
```

when try to get `onProperty listener` with such as `const listener = req.onreadystatechange`, should not return `wrapFn` but the real listener.

### DIFF
--- a/lib/common/utils.ts
+++ b/lib/common/utils.ts
@@ -170,8 +170,9 @@ export function patchProperty(obj: any, prop: string, prototype?: any) {
     if (!target) {
       return null;
     }
-    if (target[eventNameSymbol]) {
-      return wrapFn;
+    const listener = target[eventNameSymbol];
+    if (listener) {
+      return listener;
     } else if (originalDescGet) {
       // result will be null when use inline event attribute,
       // such as <button onclick="func();">OK</button>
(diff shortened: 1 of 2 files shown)


lingthing5:
Any ETA on this ? We currently have to force 0.8.16 in our project (17 is breaking)

PR #893, opened 2017-09-05
feat(spec): log URL in error when attempting XHR from FakeAsyncTestZone

vikerman:
Save the URL from xhr.open and log it when throwing an error from FakeAsyncTestZone when the XHR send it attempted. This would make it easier to debug such errors from a fakeAsync test.

### DIFF
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -339,7 +339,9 @@
                   this._setInterval(task.invoke, task.data['delay'], (task.data as any)['args']);
               break;
             case 'XMLHttpRequest.send':
-              throw new Error('Cannot make XHRs from within a fake async test.');
+              throw new Error(
+                  'Cannot make XHRs from within a fake async test. Request URL: ' +
+                  (task.data as any)['url']);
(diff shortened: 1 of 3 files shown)


lib/zone-spec/fake-async-test.ts
```diff
@@ -339,7 +339,9 @@
                   this._setInterval(task.invoke, task.data['delay'], (task.data as any)['args']);
               break;
             case 'XMLHttpRequest.send':
-              throw new Error('Cannot make XHRs from within a fake async test.');
+              throw new Error(
+                  'Cannot make XHRs from within a fake async test. Request URL: ' +
+                  (task.data as any)['url']);
```
JiaLiPassion: @vikerman , if we get the url from `task.data`, maybe we can get the url by directly calling 'task.data.target.responseURL`?

vikerman: From my testing responseURL is an empty string before the actual request is sent.

JiaLiPassion: @vikerman , yes, you are right!

PR #894, opened 2017-09-09
fix(timer): fix #314, setTimeout/interval should return original timerId

### LINKED ISSUE #314: zone.js breaks GWT code
From 0.6 zone.js' wrapper for setTimeout and setInterval returns a task object rather than the integer value. This breaks GWT timer related classes because GWT tries to convert the object to integer to match the java function signature.
Tested in GWT 2.7.

JiaLiPassion:
fix #314 

in browser and nodejs, `setTimeout`, `setInterval` should return original `timerId` (number or Timer object) instead of `ZoneTask`, because it will break some logic if other libraries expect the standard data types.

### DIFF
--- a/lib/common/timers.ts
+++ b/lib/common/timers.ts
@@ -10,7 +10,9 @@
  * @suppress {missingRequire}
  */
 
-import {patchMethod} from './utils';
+import {patchMethod, zoneSymbol} from './utils';
+
+const taskSymbol = zoneSymbol('zoneTask');
 
 interface TimerOptions extends TaskData {
   handleId: number;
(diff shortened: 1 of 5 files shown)


JiaLiPassion:
don't know why `saucelabs` can't launch `safari 10` on `macOS 10.12`, I will ask for solution in `saucelabs` site.

tkohr:
We call setInterval() in our angular 4 application. Until zone.js 0.8.17 this returned us a ZoneTask with a runCount attribute, which we used in further code. Do I understand it right that this is not possible anymore following this change (as the timerId is now returned from 0.8.18)?

JiaLiPassion:
@tkohr , yes, from 0.8.18, `setTimeout` will return original timerId instead of `ZoneTask`. Because we think we should not break the standard APIs, could you describe your use case why you need the `zoneTask.runCount`?

tkohr:
@JiaLiPassion , our use case is the following: We call REST API A every 20 seconds with `setInterval` and API B on every 20th call of A using `zoneTask.runCount`. This counter was convient to us but can be replaced by a counter we generate, of course. If `setInterval` will continue to return the id in future versions of zone.js we will adapt our code accordingly. Will this be the case?

JiaLiPassion:
@tkohr , yeah, I see, I believe the `setInterval` will return `id` just like the standard API does, so please change your application code.

Issue #891, opened 2017-09-01
es6-promise overwrites zone.js


josephliccini:
I am not sure exactly which library is exactly responsible for the issue, but if one looks at this line:

https://github.com/stefanpenner/es6-promise/blob/a5838bd94a76e31ad80ac1685bcd0dbc3d8ce963/dist/es6-promise.js#L1133-L1144

One can see that they are expecting
`Object.prototype.toString.call(window.Promise.resolve())` to be `[object Promise]`, but `zone.js` gives `[object Object]`

So, I am wondering, should `es6-promise` (coming from another widget hosted on our page served over their CDN) look for `[object Object]`, or should zone.js have `[object Promise]` for this `toString()`

josephliccini:
This seems to work, though not clean or reliable:
```ts
var oldToString = Object.prototype.toString;
Object.prototype.toString = function () {
    if (this instanceof Promise) {
        return '[object Promise]';
    }
    return oldToString.call(this);
};
```

Maybe this could be added to zone.js so other polyfills won't overwrite?

josephliccini:
Or perhaps Zone.js could make `window.Promise` a `setter` function, and anytime it is set, Zone.js wraps with ZoneAwarePromise? So then if any library were to set `window.Promise`, it could reject it, wrap it, or at least warn?  Just brainstorming here

JiaLiPassion:
@josephliccini , it seems duplicate with #499, I will check the behavior later.

josephliccini:
This one seems fixed! Thanks!

There is one issue I am seeing, but it's not affecting me but may be good for you to know.

I did a test to see what would happen with `core-js` Promise.

My code has this now: 
```ts
import 'zone.js';
import 'core-js/es6/promise';
```

And before bootstrap I get this error:
```
Uncaught TypeError: NativePromise.resolve is not a function
    at Object.setNativePromise (zone.js:634)
    at desc.set (zone.js:1028)
    at webpackJsonp.../../../../node_modules/core-js/modules/_redefine.js.module.exports (_redefine.js:19)
    at $export (_export.js:27)
    at Object.../../../../node_modules/core-js/modules/es6.promise.js (es6.promise.js:230)
    at __webpack_require__ (?mockauth=2:61)
    at Object.../../../../node_modules/core-js/es6/promise.js (promise.js:4)
    at __webpack_require__ (?mockauth=2:61)
    at Object../vendor.ts (zone.js:3026)
[...]

Issue #783, opened 2017-05-16
core-js es6.promise conflicts with ZoneAwarePromise


jahtalab:
Importing `core-js/shim` after zone.js would cause certain properties of `ZoneAwarePromise` to be overwritten. Eventually it will end up in the wrong zone.

Example:

```
import 'zone.js';
import 'core-js/shim';

window.Zone.current.fork({name:'new zone'}).run(function (){
    Promise.resolve().then(resp => {
      console.log(window.Zone.current.name) // it's <root> zone!
    })
})
```

It seems like importing `core-js/shim` first would resolve the issue. But I'm not sure about side effects of importing `zone.js` after `core-js/shim`.

Could this issue be fixed regardless of importing order?

jahtalab:
More investigation:

Turns out `es6.promise` module tries to see if it should use native api or not but gets it "wrong" when zone.js is present.

This piece of code:
```
var USE_NATIVE = !!function(){
  try {
    // correct subclassing with @@species support
    var promise     = $Promise.resolve(1)
      , FakePromise = (promise.constructor = {})[require('./_wks')('species')] = function(exec){ exec(empty, empty); };
    // unhandled rejections tracking support, NodeJS Promise without it fails @@species test
    return (isNode || typeof PromiseRejectionEvent == 'function') && promise.then(empty) instanceof FakePromise;
  } catch(e){ /* empty */ }
}();
```
Reference: https://github.com/zloirock/core-js/blob/master/modules/es6.promise.js#L22

`USE_NATIVE` is false with zone.js in the page and that's why it patches `Promise` properties like `resolve`.

I think zone.js should behave in a way that it can not be distinguished from native api.

JiaLiPassion:
@jahtalab , all modules which provide `promise` polyfill or implementation should be loaded before `zone.js`, is there any reason that you want it be loaded after `zone.js`?

jahtalab:
@JiaLiPassion , Thanks for your response.
Generally I think it's much simpler to explain to users that "`zone.js` should be loaded before any other library to avoid unpredictable results". Over time the number of libraries grow and so does the number of exceptions to that general rule (there are already three exceptions in the readme). This makes including libraries complicated.

Furthermore `core-js` provides a lot of different polyfills and they are all usually loaded together. while the promise polyfill is advised to be loaded before `zone.js` the rest of them could potentially have conflicts with `zone.js` if they are loaded before `zone.js`.

[...]

PR #899, opened 2017-09-12
fix(promise): can set native promise after loading zone.js

### LINKED ISSUE #898: Error because of a conflict with Google Tag Manager code
I am getting the error
`zone.js has detected that zoneawarepromise `(window|global).promise` has been overwritten`
in my angular2 project. That happens from time to time on page reload and probably depends on connection speed. As I figured out that is caused by a conflict with gtm code. Most likely  gtm script is asynchronously loading some other scripts (hot jar in my case) which can overwrite `window.promise`. And it does not matter where exactly these scripts were placed

The workaround which is working for me 
```
Zone.assertZonePatched = function () {
            if (global.Promise !== ZoneAwarePromise) {
                global.Promise = ZoneAwarePromise;
            }
 };
```
but how safe is it?

I would be grateful for explanation and any smarter workarounds. Changing gtm code is not acceptable for me. 

`"zone.js": "0.7.4"` (also tried v0.8.17)

JiaLiPassion:
fix #898, #891, #783 

currently after loading `zone.js`, if we load some libraries which will override `global Promise`, `Zone.assertZonePatched` will throw error, and `angular` will not started. 
Sometimes we use 3rd party libraries and hard to control the load orders. So we need a better way to handle this process.

in this PR.

1. patch `global.Promise` with a setter by using Object.defineProperty , so after loading `zone.js`, if a new Promise want to override `global.Promise`, it will not update the `global.Promise`, but update the underlying `NativePromise`, so the whole process will be:

   - before loading `zone.js`, `NativePromise` is Browser built in Promise.
   - load `zone.js`, global.Promise is `ZoneAwarePromise`, global[symbolPromise] is `NativePromise`.
   - after loading `zone.js`, set other Promise, such as `es6-promise`, global.Promise is still `ZoneAwarePromise`, and global[symbolPromise] will be updated to `es6-promise`.

2. Zone.assertZonePatched will only throw error when user set this before loading `zone.js`

```javascript
__Zone_disable_Promise = true;
```

### DIFF
--- a/karma-base.conf.js
+++ b/karma-base.conf.js
@@ -16,6 +16,7 @@ module.exports = function (config) {
       {pattern: 'node_modules/rxjs/**/**/*.js', included: false, watched: false },
       {pattern: 'node_modules/rxjs/**/**/*.js.map', included: false, watched: false },
       {pattern: 'node_modules/rxjs/**/*.js', included: false, watched: false },
+      {pattern: 'node_modules/es6-promise/**/*.js', included: false, watched: false },
       {pattern: 'node_modules/rxjs/**/*.js.map', included: false, watched: false },
       {pattern: 'test/assets/**/*.*', watched: true, served: true, included: false},
       {pattern: 'build/**/*.js.map', watched: true, served: true, included: false},
--- a/karma-dist-sauce-jasmine.conf.js
+++ b/karma-dist-sauce-jasmine.conf.js
@@ -8,5 +8,5 @@
 
 module.exports = function (config) {
   require('./karma-dist-jasmine.conf.js')(config);
-  require('./sauce.conf')(config);
+  require('./sauce.conf')(config, ['SL_IOS9']);
 };
--- a/lib/common/promise.ts
+++ b/lib/common/promise.ts
@@ -330,6 +330,44 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
   ZoneAwarePromise['all'] = ZoneAwarePromise.all;
 
   const NativePromise = global[symbolPromise] = global['Promise'];
+  const ZONE_AWARE_PROMISE = Zone.__symbol__('ZoneAwarePromise');
+
+  let desc = Object.getOwnPropertyDescriptor(global, 'Promise');
+  if (!desc || desc.configurable) {
+    desc && delete desc.writable;
+    desc && delete desc.value;
+    if (!desc) {
+      desc = {configurable: true, enumerable: true};
+    }
+    desc.get = function() {
+      // if we already set ZoneAwarePromise, use patched one
+      // otherwise return native one.
+      return global[ZONE_AWARE_PROMISE] ? global[ZONE_AWARE_PROMISE] : global[symbolPromise];
+    };
+    desc.set = function(NewNativePromise) {
+      if (NewNativePromise === ZoneAwarePromise) {
+        // if the NewNativePromise is ZoneAwarePromise
+        // save to global
+        global[ZONE_AWARE_PROMISE] = NewNativePromise;
+      } else {
+        // if the NewNativePromise is not ZoneAwarePromise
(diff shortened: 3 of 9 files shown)


mhevery:
This looks like a good idea. Let me know when it is ready.

JiaLiPassion:
@mhevery , I have resolved the conflict and rebased, please review.

josephliccini:
This is great and will alleviate a huge pain for us.  We have to load a variety of widgets from CDN that all serve all sorts of Promise polyfills.  Thank you for the fix!  Looking forward to the next release of zone.js with this fixed

Mystic-Ervo:
Looks like the error has returned, I can't execute my angular universal server-side app, node throws me that error when I try it.
This error has been fixed in multiple times and always return, looks like a nightmare!

```console
(node:15488) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
/var/www/html/proyectoclase/server.js:3024
 */!function(e){var t=e.performance;functionn(e){t&&t.mark&&t.mark(e)}function r(e,n){t&&t.measure&&t.measure(e,n)}if(n("Zone"),e.Zone)throw new Error("Zone already loaded.");var i,o=function(){function t(e,t){this._properties=null,this._parent=e,this._name=t?t.name||"unnamed":"<root>",this._properties=t&&t.properties||{},this._zoneDelegate=new s(this,this._parent&&this._parent._zoneDelegate,t)}return t.assertZonePatched=function(){if(e.Promise!==k.ZoneAwarePromise)throw new Error("Zone.js has detected that ZoneAwarePromise `(window|global).Promise` has been overwritten.\nMost likely cause is that a Promise polyfill has been loaded after Zone.js (Polyfilling Promise api is not necessary when zone.js is loaded. If you must load one, do so before loading zone.js.)")},Object.defineProperty(t,"root",{get:function(){for(var e=t.current;e.parent;)e=e.parent;return e},enumerable:!0,configurable:!0}),Object.defineProperty(t,"current",{get:function(){return O.zone},enumera

Error: listen EACCES http://https://wikos-inc.com/proyectoclase/
        at Server.setupListenHandle [as _listen2] (net.js:1313:19)
        at listenInCluster (net.js:1378:12)
        at Server.listen (net.js:1477:5)
        at Function._.listen (/var/www/html/proyectoclase/server.js:2900:4545)
        at Object.<anonymous> (/var/www/html/proyectoclase/server.js:3077:2865)
        at n (/var/www/html/proyectoclase/server.js:1:172)
        at /var/www/html/proyectoclase/server.js:1:979
        at Object.<anonymous> (/var/www/html/proyectoclase/server.js:1:990)
        at Module._compile (internal/modules/cjs/loader.js:702:30)
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
Emitted 'error' event at:
        at emitErrorNT (net.js:1357:8)
        at e.invokeTask (/var/www/html/proyectoclase/server.js:3024:7685)
        at t.runTask (/var/www/html/proyectoclase/server.js:3024:2865)
        at t.invokeTask (/var/www/html/proyectoclase/server.js:3024:8799)
        at o.useG.invoke (/var/www/html/proyectoclase/server.js:3024:8690)
        at t.args.(anonymous function) (/var/www/html/proyectoclase/server.js:3077:831)
        at process._tickCallback (internal/process/next_tick.js:63:19)
        at Function.Module.runMain (internal/modules/cjs/loader.js:746:11)
        at startup (internal/bootstrap/node.js:238:19)
        at bootstrapNodeJSCore (internal/bootstrap/node.js:572:3)

JiaLiPassion:
@Mystic-Ervo , some how the `zone.js` was imported twice, this has been fixed here , #1093, please wait for the next release.

Mystic-Ervo:
@JiaLiPassion oh, nice! ^^
Thanks for answering!

PR #901, opened 2017-09-13
fix: for Samusung TV and SourceBuffer

mhevery:


lib/common/events.ts
```diff
@@ -91,8 +91,7 @@ export function patchEventTarget(
 
   // global shared zoneAwareCallback to handle all event callback with capture = false
   const globalZoneAwareCallback = function(event: Event) {
-    const target = this || _global;
-
+    const target: any = event.target;
```
mhevery: Hi @JiaLiPassion Why do we even do `|| _global`? Is this not sufficient?

lib/common/events.ts
```diff
@@ -91,8 +91,9 @@ export function patchEventTarget(
 
   // global shared zoneAwareCallback to handle all event callback with capture = false
   const globalZoneAwareCallback = function(event: Event) {
-    const target = this || _global;
-
+    // event.target is needed for Samusung TV and SourceBuffer
+    // || global is needed https://github.com/angular/zone.js/issues/190
+    const target: any = this || event && event.target || _global;
```
JiaLiPassion: @mhevery , should we add 
```javascript
if (!event) {
  return;
} 
```

because in the next line, we use `event.type` directly.

JiaLiPassion:
@mhevery , we also need to add the same logic on `onProperty wrapFn`, https://github.com/angular/zone.js/blob/master/lib/common/utils.ts#L92
currently it also use `this` directly.

And I will try to figure out how to write a test case about this one.

mhevery:
Perfect! Closing in favor of https://github.com/angular/zone.js/pull/902

PR #902, opened 2017-09-15
fix(event): add test cases for EventTarget patch for Samsung tv

JiaLiPassion:
1. handle `this` maybe null for `onProperty` when user use such as `sourceBuffer.onupdate`.
2. add test cases.


JiaLiPassion:
test failed in IE 9, I will try to find out why.

PR #915, opened 2017-09-27
fix(patch): Worker should patch onProperties

JiaLiPassion:
fix https://github.com/angular/angular/issues/19334
`Worker` onProperties (`onerror`, `onmessage`) should also be patched.


bvdmitri:
Safari 11.0.3 Worker has no prototype

```
> window['Worker']
< function WrappedWorker() {
    [native code]
}
> window['Worker'].prototype
< undefined
```

TypeError will be raised when getOwnPropertyDescriptor will be called with 'undefined' prototype.

PR #916, opened 2017-09-28
fix(promise): fix #891, sometimes NativePromise will not ready, and reduce zone.js size

### LINKED ISSUE #891: es6-promise overwrites zone.js

JiaLiPassion:
fix #891 
1. sometimes NativePromise will not be fully ready when override global.Promise, we should add check about it.
2. now by default, typescript 2.3 will compile `iterable` and `spread operator` with new polyfill code and will make `dist/zone.js` bigger, and we don't really have to use `iterable` and `spread`, so we use simple loop to reduce the `zone.js` size.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -1320,7 +1320,12 @@ const Zone: ZoneType = (function(global: any) {
     patchOnProperties: noop,
     patchMethod: () => noop,
     setNativePromise: (NativePromise: any) => {
-      nativeMicroTaskQueuePromise = NativePromise.resolve(0);
+      // sometimes NativePromise.resolve static function
+      // is not ready yet, (such as core-js/es6.promise)
+      // so we need to check here.
+      if (NativePromise && typeof NativePromise.resolve === FUNCTION) {
+        nativeMicroTaskQueuePromise = NativePromise.resolve(0);
+      }
     },
   };
   let _currentZoneFrame: _ZoneFrame = {parent: null, zone: new Zone(null, null)};
--- a/lib/common/events.ts
+++ b/lib/common/events.ts
@@ -531,7 +531,7 @@ export function patchEventTarget(
           const captureTasks = target[symbolCaptureEventName];
 
           if (tasks) {
-            const removeTasks = [...tasks];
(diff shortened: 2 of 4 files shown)


josephliccini:
@JiaLiPassion just so I understand this a little better, the `core-js/es6/promise` will:
> Lazy load it's prototype

 can you elaborate a little more? Thanks!

lib/zone.ts
```diff
@@ -1320,7 +1320,11 @@ const Zone: ZoneType = (function(global: any) {
     patchOnProperties: noop,
     patchMethod: () => noop,
     setNativePromise: (NativePromise: any) => {
-      nativeMicroTaskQueuePromise = NativePromise.resolve(0);
+      // sometimes NativePromise lazy load it's prototype
```
josephliccini: it's => its (typo) :)

JiaLiPassion:
@josephliccini , sure, in `core-js es6.promise`, the initialization of `ES6 Promise` is 

1. the `Promise constructor` was created, https://github.com/zloirock/core-js/blob/master/modules/es6.promise.js#L166
2. override `global.Promise`
https://github.com/zloirock/core-js/blob/master/modules/es6.promise.js#L216
3. prepare `Promise.resove` static method.
https://github.com/zloirock/core-js/blob/master/modules/es6.promise.js#L222

So this will cause the error.

lib/zone.ts
JiaLiPassion: thanks, I will update the comment.

josephliccini:
Got it.  Thank you!!

PR #931, opened 2017-10-18
fix(jasmine): store original jasmine method

vikerman:
This lets us skip the ProxyZone if we really don't need it in our tests.

### DIFF
--- a/lib/jasmine/jasmine.ts
+++ b/lib/jasmine/jasmine.ts
@@ -56,6 +56,7 @@
   });
   ['it', 'xit', 'fit'].forEach((methodName) => {
     let originalJasmineFn: Function = jasmineEnv[methodName];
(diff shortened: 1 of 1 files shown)


lib/jasmine/jasmine.ts
```diff
@@ -56,6 +56,7 @@
   });
   ['it', 'xit', 'fit'].forEach((methodName) => {
     let originalJasmineFn: Function = jasmineEnv[methodName];
+    jasmineEnv['original_' + methodName] = originalJasmineFn;
```
JiaLiPassion: @vikerman , currently Zone store the key of `original` method as `Zone.__symbol__(methodName)`, so should we keep it that way and change this line as below?
```
jasmineEnv[Zone.__symbol__(methodName)] = originalJasmineFn;
```

PR #935, opened 2017-10-21
fix(timer): setInterval should not auto cancel after callback invoked, fix rxjs version to pass CI

### LINKED ISSUE #934: zone.js 0.8.18 setInterval/clearInterval
I've been working on a Angular app and I tracked a bug I've been seeing back to zone.js ... if my application runs setInterval/clearInterval, the main Angular zone never reports itself to be stable afterward. This is causing problems when I run Protractor tests.

I've built a small Angular 4 application that demonstrates the problem.

https://github.com/bredmold/zone-bug

What seems to be happening when I step through the JS in the browser is that the timer handle is being thrown out after the browser's native setInterval method returns (presuming I'm interpreting the code correctly).

https://github.com/angular/zone.js/blob/aec4bd41d17b2332d8cdcf16e1ea57010d14994f/lib/common/timers.ts#L45

When clearInterval is called, zone.js has no way to address the timer handle and locate the task being cancelled. Thus, the task counts are never updated, hence the zone cannot report itself to be stable... and my tests time out.

The obvious workaround would be to not call setInterval/clearInterval in my application, but they're in a third-party library. I hope there's enough information here to act on.

JiaLiPassion:
fix #934.
`setInterval` will auto clear timerId  after  first time invoked.
the issue was imported several versions ago, but it doesn't become a problem because we 
return `ZoneTask` from `setInterval` before `zone.js 0.8.17`, from `zone.js 0.8.18` , `setInterval` will return the original `timerId`, so when we call `clearInterval` the `timerId` have been cleared, which will cause zone.js `onHasTask` not be called, the `setInterval` task was always there.
And will cause `ngZone` not stable.

Also fix `rxjs` version to `5.4.2` to avoid `karam` error, I will continue to figure out how to resolve `karam` test error with `rxjs 5.5.0`. 
The issue is here, and it seems it will be fixed in `rxjs 5.5.1`
https://github.com/ReactiveX/rxjs/issues/2971

### DIFF
--- a/lib/common/timers.ts
+++ b/lib/common/timers.ts
@@ -39,6 +39,12 @@ export function patchTimer(window: any, setName: string, cancelName: string, nam
       try {
         task.invoke.apply(this, arguments);
       } finally {
+        if (task.data && task.data.isPeriodic) {
+          // issue-934, task will be cancelled
+          // even it is a periodic task such as
+          // setInterval
+          return;
+        }
         if (typeof data.handleId === NUMBER) {
           // in non-nodejs env, we remove timerId
           // from local cache
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "phantomjs": "^2.1.7",
     "promises-aplus-tests": "^2.1.2",
     "pump": "^1.0.1",
-    "rxjs": "5.4.2",
+    "rxjs": "^5.5.3",
     "selenium-webdriver": "^3.4.0",
     "systemjs": "^0.19.37",
     "ts-loader": "^0.6.0",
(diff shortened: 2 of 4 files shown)


JiaLiPassion:
@adrienverge, thank you for confirm, we don't need to wait for rxjs new version, because `zone.js` doesn't depends on `rxjs`, but I am not sure when this will be merged.

sarunint:
Since `rxjs` issue mentioned is resolved, shouldn't you change the `rxjs` version to `^5.4.2` again?

JiaLiPassion:
@sarunint , yes, I will update the `package.json`.

PR #943, opened 2017-10-27
fix(mocha): use global if window does not exist

justindujardin:
- `window` does not exist in NativeScript applications.
- this removes a hacky workaround where you assign `global.window` to `global` before including the mocha patch.
- preserve existing behavior by using window first if it exists.

### DIFF
--- a/lib/mocha/mocha.ts
+++ b/lib/mocha/mocha.ts
@@ -170,4 +170,4 @@
 
   })(Mocha.Runner.prototype.runTest, Mocha.Runner.prototype.run);
 
-})(window);
+})(typeof window !== 'undefined' && window || typeof self !== 'undefined' && self || global);


lib/mocha/mocha.ts
```diff
@@ -170,4 +170,4 @@
 
   })(Mocha.Runner.prototype.runTest, Mocha.Runner.prototype.run);
 
-})(window);
\ No newline at end of file
+})(typeof window !== 'undefined' ? window : global);
```
JiaLiPassion: please use 
```javascript
})(typeof window !== 'undefined' && window || typeof self !== 'undefined' && self || global);
```

which is used here https://github.com/angular/zone.js/blob/master/lib/zone.ts#L1344

justindujardin: done

justindujardin:
The CI build appears to be broken unrelated to this change. System.js is failing to import some rxjs operators.

JiaLiPassion:
@justindujardin , yes, current CI is broken, you can just ignore that.

PR #952, opened 2017-11-03
add missing star in zone jsdoc

evmar:
Without this extra star, the Closure commands in the jsdoc are ignored.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-/*
+/**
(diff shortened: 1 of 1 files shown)


alexeagle:
maybe the root cause here is a missing tslint check?

JiaLiPassion:
@alexeagle , I would like to add the `tslint` check, I would also like to add `closure compiler` of `zone.js` into the `Travis CI`, could you tell me which `closure compiler` I should use? Thank you!

evmar:
This actually still doesn't work, let me experiment more.

PR #972, opened 2017-12-01
fix(core): fix #969, should be able to access target/eventName info in onScheduleTask

### LINKED ISSUE #969: Need a way to access target for eventTasks
In newest versions of zone.js task.data.target is removed, is there anyway to access target for eventTasks?

JiaLiPassion:
fix #969.

Should be able to access `target/eventName/capture` information when using `addEventListener` in `onScheduleTask` callback. 

in `onScheduleTask`.

```javascript
   onScheduleTask: (parentZoneDelegate: ZoneDelegate, currentZone: Zone, targetZone: Zone,
                           task: Task): any => {
            const target = (task.data as any).taskData.target;
            const eventName = (task.data as any).taskData.eventName;
            const capture = (task.data as any).taskData.capture;
            return parentZoneDelegate.scheduleTask(targetZone, task);
    }
```

in `onInvokeTask`, `onCancelTask`.

```javascript
onInvokeTask(
                 parentZoneDelegate: ZoneDelegate, currentZone: Zone, targetZone: Zone, task: Task,
                 applyThis: any, applyArgs: any) {
      const target = (task as any).target;
      const eventName = (task as any).eventName;
      const capture = (task as any).capture;
}
```

it is not consistent because of `performance` concern (don't want to create so many object).

### DIFF
--- a/lib/common/events.ts
+++ b/lib/common/events.ts
@@ -396,12 +396,23 @@ export function patchEventTarget(
         taskData.isExisting = isExisting;
 
         const data = useGlobalCallback ? OPTIMIZED_ZONE_EVENT_TASK_DATA : null;
+
+        // keep taskData into data to allow onScheduleEventTask to acess the task information
+        if (data) {
+          (data as any).taskData = taskData;
+        }
+
         const task: any =
             zone.scheduleEventTask(source, delegate, data, customScheduleFn, customCancelFn);
 
         // should clear taskData.target to avoid memory leak
         // issue, https://github.com/angular/angular/issues/20442
         taskData.target = null;
+
+        // need to clear up taskData because it is a global object
+        if (data) {
+          (data as any).taskData = null;
+        }
 
         // have to save those information to task in case
         // application may call task.zone.cancelTask() directly
--- a/test/browser/browser.spec.ts
+++ b/test/browser/browser.spec.ts
@@ -378,6 +378,40 @@ describe('Zone', function() {
         expect(eventListenerSpy).toHaveBeenCalled();
       });
 
+      it('should be able to access addEventListener information in onScheduleTask', function() {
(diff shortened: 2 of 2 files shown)


lib/common/events.ts
```diff
@@ -396,9 +396,16 @@ export function patchEventTarget(
         taskData.isExisting = isExisting;
 
         const data = useGlobalCallback ? OPTIMIZED_ZONE_EVENT_TASK_DATA : null;
+
+        // keep taskData into data to allow onScheduleEventTask to acess the task information
+        (data as any).taskData = taskData;
```
jahtalab: if the `useGlobalCallback` is false, data would be `null` and this line would fail.

JiaLiPassion: @jahtalab, yes, you are right, thank you.

jahtalab:
Is `useGlobalCallback` a configuration option? AFAICS it's always true.
I'm worried about if `useGlobalCallback` is false then the taskData would not be accessible.

JiaLiPassion:
@jahtalab , it is `configurable` when you want to patch some `new API`, but for the `built in` patch, it is always true, and currently it will be `false` in `nodejs` environment and `RTCPeerConnection` patch.

jahtalab:
Thanks @JiaLiPassion ,
Is there any timeline for the next release that could include this fix?

JiaLiPassion:
@jahtalab,  you are welcome and I am not sure the release date.

jahtalab:
Tests are failing because of rxjs changes in exposed API in the newer versions. I had to lock rxjs version to `5.4.2` to make them pass.

JiaLiPassion:
@jahtalab , yes, this is a known issue, the other PR has used the `rxjs 5.5.3` will fix this issue.

PR #980, opened 2017-12-18
fix(error): fix #979, should not attach non serializable info on error

### LINKED ISSUE #979: datacloneerror on postMessage in worker.
Zone modifies Promise by adding extra properties to object, so it cannot be serialized. Is there any workaround for this case?

JiaLiPassion:
fix #979.

- should not keep a non serializable `ZoneTask` on `error` when `Promise.reject`.
- and don't save the `ZoneTask` instead save the error stacks to avoid memory leak.
- only save the information when `longStackTraceZone` in there.

### DIFF
--- a/lib/common/promise.ts
+++ b/lib/common/promise.ts
@@ -17,6 +17,7 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
   const _uncaughtPromiseErrors: UncaughtPromiseError[] = [];
   const symbolPromise = __symbol__('Promise');
   const symbolThen = __symbol__('then');
+  const creationTrace = '__creationTrace__';
 
   api.onUnhandledError = (e: any) => {
     if (api.showUncaughtError()) {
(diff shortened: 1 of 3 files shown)


alexeyrv:
Hi. Have now the same issue again.
Angular 6 (CLI 6), Ecectron v.2.0, PDFjs v 2.0.489
Electron + Angular work good. Electron + PDFjs work good. 
When I put it together I have the issue with Worker or Promises.

I cannot understand reason and how to solve it.
What can I do to use Angular6/cli 6 with PDFjs in Electron framework ?
Thanks in advande.
```
`Failed to execute 'postMessage' on 'Worker': TypeError: https.request is not a function could not be cloned.
Error: Failed to execute 'postMessage' on 'Worker': TypeError: https.request is not a function could not be cloned.
    at MessageHandler.postMessage (pdf.js:1357)
    at sendStreamRequest (pdf.js:1174)
    at pdf.js:1239
````

JiaLiPassion:
@alexeyrv, I am not sure what is the error, could you post a reproduce repo?

Izayer:
Same error as @alexeyrv, im using https://github.com/maximegris/angular-electron (Angular 6 / Electron ) template, and i cant use https://github.com/VadimDez/ng2-pdf-viewer

`ERROR Error: Uncaught (in promise): DataCloneError: Failed to execute 'postMessage' on 'Worker': TypeError: http.request is not a function could not be cloned.
Error: Failed to execute 'postMessage' on 'Worker': TypeError: http.request is not a function could not be cloned.
    at MessageHandler.postMessage (pdf.js:12334)
    at sendStreamRequest (pdf.js:12151)`

PR #982, opened 2017-12-20
fix(bazel): add zone-testing bundle for bazel build

### LINKED ISSUE #404: 0.6.14 Uncaught Error: Missing: SyncTestZoneSpec for Angular RC4 tests
Since 0.6.14 I'm getting `Uncaught Error: Missing: SyncTestZoneSpec` for my Angular2 RC4 unit tests. They work with 0.6.13.

 Affected [project](https://github.com/Sirimangalo/meditation-plus-angular) and [failing build](https://travis-ci.org/Sirimangalo/meditation-plus-angular/builds/152888816#L2389) as a reference.

I'm using [angular2-webpack-starter](https://github.com/AngularClass/angular2-webpack-starter) for my Angular2 project (like many others) which references zone.js via `~0.6.12` ([package.json](https://github.com/AngularClass/angular2-webpack-starter/blob/4cf9947347663d8a61b7c7138a68e2d125bc39f9/package.json#L92)). This issue is likely affecting a lot of people.

JiaLiPassion:
fix #404.
add `zone-testing-bundle.js` bundle and `zone-testing.js`.

1. for `bazel` build for https://github.com/angular/angular/issues/21048#issuecomment-352795938
2. to resolve import order issue. 

- zone-testing-bundle.js
in this bundle, following modules are loaded.

```javascript
import '../zone';
...

import '../zone-spec/long-stack-trace';
import '../zone-spec/proxy';
import '../zone-spec/sync-test';
import '../jasmine/jasmine';
import '../zone-spec/async-test';
import '../zone-spec/fake-async-test';
```

@alexeagle , please review.

- zone-testing.js, fix #404, zone test module load order issue.
in this file, only test related files are loaded (in correct order).
so in `angular/cli` generated `src/test.ts`, user don't need to load individual zone test files themselves, they  can just load

```javascript
import 'zone.js/dist/zone-testing';
```


gulpfile.js
```diff
@@ -104,15 +104,33 @@ gulp.task('build/zone.js', ['compile-esm'], function(cb) {
...
+gulp.task('build/zone-testing-bundle.min.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/browser/rollup-test-main.ts', 'zone-testing-bundle.min.js', true, cb);
+});
+
+// Zone test related files for the browser.
+gulp.task('build/zone-testing.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/testing/zone-testing.ts', 'zone-testing.js', false, cb);
+});
+
+gulp.task('build/zone-testing.min.js', ['compile-esm'], function(cb) {
```
alexeagle: IIUC there is no benefit to using a minified script in tests, the only possible thing someone could do is make their debugging harder. They don't care about how many bytes are loaded in Karma. what do you think?

gulpfile.js
```diff
@@ -104,15 +104,33 @@ gulp.task('build/zone.js', ['compile-esm'], function(cb) {
...
+  return generateScript('./lib/browser/rollup-main.ts', 'zone.min.js', true, cb);
+});
+
+// Zone test bundle for the browser.
+gulp.task('build/zone-testing-bundle.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/browser/rollup-test-main.ts', 'zone-testing-bundle.js', false, cb);
+});
+
+gulp.task('build/zone-testing-bundle.min.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/browser/rollup-test-main.ts', 'zone-testing-bundle.min.js', true, cb);
```
alexeagle: same as comment below about minified test bundle

gulpfile.js
```diff
@@ -246,6 +264,10 @@ gulp.task('build', [
   'build/zone.js',
   'build/zone.js.d.ts',
   'build/zone.min.js',
+  'build/zone-testing.js',
```
alexeagle: the scripts needed are different depending if you are testing in node or in a browser.

For node, you need to start with zone-node.js, like Misko did in this PR:
https://github.com/angular/angular/pull/21053/files#diff-8ac936419da1e3d027833281525dba0e

Maybe we should also publish `zone-node-testing.js`?

lib/testing/zone-testing.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// load test related files into bundle in correct order
+import '../zone-spec/long-stack-trace';
+import '../zone-spec/proxy';
+import '../zone-spec/sync-test';
+import '../jasmine/jasmine';
```
alexeagle: this makes an assumption about zone and jasmine being installed as siblings. Why not just `import jasmine`?

gulpfile.js
JiaLiPassion: yes, you are right, we don't need minified bundle.

gulpfile.js
JiaLiPassion: yes, I agree, I will remove them.

gulpfile.js
JiaLiPassion: got it, I will make `zone-node-testing-bundle`

lib/testing/zone-testing.ts
JiaLiPassion: Sorry I am not sure I understand your suggestion, I just import `../jasmine/jasmine` which include `zone patch` for jasmine method.

alexeagle: Oh, I see, I thought this was importing jasmine library itself, but it's https://github.com/angular/zone.js/blob/master/lib/jasmine/jasmine.ts

The user does need to ensure that jasmine itself is loaded before the zone-testing bundle

JiaLiPassion: yes, if `jasmine` is not loaded, here will throw error `Missing: jasmine.js`

JiaLiPassion:
@alexeagle , thank you for review, I have updated the PR. please review.

- add `zone-testing-node-bundle.js` for `node build`, compare with @mhevery 's commit, https://github.com/angular/angular/pull/21053/files#diff-8ac936419da1e3d027833281525dba0e, 
`jasmine patch` is added.

- for browser, `zone-testing-bundle.js` include `zone.js` itself and `testing` related files.

-------------------
- in this PR, there is another `zone-testing.js` which doesn't have direct relation with `bazel build`, it just make `angular/cli` produced `src/test.ts` easier and don't need to care about the import order. in `zone-testing.js`, `zone.js` itself not included.

alexeagle:
LGTM, I think we should probably let Misko look before merging?

/cc @hansl about improving the UX for CLI users

JiaLiPassion:
@alexeagle , got it. thank you!

@mhevery , please review, and about `zone-testing-node-bundle.js`, it also include `zone jasmine patch`, please check it is correct or not. thank you.

PR #990, opened 2018-01-06
fix(core): fix #989, remove unuse code, use shorter name to reduce bundle size

### LINKED ISSUE #989: 3.4kb size regressions between 0.8.16 and 0.8.19
Size of various polyfills.js build for angular.io with zone.js version changing as the only variable:

v0.8.16 = 37259 bytes
v0.8.17 = 38245 bytes
v0.8.18 = 40274 bytes
v0.8.19 = 40684 bytes

3.4kb is a lot of bytes. Is the increase justified? Can we reduce the size and prevent further regressions?

Justification: zone.js is a prolyfill and as such it needs to be downloaded, parsed and evaluated before anything else gets its turn. This makes the size of zone.js a significant contributor to the overall TTI.

JiaLiPassion:
fix #989.
1. remove unused code
2. move `getUserMedia` and `electron` related code to standalone files.
3. use `shorter` variable and function name to reduce bundle size.
now the size of `zone.min.js` will be `38510` or `38955` bytes. (the size will change when I rebuild)

4. add file size check logic in `travis` build.
5. change `func.apply` to `func.call` for better performance.

now I set a `limitation` of `zone.min.js` to `40000` bytes.

@mhevery , please review the idea is ok or not. thank you!

### DIFF
--- a/lib/browser/define-property.ts
+++ b/lib/browser/define-property.ts
@@ -38,6 +38,7 @@ export function propertyPatch() {
   };
 
   Object.create = <any>function(obj: any, proto: any) {
+    // o is 'object' string
     if (typeof proto === p && !Object.isFrozen(proto)) {
       Object.keys(proto).forEach(function(prop) {
         proto[prop] = rewriteDescriptor(obj, prop, proto[prop]);
@@ -89,6 +90,7 @@ function _tryDefineProperty(obj: any, prop: string, desc: any, originalConfigura
     if (desc.configurable) {
       // In case of errors, when the configurable flag was likely set by rewriteDescriptor(), let's
       // retry with the original flag value
+      // o is 'undefined' string
       if (typeof originalConfigurableFlag == o) {
         delete desc.configurable;
       } else {
--- a/lib/browser/define-property.ts
+++ b/lib/browser/define-property.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {zoneSymbol} from '../common/utils';
+import {o, p, zoneSymbol} from '../common/utils';
 /*
  * This is necessary for Chrome and Chrome mobile, to enable
  * things like redefining `createdCallback` on an element.
@@ -17,9 +17,6 @@ const _getOwnPropertyDescriptor = (Object as any)[zoneSymbol('getOwnPropertyDesc
     Object.getOwnPropertyDescriptor;
 const _create = Object.create;
 const unconfigurablesKey = zoneSymbol('unconfigurables');
-const PROTOTYPE = 'prototype';
-const OBJECT = 'object';
-const UNDEFINED = 'undefined';
 
 export function propertyPatch() {
   Object.defineProperty = function(obj, prop, desc) {
@@ -27,7 +24,7 @@ export function propertyPatch() {
       throw new TypeError('Cannot assign to read only property \'' + prop + '\' of ' + obj);
     }
     const originalConfigurableFlag = desc.configurable;
-    if (prop !== PROTOTYPE) {
+    if (prop !== 'prototype') {
       desc = rewriteDescriptor(obj, prop, desc);
     }
     return _tryDefineProperty(obj, prop, desc, originalConfigurableFlag);
@@ -41,7 +38,7 @@ export function propertyPatch() {
   };
 
   Object.create = <any>function(obj: any, proto: any) {
-    if (typeof proto === OBJECT && !Object.isFrozen(proto)) {
+    if (typeof proto === p && !Object.isFrozen(proto)) {
       Object.keys(proto).forEach(function(prop) {
         proto[prop] = rewriteDescriptor(obj, prop, proto[prop]);
       });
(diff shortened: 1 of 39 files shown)


lib/common/utils.ts
```diff
@@ -11,50 +11,51 @@
  * @suppress {undefinedVars,globalThis,missingRequire}
  */
 
+// issue #989, to reduce bundle size, use short name
+export const a = Object.getOwnPropertyDescriptor;
```
mhevery: there is no need to have a single char letters here. You can have long names, they will get renamed using uglify. Single letters are hard on readability.

lib/common/utils.ts
```diff
@@ -11,50 +11,51 @@
...
+export const e = Object.create;
+export const f = Array.prototype.slice;
+export const g = 'addEventListener';
+export const h = 'removeEventListener';
+export const i = (Zone as any).s(g);
+export const j = (Zone as any).s(h);
+export const k = 'true';
+export const l = 'false';
+export const m = '__zone_symbol__';
+export const n = 'function';
```
mhevery: i think `typeof x == 'string` is faster than `typeof x == STRING_CONST`. That is because in one case the jitter can create the check code which directly checks type to be string, in tho other case it has to generate a more complex code which could be actual type and than compare it to the constant. I would revert this.

lib/browser/define-property.ts
```diff
@@ -41,7 +38,8 @@ export function propertyPatch() {
   };
 
   Object.create = <any>function(obj: any, proto: any) {
-    if (typeof proto === OBJECT && !Object.isFrozen(proto)) {
+    // o is 'object' string
```
alfaproject: `o`?

lib/browser/property-descriptor.ts
```diff
@@ -265,21 +265,19 @@ export function propertyDescriptorPatch(api: _ZonePrivate, _global: any) {
     return;
   }
 
-  const supportsWebSocket = typeof WebSocket !== 'undefined';
+  const supportsWebSocket = typeof WebSocket !== o;
   if (canPatchViaPropertyDescriptor()) {
     const ignoreProperties: IgnoreProperty[] = _global.__Zone_ignore_on_properties;
     // for browsers that we can patch the descriptor:  Chrome & Firefox
     if (isBrowser) {
+      const w: any = window;
```
mhevery: local vars get renamed. Calling it `w` makes it hard to read. Could we call it `window` and that way we don't have to change the code bellow? I know `window = window` would not work, but something like that?

lib/common/utils.ts
```diff
@@ -11,50 +11,72 @@
  * @suppress {undefinedVars,globalThis,missingRequire}
  */
 
+// issue #989, to reduce bundle size, use short name
```
alfaproject: You shouldn't do the job of a compressor.
In fact, you are probably adding entropy.

P.S.: all of these extra imports/exports add up to browser parsing/interpreting time as well

lib/zone.ts
```diff
@@ -1338,7 +1343,19 @@ const Zone: ZoneType = (function(global: any) {
     return '__zone_symbol__' + name;
   }
 
-
   performanceMeasure('Zone', 'Zone');
+  const z: any = Zone.prototype;
+  z.w = z.wrap;
```
mhevery: I would prefer that we don't pollute the Zone public API with these short names. I realized they are not part of `.d.ts` but they will show up in debugger.

How about this idea
`Zone.current` could be replaced with `getZoneCurrent()` which is declared in `util` file. The uglify will rename the `getZoneCurrent` to something short like `x`.

`Zone.current.wrap(...)` could be `wrapWithCurrentZone(...)`

and so on. What do you think?

lib/zone.ts
```diff
@@ -1338,7 +1343,19 @@ const Zone: ZoneType = (function(global: any) {
...
+  z.f = z.fork;
+  z.r = z.run;
+  z.rg = z.runGuarded;
+  z.rt = z.runTask;
+  z.st = z.scheduleTask;
+  z.sc = z.scheduleMacroTask;
+  z.si = z.scheduleMicroTask;
+  z.se = z.scheduleEventTask;
+  z.ct = z.cancelTask;
+  (Zone as any).l = Zone.__load_patch;
```
mhevery: same thing here, helpers in `util.ts` can hide the long name without loosing readability

lib/browser/property-descriptor.ts
JiaLiPassion: yes, you are right, I have changed the variable to `internalWindow`.

lib/common/utils.ts
JiaLiPassion: Thanks for pointing out this one, I have misunderstanding about uglfiyjs, I have changed the variable name to long names.

lib/common/utils.ts
JiaLiPassion: Got it, I have reverted all `typeof x ===` parts.

lib/zone.ts
JiaLiPassion: Yes, this is a better solution, I will create those helper functions.

lib/zone.ts
JiaLiPassion: Got it.

lib/browser/define-property.ts
JiaLiPassion: yes, I have reverted this one, thank you!

lib/common/utils.ts
JiaLiPassion: Yes, you are right, I have misunderstanding about uglifyjs, I have modified the PR.

IgorMinar:
awesome! thanks!

PR #993, opened 2018-01-13
add OriginalDelegate prop to Function::toString

gdh1995:
This PR complements PR #686 , making `Function.prototype.toString.toString()` also returns `"function toString() { [native code] }"`, which should be a little safer.

Update: `Function.prototype.toString.call(Function.prototype.toString)` will return `[native code]`, too.

### DIFF
--- a/lib/common/to-string.ts
+++ b/lib/common/to-string.ts
@@ -9,15 +9,14 @@ import {zoneSymbol} from './utils';
 
 // override Function.prototype.toString to make zone.js patched function
 // look like native function
-Zone.__load_patch('toString', (global: any, Zone: ZoneType) => {
+Zone.__load_patch('toString', (global: any) => {
   // patch Func.prototype.toString to let them look like native
-  const originalFunctionToString = (Zone as any)['__zone_symbol__originalToString'] =
-      Function.prototype.toString;
+  const originalFunctionToString = Function.prototype.toString;
 
   const ORIGINAL_DELEGATE_SYMBOL = zoneSymbol('OriginalDelegate');
   const PROMISE_SYMBOL = zoneSymbol('Promise');
   const ERROR_SYMBOL = zoneSymbol('Error');
-  Function.prototype.toString = function() {
+  const newFunctionToString = function toString() {
     if (typeof this === 'function') {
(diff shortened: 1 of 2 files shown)


JiaLiPassion:
@gdh1995 , thank you for making the PR, could you add a test case in https://github.com/angular/zone.js/blob/master/test/common/toString.spec.ts?

gdh1995:
@JiaLiPassion Thanks, I've added a new test.

lib/common/to-string.ts
```diff
@@ -42,6 +42,8 @@ Zone.__load_patch('toString', (global: any, Zone: ZoneType) => {
     }
     return originalFunctionToString.apply(this, arguments);
   };
+  (newFunctionToString as any)[ORIGINAL_DELEGATE_SYMBOL] = originalFunctionToString;
```
JiaLiPassion: current version, we attached `originalFunctionToString` to `Zone`, and if we attached it to `Function.prototype`, we don't need to attached it to `Zone` any more.

so please just change this line https://github.com/gdh1995/zone.js/blob/b98f284030a4296a0e476728829298515edc3b0d/lib/common/to-string.ts#L14
```javascript
const originalFunctionToString = Function.prototype[ORIGINAL_DELEGATE_SYMBOL] =
      Function.prototype.toString;
```

JiaLiPassion:
@gdh1995 , I just leave a comment, we don't need to keep `original toString` in two places, and please `squash` your commits by `git rebase`.

gdh1995:
Thanks. I've rebased those commits.

lib/common/to-string.ts
```diff
@@ -11,13 +11,12 @@ import {zoneSymbol} from './utils';
 // look like native function
 Zone.__load_patch('toString', (global: any, Zone: ZoneType) => {
```
JiaLiPassion: Please also remove `Zone: ZoneType` here to reduce the bundle size.

JiaLiPassion:
@gdh1995 , thank you, please also remove the unused `Zone` in parameter list.

gdh1995:
The commit has been rebased secondly.

PR #994, opened 2018-01-16
feat(patch): fix #828, patch socket.io client

JiaLiPassion:
fix #828, patch socket.io-client.

I created a sample repo in here, https://github.com/JiaLiPassion/zone-socketio

PR #999, opened 2018-01-25
fix(core): don't update taskCounts of setInterval

JiaLiPassion:
fix https://github.com/angular/angular/issues/20970#issuecomment-360444160

don't update taskCount or trigger `hasTask` for `setInterval`.
otherwise the `zone` will always be `unstable`.

### update
now the CI failed because new version `jasmine` complains,  I have pinned the version of `jasmine` in #994.

### DIFF
--- a/test/common/setInterval.spec.ts
+++ b/test/common/setInterval.spec.ts
@@ -60,10 +60,15 @@ describe('setInterval', function() {
     }, null, null, 'unit-test');
   });
 
-  it('should not cancel the task after invoke the setInterval callback', (done) => {
+  it('setInterval should not update task count to trigger onHasTask', (done: Function) => {
     const logs: HasTaskState[] = [];
+    let task: Task;
     const zone = Zone.current.fork({
       name: 'interval',
+      onScheduleTask: (delegate: ZoneDelegate, currentZone: Zone, targetZone: Zone, t: Task) => {
+        task = t;
+        return delegate.scheduleTask(targetZone, task);
+      },
       onHasTask:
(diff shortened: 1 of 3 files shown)


test/common/setInterval.spec.ts
```diff
@@ -60,10 +60,15 @@ describe('setInterval', function() {
     }, null, null, 'unit-test');
   });
 
-  it('should not cancel the task after invoke the setInterval callback', (done) => {
+  it('setInterval should not update task count to trigger onHasTask', (done: Function) => {
```
mhevery: I don't think this is right. Why should `setInterval ` be exempt from tasks?

JiaLiPassion: @mhevery , before this PR, I believe `setInterval` was exempt from tasks but still not correctly `updateTaskCount`.

https://github.com/angular/zone.js/blob/master/lib/zone.ts#L808

if the task is `isPeriodic`, it will be reset to `scheduled` state and it will not `updateTaskCount` after `runTask`, I think this is the correct behavior.
so it should not `updateTaskCount` when `scheduleTask` and `cancelTask` either.  so it will not trigger `onHasTask` wrongly.

JiaLiPassion:
@mhevery
I just re-think it, you are right, we should not exempt interval from tasks, the interval should behave like

call updateTaskCount +1 and trigger onHasTask when scheduleTask.
will ** not ** call updateTaskCount and trigger onHasTask after runTask.
will call updateTaskCount -1 and trigger onHasTask when cancelTask.
so if some interval task is active, zone will not be stable, but it is the correct behavior.

I will close this PR, the current behavior is correct. thank you!

PR #1009, opened 2018-01-31
feat(jasmine): support Date.now in fakeAsyncTest

JiaLiPassion:
fix https://github.com/angular/angular/issues/8678

support `Date.now` in `fakeAsyncTest`. there are `3` features in this PR.

- support auto patch `Date.now` and `new Date()` in `fakeAsync`.

```javascript
 fakeAsyncTestZone.run(() => {
        const start = Date.now();
        testZoneSpec.tick(100);
        const end = Date.now();
        expect(end - start).toBe(100);
  });

 fakeAsyncTestZone.run(() => {
        const start = new Date();
        testZoneSpec.tick(100);
        const end = new Date();
        expect(end.getTime() - start.getTime()).toBe(100);
  });
```

- automatically run a `fakeAsync` test when `jasmine.clock().install` is called.

```javascript
beforeEach(() => {
      jasmine.clock().install();
    });

    afterEach(() => {
      jasmine.clock().uninstall();
    });

    it('should get date diff correctly', () => {  // we don't need fakeAsync here.
      // automatically run into fake async zone, because jasmine.clock() is installed.
      const start = Date.now();
      jasmine.clock().tick(100);
      const end = Date.now();
      expect(end - start).toBe(100);
    });
```

- rxjs `Scheduler` support, need to import `zone.js/dist/zone-patch-rxjs-fake-async`.

```javascript
    import '../../lib/rxjs/rxjs-fake-async';
    it('should get date diff correctly', (done) => {
      fakeAsyncTestZone.run(() => {
        let result = null;
        const observable = new Observable((subscribe: any) => {
          subscribe.next('hello');
        });
        observable.delay(1000).subscribe(v => {
          result = v;
        });
        expect(result).toBeNull();
        testZoneSpec.tick(1000);
        expect(result).toBe('hello');
        done();
      });
    });
```

@mhevery , @vikerman , please review, thank you!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -276,6 +276,14 @@ gulp.task('build/rxjs.min.js', ['compile-esm'], function(cb) {
   return generateScript('./lib/rxjs/rxjs.ts', 'zone-patch-rxjs.min.js', true, cb);
 });
 
+gulp.task('build/rxjs-fake-async.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/rxjs/rxjs-fake-async.ts', 'zone-patch-rxjs-fake-async.js', false, cb);
+});
+
+gulp.task('build/rxjs-fake-async.min.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/rxjs/rxjs-fake-async.ts', 'zone-patch-rxjs-fake-async.min.js', true, cb);
+});
+
 gulp.task('build/closure.js', function() {
   return gulp.src('./lib/closure/zone_externs.js')
              .pipe(gulp.dest('./dist'));
@@ -327,6 +335,8 @@ gulp.task('build', [
   'build/sync-test.js',
   'build/rxjs.js',
   'build/rxjs.min.js',
+  'build/rxjs-fake-async.js',
+  'build/rxjs-fake-async.min.js',
   'build/closure.js'
 ]);
 
--- a/lib/jasmine/jasmine.ts
+++ b/lib/jasmine/jasmine.ts
@@ -35,6 +35,8 @@
   // a `beforeEach` or `it`.
   const syncZone = ambientZone.fork(new SyncTestZoneSpec('jasmine.describe'));
 
+  const symbol = Zone.__symbol__;
+
   // This is the zone which will be used for running individual tests.
   // It will be a proxy zone, so that the tests function can retroactively install
   // different zones.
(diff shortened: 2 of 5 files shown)


Jopie64:
@JiaLiPassion 
For your info, your last test leaks a subscription.
Instead of this:

```javascript
const observable = new Observable((subscribe: any) => {
  subscribe.next('hello');
});
observable.delay(1000).subscribe(v => {
  result = v;
});
```

You can write this to fix it:
```javascript
Observable.of('hello').delay(1000).subscribe(v => result = v);
```
(`Observable.of()` completes after sending the single value)

Anyway, thanks for making my library obsolete! 👍 XD

JiaLiPassion:
@Jopie64 , thank you for the comment.

I didn't use `Observable.of()` because there is a bug in another patch `zone-patch-rxjs`. I have updated the test case to `complete subscriber`.

alfaproject:
I couldn't make this work, so I'm a bit confused. `angular-testing-ext` npm package works out of the box as expected, and I was hoping to replace it, but couldn't make this work. ):

JiaLiPassion:
@alfaproject , with https://github.com/angular/angular/pull/23108 merged, it should work, current `angular` not using newest version of `zone.js`.

alfaproject:
But can't I use the latest version of `zone.js` with Angular 5? If not, I guess I'll have to wait then. ):

alfaproject:
@JiaLiPassion that works, thank you for helping!

ohjames:
Tried this out but my tests are still broken as `clock.mockDate(new Date(...));` has no effect now.

ohjames:
Okay so `clock.mockDate(new Date(...));` but only if used inside of your `it`, not when called in the `beforeEach` along with `jasmine.clock().install()` (before or after). Inconvenient when you have many tests grouped in a `describe` that should all be using the same mocked date.

schippie:
@JiaLiPassion  Thanks for that, it's finally working though a note i figure out the had way is that you have to import the delay operator above it for this to function properly in angular 5:

```typescript
import { Observable } from "rxjs/Observable";
import "rxjs/add/operator/delay";

declare let Zone: any;
const async = Zone[Zone.__symbol__("asyncTest")];
const {fakeAsync, tick, flush, flushMicrotasks, discardPeriodicTasks, resetFakeAsyncZone} = Zone[Zone.__symbol__(
	"fakeAsyncTest")];

import 'zone.js/dist/zone-patch-rxjs-fake-async';
```

vicneanschi:
Thanks @JiaLiPassion!

afeinber:
@JiaLiPassion Where are we supposed to get fakeAsyncTestZone from to call fakeAsyncTestZone.run?

PR #1012, opened 2018-02-03
feat(patch): fix #1011, patch ResizeObserver

### LINKED ISSUE #1011: ResizeObserver issue
I'm using [ResizeObserver](https://caniuse.com/#feat=resizeobserver) for which I use a [polyfill](https://github.com/que-etc/resize-observer-polyfill) because its browser support is limited. Now, recently, chrome 64 natively supports this API with the result that changes trigged by ResizeObserver don't trigger `changedetection`  in my angular app. 

[stackblitz demo](https://stackblitz.com/edit/resize-observer?file=polyfills.ts)

```typescript
const observer = new window.ResizeObserver(entries => {
          this.observedWidth = entries[0].contentRect.width;  // Outside a zone??!!
});
observer.observe(this.resizable.nativeElement);
```
The expected behaviour can be seen in firefox/safari and chrome 63 of course. 

Now I don't know that much about Zone.js. but I would expect Zone.js to take care of this?

For more details:
[stackoverflow question](https://stackoverflow.com/questions/48579545/angular-async-process-not-triggering-changedetection-only-in-chrome-64)

JiaLiPassion:
fix #1011.

patch `ResizeObserver`.

support one observer patch different HtmlElements in different zones.

```javascript
    const observer1 = new ResizeObserver((entries: any, ob: any) => {
      console.log('observer1 zone', Zone.current);
      for (const entry of entries) {
        console.log('observer1 resize entry target', entry.target);
        console.log('observer1 resize entry size', entry.contentRect);
      }
    });

    observer1.observe(div1);
    observer1.observe(div2);
    Zone.root.run(() => {
      observer1.observe(div3);
    });
```

full sample can be find here, https://github.com/JiaLiPassion/zone-resize-observer, 
only work in `Chrome 64`.

### DIFF
--- a/NON-STANDARD-APIS.md
+++ b/NON-STANDARD-APIS.md
@@ -170,7 +170,7 @@ In electron, we patched the following APIs with `zone.js`
 
 ## Usage.
 
-add/update following line into `polyfill.ts`.
+add following line into `polyfill.ts` after loading zone-mix.
 
 ```
 //import 'zone.js/dist/zone'; // originally added by angular-cli, comment it out
@@ -194,7 +194,6 @@ user need to patch `io` themselves just like following code.
     </script>
 ```
 
-
 please reference the sample repo [zone-socketio](https://github.com/JiaLiPassion/zone-socketio) about 
 detail usage.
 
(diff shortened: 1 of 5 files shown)


JiaLiPassion:
@mhevery , got it, I will add test cases, because currently it only support `Chrome 64`, and `travis saucelabs` only run on `Chrome 63`, so I didn't add those cases now, I will add cases to make them runnable in local.

JiaLiPassion:
@mhevery , I have added test cases, and tested locally with `Chrome 64`, please review. thank you.

PR #1014, opened 2018-02-07
feat(test): can handle non zone aware task in promise within AsyncTestZoneSpec

### LINKED ISSUE #1013: HasTaskState says there are no micro and macro tasks but I think there are outstanding tasks. Who is right?
Hello!

I wanted to use Zone.js for our project in E2E testing. We don't use Angular so we just wanted to use the raw Zone.js. In our app there are a lot of async tasks and I want every E2E test to wait until all async tasks are finished and then do a cleanup. Zone.js seems to help with this issue, but I can't get it to work. When I wait in `onHasTask` for `hasTaskState.macroTask === false && hasTaskState.microTask === false` I encounter situations where it seems that the E2E test finishes more often. I was able to create a small jsfiddle which demonstrates my problem: https://jsfiddle.net/r0sokrvg/5/

I'm not sure if it is an issue that Zone.js sets  both `hasTaskState.macroTask` and `hasTaskState.microTask` to false but there are still tasks to process. Maybe I'm just using Zone.js wrong. I tried many things but I think I'm missing some important point.

It would be great if someone could help me. 

Thanks a lot

JiaLiPassion:
fix #1013.
currently if there are some `async` task which is not `zone aware`, `AsyncTestZoneSpec` will not handle it. such as, 

```javascript
it('test jsonp', () => {
 let finished = false; 
  const asyncTestZoneSpec = new AsyncTestZoneSpec(() => {
     expect(finished).toBe(true);
   }, () => {}, 'async');
   asyncTestZone.run(() => {
     jsonp(url, () =>{
        // success callback
        finished = true;
     });
   });
}); 
 
```

this case will fail because `jsonp` is not a `zone aware` async operation, so finished will be still be false, when `AsyncTestZone` think all `microTasks` and `macroTasks` have been done (but jsonp is still running).

We can't handle all `non zone aware` async task, so when such kind of issue occurs, we can help user use `Zone.prototype.scheduleMicroTask` or `Zone.prototype.scheduleMacroTask` to handle them.

But some case is much more confusing the user with `promise`, such as, 

```javascript
it('test jsonp', () => {
 let finished = false; 
  const asyncTestZoneSpec = new AsyncTestZoneSpec(() => {
     expect(finished).toBe(true);
   }, () => {}, 'async');
   asyncTestZone.run(() => {
     new Promise((res, rej) => {
       jsonp(url, () =>{
        // success callback and resolve the promise
        res();
       });
     }).then(() => {
        finished = true;
     });
   });
}); 
 
```

now even with the `promise` and `promise.then`, the case still failed. Because 
- `jsonp` is not `zone aware`.
- so `new Promise((res, rej) => jsonp(...))` will not trigger any `async` operation (promise itself it not `async`).
- `promise.then` is also not `async`  until the `promise` is resolved.

But the asyncTestZone will think all `async` finished (in this case, no async happened) before `promise.then` is invoked and schedule a microTask. So the test case will still fail.

We can still handle this issue by make those `non zone aware` task to `zone aware`, but it will cost time, and in most case, user call `promise.then`, they believe promise will be resolved or rejected, so I think `zone.js` may handle it more friendly.

in this PR, if there is such case.

- promise is chained (means promise.then is called)
- parent promise is unresolved.

I will increase a chainedPromise number in AsyncTestZoneSpec. And when the chainedPromise's parent resolved and `schedule` a microTask, I will decrease the number, and when `AsyncTestZoneSpec` check whether should call `finish or not` not only check `hasPendingMicroTask/MacoTask` but also check is there any `promise.then` waiting for `non zone aware` promise.

```javascript
it('test jsonp', () => {
 let finished = false; 
  const asyncTestZoneSpec = new AsyncTestZoneSpec(() => {
     expect(finished).toBe(true);
   }, () => {}, 'async');
   asyncTestZone.run(() => {
     new Promise((res, rej) => {
       jsonp(url, () =>{
        // success callback and resolve the promise
        res();  // 2. here, when parent promise resolved, and a microTask is scheduled, decrease the chainedPromise number.
       });
     }).then(() => {  // 1. here, when promise.then is called, increase the chainedPromise number
        finished = true; 
     });
   });
}); 
```

@mhevery , could you take a look about this idea is ok or not? thank you!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -200,6 +200,14 @@ gulp.task('build/zone-patch-socket-io.min.js', ['compile-esm'], function(cb) {
   return generateScript('./lib/extra/socket-io.ts', 'zone-patch-socket-io.min.js', true, cb);
 });
 
+gulp.task('build/zone-patch-promise-testing.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/testing/promise-testing.ts', 'zone-patch-promise-test.js', false, cb);
+});
+
+gulp.task('build/zone-patch-promise-testing.min.js', ['compile-esm'], function(cb) {
+  return generateScript('./lib/testing/promise-testing.ts', 'zone-patch-promise-test.min.js', true, cb);
+});
+
 gulp.task('build/bluebird.js', ['compile-esm'], function(cb) {
     return generateScript('./lib/extra/bluebird.ts', 'zone-bluebird.js', false, cb);
 });
@@ -307,6 +315,8 @@ gulp.task('build', [
   'build/zone-patch-user-media.min.js',
   'build/zone-patch-socket-io.js',
   'build/zone-patch-socket-io.min.js',
+  'build/zone-patch-promise-testing.js',
+  'build/zone-patch-promise-testing.min.js',
   'build/zone-mix.js',
   'build/bluebird.js',
   'build/bluebird.min.js',
--- a/karma-dist.conf.js
+++ b/karma-dist.conf.js
@@ -20,5 +20,6 @@ module.exports = function (config) {
   config.files.push('dist/sync-test.js');
   config.files.push('dist/task-tracking.js');
   config.files.push('dist/wtf.js');
+  config.files.push('dist/zone-patch-promise-test.js');
   config.files.push('build/test/main.js');
 };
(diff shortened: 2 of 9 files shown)


mhevery:
LGTM, what is going on with the Travis?

JiaLiPassion:
@mhevery, thank you for review, travis failed for almost all pr because current test case or jasmine patch not compatible with newest version of jasmine.

I have fixed jasmine to 2.4.1 in this pr #994, Travis should pass.

I will continue to find out how to compatible with jasmine new version.

PR #1016, opened 2018-02-09
feat(jasmine): fix #1015, make jasmine patch compatible to jasmine 3.x

### LINKED ISSUE #1015: Jasmine 3.0.0 breaks Angular 5.2 tests
Duplicated from: https://github.com/jasmine/jasmine/issues/1496
## Expected Behavior
Tests work with Jasmine 2.9.1

## Current Behavior
With Jasmine 3.0.0, **all** tests fail with this error:
```
TypeError: Cannot read property 'run' of null
  at <Jasmine>
  at UserContext.<anonymous> home/bobn/Projects/app/node_modules/zone.js/dist/jasmine-patch.js:106:1)
  at <Jasmine>
  at ZoneDelegate.webpackJsonp.../../../../zone.js/dist/zone.js.ZoneDelegate.invokeTask home/bobn/Projects/app/node_modules/zone.js/dist/zone.js:421:1)
  at Zone.webpackJsonp.../../../../zone.js/dist/zone.js.Zone.runTask home/bobn/Projects/app/node_modules/zone.js/dist/zone.js:188:1)
  at drainMicroTaskQueue home/bobn/Projects/app/node_modules/zone.js/dist/zone.js:595:1)
  at <Jasmine>
```
Same behavior using Angular 5.2.3 and 5.2.4.
It seems as though **jasmine-patch.js** is no longer compatible with Jasmine 3.0.0.
## Your Environment

Ubuntu 16.04
Chrome 64.0.3282 (Linux)
package.json: **"zone.js": "^0.8.20"**
`ng -v`
> Angular CLI: 1.6.7
> Node: 8.8.1
> OS: linux x64
> Angular: 5.2.4
> ... animations, common, compiler, compiler-cli, core, forms
> ... http, language-service, platform-browser
> ... platform-browser-dynamic, router
> 
> <nowiki>@</nowiki>angular/cli: 1.6.7
> <nowiki>@</nowiki>angular-devkit/build-optimizer: 0.0.42
> <nowiki>@</nowiki>angular-devkit/core: 0.0.29
> <nowiki>@</nowiki>angular-devkit/schematics: 0.0.52
> <nowiki>@</nowiki>ngtools/json-schema: 1.1.0
> <nowiki>@</nowiki>ngtools/webpack: 1.9.7
> <nowiki>@</nowiki>schematics/angular: 0.1.17
> typescript: 2.7.1
> webpack: 3.10.0

JiaLiPassion:
Current `zone.js` patch `jasmine` to make jasmine method run in `ProxyZone`, so we can do `async`, `fakeAsync` to `spec`.

But the patch not work for `jasmine 2.9.x` and `jasmine 3.x` version, there are two issues, 

1. the first issue is for `jasmine 3.x`,  in the current patch, we have a global `testProxyZone` in `jasmine`, this `zone` s lifecycle is,

- created when `QueueRunner.prototype.execute` is invoked
- cleared (set to `null`) when `QueueRunner.prototype.onComplete` is triggered.

before `jamsine 3.x`, the patch work because there will only be one `onComplete` being triggered in `execute`.

but in `jasmine 3.x`, in `QueueRunner.prototype.execute`, there will be `sub QueueRunner` being initialized for `jasmineReporter`, so in one `execute`, there will be several `QueueRunner` and trigger multiple `onComplete`. The sequence will look like:

- QueueRunner.execute  -> testProxyZone is initialized
- sub QueueRunner.execute (reporter.start) -> testProxyZone is initialized again
- sub QueueRunner.onComplete -> testProxyZone is null
- QueueRunner spec run -> testProxyZone is null throw error

so we can't use `global` testProxyZone any more, we need to keep the `testProxyZone` inside `QueueRunner`, and jasmine QueueRunner has a `UserContext` which will be initialized when `new QueueRunner` and will be passed to all methods such as `beforeEach`, `it`, `describe` and so on, so in this PR, we use `UserContext` to store `testProxyZone` per `QueueRunner`.

2. the second issue is for `jasmine 2.9.x`, currently in test cases of `zone.js` we will dynamically run an `it` spec when `ifEnvSupports` return false, such kind of action is not allowed from `jasmine 2.9.x` because `it` can be only executed inside a `describe`. so we just print the `env not support warning` message.


JiaLiPassion:
Some tests failed, will continue to check it, and will do test to `angular` repository and some `test` angular project which use `async` or `fakeAsync`.

JiaLiPassion:
This PR is finished.
- support jasmine 2.x and jasmine 3.x
- because jasmine 3.x not support IE8 and IE9, so in travis, I will run all browsers for jasmine 2.9 and only run mainstream browsers for jasmine 3.x

@mhevery , please review, thank you.

Issue #995, opened 2018-01-19
set custom global zone


matthewjh:
Hello,

In my use case, I need to have a custom global zone. I want my own zone's hooks to be run, even when not using zone.run. Basically I need to replace the default Zone.current with my own zone, and use that as the "default" zone.

Is there any way to do this?

Thanks,
Matt

mleibman:
You can always just monkey patch it.
Something like this for `runTask`, and similar wrappers for the rest as needed:

```javascript
const runTask = Zone.prototype.runTask;
Zone.prototype.runTask = function(task, applyThis, applyArgs) {
  // Run your own handlers.
  return runTask.call(this, task, applyThis, applyArgs);
};
```

JiaLiPassion:
Will close this for now.

PR #1027, opened 2018-02-23
feat(core): fix #1024, #513, #995, add a helper lib to configure root zone

### LINKED ISSUE #1024: Could not intercept requestAnimationFrame from Google Maps
**Background:** Google Maps triggers `requestAnimationFrame` when zooming in and out of the map and when panning, I want to intercept these `requestAnimationFrame` via Zone.js to debounce it and may be improve performance on mobile devices...

I created a simple repro code on [StackBlitz](https://stackblitz.com/edit/typescript-ugi7tp?file=index.ts) (please open console to see logs)

The problem here is that I can't seem to intercept the `requestAnimationFrame ` on the `onScheduleTask` method defined on the ZoneSpec of the created Zone called `mapZone` (but I'm seeing intercepted events such as `setTimeout` inside the map when it's being instantiated) ... as you can see on the repro code, I ran the initialization of Google Maps inside the created Zone

I'm newbie on Zone.js, this probably might be a simple problem, but I can't seem to find a solution... what do you think am I missing here? :)

Thanks so much in advance!

JiaLiPassion:
fix #1024, #513, #995 
Add a help lib to configure root zone. It is a monkey-patch to zone (some kind of hack). And it is a lib need to be loaded by user themselves, so `zone public API` will not be damaged. 

- this is not a public api or formal feature of zone.js
- it will not be included in future tc39 proposal,
- and it is not by design, it is just a helper method to let you customize rootZone for some test and suppose to change in the future, please use it on your own risk

@mhevery , please review the idea is ok or not, I know it is not correct according to the design of Zone, but there a some use cases that user want to do that, so with or without this lib, user can still monkey-patch `Zone` to customize root zone, so I just provide the helper tool for them. thank you.


mhevery:
Not sure how I feel about this. Let me think about it...

mleibman:
FWIW, we ended up needing to customize a root zone in in GCP Console, but we currently do it via monkey-patching.

For us, it bridges the gap in the Zone.js API where interception is based purely on the current zone, but  zone configuration is fairly static. For Angular, it captures the 'outer' zone at start up, so any code calling NgZone.runOutsideAngular() would fall back to that and escape the interceptors added on by forking the 'inner' Angular zone.

PR #1028, opened 2018-02-23
fix(test): fix mocha compatible issue

JiaLiPassion:
fix https://github.com/angular/angular/issues/22150

in new version of mocha, the code like below 

```javascript
 beforeEach(async(() => {
    TestBed.configureTestingModule({
      declarations: [
        AppComponent
      ],
    }).compileComponents();
  }));

  it(`should have as title 'app'`, async(() => {
    console.log('title app test');
    const fixture = TestBed.createComponent(AppComponent);
    const app = fixture.debugElement.componentInstance;
    expect(app.title).to.equal('app');
  }));
```

will in the same `Mocha.Runner.prototype.run`.
so `Mocha.Runner.prototype.run` will run into `ProxyZone` instead of `rootZone`, we need to remove the `root zone` check, otherwise all mocha test will not begin.

also update `mocha` to 5.0.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,8 @@ script:
   - node_modules/.bin/gulp test/node
   - node simple-server.js 2>&1> server.log&
   - node ./test/webdriver/test.sauce.js
-  - yarn add jasmine@3.0.0 jasmine-core@3.0.0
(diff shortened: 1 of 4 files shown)


mhevery:
can you fix the conflicts?

JiaLiPassion:
@mhevery , I have rebased, please review.

PR #1044, opened 2018-03-10
fix(tsc): tsconfig.json strict:true

JiaLiPassion:
change `tsconfig.json`

```
"strict" : true
```

There are a lot of `null` error, and current this PR is `WIP`, to make it compilable, we need to modify some `interface` such as `Zone.current`, `Zone.currentTask`, `Zone.scheduleMacroTask`, `ZoneSpec.onXXX` because they all may return null or undefined.

@mhevery , do you think we should do this now? it will have breaking change to `zone API` and may affect `angular`. please review, thank you.


mhevery:
can you resolve this?

JiaLiPassion:
@mhevery , this PR is not finished, I want to confirm that if we add `strict : true`, some API may changes (add optional `?` or add `| null` type definition) such as 
```
scheduleMacroTask(
      source: string, callback: Function, data: TaskData, customSchedule: (task: Task) => void,
      customCancel: (task: Task) => void): MacroTask;
```

will become 

```
scheduleMacroTask(
      source: string, callback: Function, data?: TaskData, customSchedule?: (task: Task) => void,
      customCancel?: (task: Task) => void): MacroTask;
```

so should I do it now? it may have impact to `angular`

mhevery:
Yes we should do this. If angular changes will update...

JiaLiPassion:
@mhevery , got it, I will continue to finish this one.

JiaLiPassion:
@mhevery , I have finished this one and resolved the conflicts, please review. thank you.

PR #1045, opened 2018-03-11
feat(error): fix #975, can config how to load blacklist zone stack frames

### LINKED ISSUE #975: ZoneAwareError is 15-20x slower than native, should lazy-evaluate "stack"
Getting a stack trace can be an expensive operation, and both Firefox and Chrome/V8 (haven't checked IE/Edge) optimize it by formatting the unstructured stack trace of an error on demand, the first time the `stack` property is accessed (https://github.com/v8/v8/wiki/Stack-Trace-API#customizing-stack-traces).  

When building instrumentation tooling, I use this to minimize the overhead of capturing the stack trace using `new Error()` and then evaluate its `stack` property if and when needed. Unfortunately, `ZoneAwareError` does the trace evaluation and formatting in its constructor, so it ends up a lot slower, both from invoking the native error's formatting, and Zone.js's own overhead. I have also noticed it resulting in increased memory pressure, leading to more frequent GC pauses:

zoneawareerror

(The first half is running `new NativeError()`, the second is running `new Error()`.)

Benchmark demonstrating the issue - https://jsperf.com/zoneawareerror.
Note that this runs the benchmark with an unrealistically shallow call stack. The real life impact in a complex application is likely to be more pronounced.

Is it possible to make it evaluate and rewrite the `stack` on demand?

JiaLiPassion:
fix #975.

current `zone-error` provide several functionality.

1. handle `class MyError extends Error`, `new MyError() instanceof Error` will report `false` issue. 

2. remove `zone.js internal stack trace frames`.

3. add `zone` information to each `zone frame`.

The No2 and No3. feature  will slow down the ` new Error()` performance, 
So the motivation of this PR is to provide a flag to let user be able to disable No2 and No3.

 The flag is `__Zone_Error_BlacklistedStackFrames_policy`. And the available options is:

1. default: this is the default one, if you load `zone.js/dist/zone-error` without
 setting the flag, `default` will be used, and `BlackListStackFrames` will be available 
 when `new Error()`, you can get a `error.stack`  which is `zone stack free`. But this 
 will slow down `new Error()` a little bit.

2. disable: this will disable `BlackListZoneStackFrame` feature, and if you load
 `zone.js/dist/zone-error`, you will only get a `wrapped Error` which can handle
  `Error inherit` issue.

3. lazy: this is a feature to let you be able to get `BlackListZoneStackFrame` feature,
 but not impact performance. But as a trade off, you can't get the `zone free stack 
 frames` by access `error.stack`. You can only get it by access `error.zoneAwareStack`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,9 +39,13 @@ script:
   - node_modules/.bin/karma start karma-build-sauce-mocha.conf.js --single-run
   - node_modules/.bin/karma start karma-dist-sauce-selenium3-jasmine.conf.js --single-run
   - node_modules/.bin/karma start karma-build-sauce-selenium3-mocha.conf.js --single-run
+  - node_modules/.bin/karma start karma-dist-sauce-jasmine3.conf.js --single-run --errorpolicy=disable
+  - node_modules/.bin/karma start karma-dist-sauce-jasmine3.conf.js --single-run --errorpolicy=lazy
   - node_modules/.bin/gulp test/node
   - node_modules/.bin/gulp test/node -no-patch-clock
   - node_modules/.bin/gulp test/bluebird
+  - node_modules/.bin/gulp test/node/disableerror
+  - node_modules/.bin/gulp test/node/lazyerror
   - node simple-server.js 2>&1> server.log&
   - node ./test/webdriver/test.sauce.js
   - yarn add jasmine@3.0.0 jasmine-core@3.0.0 mocha@5.0.1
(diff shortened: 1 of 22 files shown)


mhevery:
Could you resolve conflicts please?

JiaLiPassion:
@mhevery , I have resolved the conflict, the CI is red because `saucelabs` iphone is not available, I have checked that all cases passed, please review, thank you!

mhevery:
Sorry, more conflicts.

JiaLiPassion:
@mhevery , yes, the `strict` PR will have many conflicts with all others, I will rebase them all.

PR #1049, opened 2018-03-16
fix(test): karma-dist should test bundle under dist

JiaLiPassion:
### update karma-dist.conf.js

current `karma-dist.conf.js` will load bundles under `dist` and the purpose is to test those built bundles instead of the dev version which under `build` folder.
so in `karma-dist.conf.js`, we will load 

```javascript
 config.files.push('dist/zone.js');
 config.files.push('dist/async-test.js');
...
``` 

But in current test cases will load some of the lib directly from `build` folder, for example in `async-test.spec.ts`, in current version, it will load 

```javascript
import '../../lib/zone-spec/async-test';
```
so the test case in fact will not test the bundle under `dist` but `build`.

-------------

in this PR, I load all needed bundles in `karma-dist.conf.js` and remove all directly import from test specs. And also modify the test entry classes to make sure currently test cases will be ok for both `build` test and `dist` test.

- in dev mode, test bundles under `build`.
- in dist mode, test bundles under `dist`.

### add ios 11, edge 15, android 7.1 to ci test.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,7 +43,9 @@ function generateScript(inFile, outFile, minify, callback) {
               'rxjs/Observable': 'Rx',
               'rxjs/Subscriber': 'Rx',
               'rxjs/Subscription': 'Rx',
+              'rxjs/Scheduler': 'Rx.Scheduler',
               'rxjs/scheduler/asap': 'Rx.Scheduler',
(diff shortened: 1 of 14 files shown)


mhevery:
can you resolve these?

PR #1055, opened 2018-03-23
fix(xhr): should invoke xhr task after onload is triggered

JiaLiPassion:
fix https://github.com/angular/angular/issues/20921
fix https://github.com/angular/angular/issues/17192

currently `zone.js` patch `xhr` as a `macroTask` and after `xhr.send`, `zone.js` will add a `readystatechange` listener, when `readystate=4`, the `macroTask` will be invoked.

this logic work fine, but because `xhr` can also add `onload` or `addEventListener('load', listener)`.
and the `load` event handler will run after `readystatechange`, so in the `load` event handler, the zone may have been stabled, and cause some issues. such as in https://github.com/angular/angular/issues/20921, the `chained http` will trigger multiple `zone.stable`, and https://github.com/angular/angular/issues/17192, the onload will run outside of fixture.whenStable.

So in this PR, I just make sure all `onload` is finished, then finally invoked the internal `readystatechange` task.


quanterion:
Does it have a chance to be merged?

PR #1070, opened 2018-04-05
fix(test): fix #1069, FakeDate should handle constructor parameter

### LINKED ISSUE #1069: FakeDate does not accept a timestamp value
Expected Behaviour: Passing a timestamp value into the Date object, as shown below, will produce a Date with that timestamp
Actual Behaviour: The timestamp is ignored and the current Date is used

In later versions (0.8.21 and onwards) this is happening, while using fakeAsync(worked with 0.8.20 and earlier)

The following code sample demonstrates this:

```
it('should set the right date', fakeAsync(() => {
    test();
    tick(1000);
    console.log("second", Date, new Date(0));
  }));

  async function test() {
    await (new Promise(resolve => {
      resolve();
    }));
    console.log("first", Date, new Date(0));
  }
```

The output is:
```
'first', function FakeDate() { ... }, Thu Apr 05 2018 11:05:19 GMT+0200 (CEST)
'second', function Date() { ... }, Thu Jan 01 1970 01:00:00 GMT+0100 (CET)
```

The first line of output is using the FakeDate object which is ignoring the value passed in and just producing the current Date.
The second line is working as expected.

This only occurs inside an async block, after an 'await' has been used

### LINKED ISSUE #1071: fakeAsync breaks Date constructor
I have reported this bug in Angular (https://github.com/angular/angular/issues/23223), but was asked to report it here. So here it goes:

Environment:
Zone.js v0.8.25, Angule 6.0.0-rc.2.

```
import { fakeAsync, TestBed } from '@angular/core/testing';

describe('Dates', () => {
  beforeEach(() => {
    TestBed.configureTestingModule({});
  });

  it('should deal with dates correctly in fakeAsync', fakeAsync(() => {
    const jsDate = new Date(1985, 2, 3, 12);
    jsDate.setFullYear(1985);

    expect(isNaN(jsDate.getTime())).toBeFalsy();
    expect(jsDate.getFullYear()).toBe(1985);
    expect(jsDate.getMonth()).toBe(2);
    expect(jsDate.getDate()).toBe(3);
  }));

  it('should deal with dates correctly out of fakeAsync', () => {
    const jsDate = new Date(1985, 2, 3, 12);
    jsDate.setFullYear(1985);

    expect(isNaN(jsDate.getTime())).toBeFalsy();
    expect(jsDate.getFullYear()).toBe(1985);
    expect(jsDate.getMonth()).toBe(2);
    expect(jsDate.getDate()).toBe(3);
  });
});
```

The second test passes, but the first one fails:
```
        Expected 3 to be 2.
	Expected 6 to be 3.
```

JiaLiPassion:
fix #1069.
fix #1071.

- in `fakeAsync`, now `Date` is monkey-patched with `FakeDate`, `FakeDate` should handle parameters in constructor correctly, such as `new Date(1995, 1)`, `new Date(0)`.

- should handle `Date.UTC()` in `fakeAsync` correctly

- should handle `Date.parse()` in `fakeAsync` correctly

@mhevery , sorry this is a regression bug, please review and please check should we cut a new release or not.

### DIFF
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -32,13 +32,14 @@
   const OriginalDate = global.Date;
   class FakeDate {
     constructor() {
-      const d = new OriginalDate();
-      d.setTime(global.Date.now());
-      return d;
-    }
-
-    static UTC() {
-      return OriginalDate.UTC();
+      if (arguments.length === 0) {
+        const d = new OriginalDate();
+        d.setTime(FakeDate.now());
+        return d;
+      } else {
+        const args = Array.prototype.slice.call(arguments);
+        return new OriginalDate(...args);
+      }
     }
 
     static now() {
@@ -48,11 +49,18 @@
       }
       return OriginalDate.now.apply(this, arguments);
     }
-
-    static parse() {
-      return OriginalDate.parse();
-    }
   }
+
+  (FakeDate as any).UTC = OriginalDate.UTC;
+  (FakeDate as any).parse = OriginalDate.parse;
+
+  // keep a reference for zone patched timer function
+  const timers = {
+    setTimeout: global.setTimeout,
+    setInterval: global.setInterval,
+    clearTimeout: global.clearTimeout,
+    clearInterval: global.clearInterval
+  };
 
   class Scheduler {
     // Next scheduler id.
@@ -63,7 +71,7 @@
     // Current simulated time in millis.
     private _currentTime: number = 0;
     // Current real time in millis.
-    private _currentRealTime: number = Date.now();
+    private _currentRealTime: number = OriginalDate.now();
 
     constructor() {}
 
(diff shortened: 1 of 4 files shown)


lib/zone-spec/fake-async-test.ts
```diff
@@ -32,13 +32,30 @@
...
-
-    static UTC() {
-      return OriginalDate.UTC();
+      switch (arguments.length) {
+        case 0:
+          const d = new OriginalDate();
+          d.setTime(FakeDate.now());
+          return d;
+        case 1:
+          return new OriginalDate(arguments[0]);
```
mhevery: did you try `new OriginalDate(...arguments)` That would be nicer than having a big switch statemente.

JiaLiPassion: @mhevery , yes, this is a better way, I will change it , thank you.

JiaLiPassion:
find another `compatible issue`.

- if user use the new version of `zone.js`, and they don't set the flag `fakeAsyncPatchLock` to true.
And they use `fakeAsync` themselves, the behavior may wrong in some cases.

```javascript
describe('', () => {
  beforeEach(jasmine.clock().install);
  afterEach(jasmine.clock().uninstall);
});

it('user do fakeAsync themselves', fakeAsync(() => {
   // 1. user use jasmine.clock() 
  //  2. and don't use clock patch
  //  3. and user use fakeAsync themselves
 setTimeout(spy, 100);
 expect(spy).not.toHaveBeenCalled();
 jasmine.clock().tick(100);
 expect(spy).toHaveBeenCalled();
}));
```
This case will `fail` because `jasmine.clock()` will also patch `setTimeout`, so `setTimeout` will not trigger `onScheduleTask`.

- don't patch `jasmine.clock()` method multiple times.

- can handle `jasmine.clock().mockDate()` with empty parameters.

- add more test cases.

PR #1076, opened 2018-04-16
Perf(test): add benchmark page

JiaLiPassion:
add benchmark page to measure the performance gap between `zone.js` patched and `native` API.
this is still working in progress.

the result looks like this.


mhevery:
It says WIP, is this ready to merge?

JiaLiPassion:
@mhevery , I will add tests for all supported APIs, and when I finish, I will let you know.

PR #1093, opened 2018-05-22
feat(Core): fix #910, add a flag to allow user to check duplicate Zone error.

### LINKED ISSUE #910: zone.js loaded by twice, by two separate sources
I built a JS widget using angular 2 that is embedded on other websites. One of the sites that the widget is embedded on has another script that loads zone.js. So when my script loads I get the error: 
```
Error: Zone already loaded. at new g...
```
Is there a way to configure angular-cli (1.0.0-rc.0)/webpack to not load zone.js if it has previously been loaded?

JiaLiPassion:
fix #910.

add a flag, `__zone_symbol__forceDuplicateZoneCheck` to let user choose whether to throw `error` when load `duplicate` zone.js.
By default, if the `global['Zone']` is from `zone.js`, no error will be thrown, and if the `flag` is set to `trure`, then we will throw the error.

sometimes user may not want this `Zone already loaded` error.
For example,
we have two web pages, page1 includes `zone.js`, page2 doesn't.
and the 1st time user load page1 and page2, everything work fine,
but when user load page2 again, error occurs because global['Zone'] already exists.
so we add a flag to let user choose whether to throw this error or not.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -643,7 +643,21 @@ const Zone: ZoneType = (function(global: any) {
   }
   mark('Zone');
   if (global['Zone']) {
-    throw new Error('Zone already loaded.');
+    // if global['Zone'] already exists (maybe zone.js was already loaded or
+    // some other lib also registered a global object named Zone), we may need
+    // to throw an error, but sometimes user may not want this error.
+    // For example,
+    // we have two web pages, page1 includes zone.js, page2 doesn't.
+    // and the 1st time user load page1 and page2, everything work fine,
+    // but when user load page2 again, error occurs because global['Zone'] already exists.
+    // so we add a flag to let user choose whether to throw this error or not.
+    // By default, if existing Zone is from zone.js, we will not throw the error.
+    if (global[('__zone_symbol__forceDuplicateZoneCheck')] === true ||
+        typeof global['Zone'].__symbol__ !== 'function') {
+      throw new Error('Zone already loaded.');
+    } else {
+      return global['Zone'];
+    }
   }
 
   class Zone implements AmbientZone {


chandramuralis:
@JiaLiPassion  Any idea when this fix will be released?

RichardZandi:
@JiaLiPassion there is a similar issue, i believe, with importing native-shim in polyfills when an Angular Element tries to consume another Angular Element. It'd be great if the forthcoming zone fix would address that issue as well. Thank you.

JiaLiPassion:
@RichardZandi, sure, thank you for the information.
@chandramuralis, I am not sure when will release, please wait for a while...

JiaLiPassion:
@benoitjchevalier, I am not sure about the release date, you can use this branch for test by updating your `package.json`.

```
"zone.js": "git://github.com/JiaLiPassion/zone.js#duplicate-patch-dist"
```

benoitjchevalier:
@JiaLiPassion thanks for the tips, we nicked your fix from a zip file you posted in another issue and added it to our project but pointing to the branch will be cleaner. It solves our problem 👍 . 

However we're planning on distributing a packaged js file with a bunch of angular element, so we don't have control over whether consuming angular application will have this fix, hence eagerly waiting for it :)

Thank you for your work!

aaaa0441:
I need to subscribe myself to this issue, waiting for a release of the fix.

liorwohl:
@JiaLiPassion solutions works for me. how is angular 7.0.0 released without this?

hyahyaoui:
> @JiaLiPassion thanks for the tips, we nicked your fix from a zip file you posted in another issue and added it to our project but pointing to the branch will be cleaner. It solves our problem 👍 .
> 
> However we're planning on distributing a packaged js file with a bunch of angular element, so we don't have control over whether consuming angular application will have this fix, hence eagerly waiting for it :)
> 
> Thank you for your work!

@benoitjchevalier do you have a link for the zip please ? thanks

benoitjchevalier:
@hyahyaoui I ended up forking and manually integrating one of the PR that hasn't been merged yet and rebuilt it: `"zone.js":  "https://github.com/benoitjchevalier/zone.js.git#duplicate"`
@JiaLiPassion also released another dist on a branch of his, probably best to use that (don't think that includes the duplicate check  PR if you need it): `"zone.js": "git://github.com/JiaLiPassion/zone.js#release-20181029"`

nakshay:
is this fix released, I need to fix similar issue.

JiaLiPassion:
@nakshay, yes, this is released, you can check it with `zone.js 0.9.0`.

PR #1099, opened 2018-06-11
Ensure FakeAsyncTestZoneSpec tick always doTick

mitchellwills:
In the case where there is pending work in the scheduler queue, but the duration
of the tick did not causes it to run the doTick callback would not be called (or
would not be called with intervals summing to the total time ellapsed).

### DIFF
--- a/lib/zone-spec/fake-async-test.ts
+++ b/lib/zone-spec/fake-async-test.ts
@@ -150,7 +150,11 @@ class Scheduler {
         }
       }
     }
+    lastCurrentTime = this._currentTime;
     this._currentTime = finalTime;
+    if (doTick) {
+      doTick(this._currentTime - lastCurrentTime);
+    }
   }
 
   flush(limit = 20, flushPeriodic = false, doTick?: (elapsed: number) => void): number {
--- a/test/zone-spec/fake-async-test.spec.ts
+++ b/test/zone-spec/fake-async-test.spec.ts
@@ -172,6 +172,25 @@ describe('FakeAsyncTestZoneSpec', () => {
       });
     });
 
+    it('should run doTick callback even if no work ran', () => {
+      fakeAsyncTestZone.run(() => {
+        let totalElapsed = 0;
+        function doTick(elapsed: number) {
+          totalElapsed += elapsed;
+        }
+        setTimeout(() => {}, 10);
+
+        testZoneSpec.tick(6, doTick);
+        expect(totalElapsed).toEqual(6);
+
+        testZoneSpec.tick(6, doTick);
+        expect(totalElapsed).toEqual(12);
+
+        testZoneSpec.tick(6, doTick);
+        expect(totalElapsed).toEqual(18);
+      });
+    });
+
     it('should run queued timer created by timer callback', () => {
       fakeAsyncTestZone.run(() => {
         let counter = 0;


lib/zone-spec/fake-async-test.ts
```diff
@@ -150,7 +150,11 @@
           }
         }
       }
+      lastCurrentTime = this._currentTime;
       this._currentTime = finalTime;
+      if (doTick) {
```
JiaLiPassion: If the code is added here, the `doTick` may be triggered twice.
in your test case, when `setTimeout` callback is triggered, the `doTick` will be called twice.

I think we need a `isTicked` flag to check whether doTick is called.

mitchellwills: `doTick` will already be called with 0 today when there are multiple functions scheduled for execution at the same end time since `doTick` will always be called every time a scheduled function is invoked. Generally it's very unclear what the behavior of this callback should be since I couldn't find any existing tests for it.

I'm also not sure what `queueConsumed` in your snipit is for. Regardless of what scheduled work ran or not this function should always make sure it ticks any remaining time. And if you believe that `doTick` should not be invoked with 0 then the `this._currentTime === lastCurrentTime` check should be enough.

JiaLiPassion: > doTick will already be called with 0 today when there are multiple functions scheduled for execution at the same end time since doTick will always be called every time a scheduled function is invoked.
This is a little different, because in this case, doTick is triggered because the scheduled function is invoked.

`queueConsumed` just means if the `doTick` is triggered once by the scheduled function, and the finalTime equals to currentTime, we should not `doTick` again, but if `doTick` is not triggered by schedule function, even finalTime equals to currentTime, we still need to call `doTick`.

> Generally it's very unclear what the behavior of this callback should be since I couldn't find any existing tests for it.
Yeah, this is old code, I am not sure when we should call `doTick` either, I will clarify the requirement and let's discuss again.

JiaLiPassion: @mitchellwills 
The `doTick` callback is from here. https://github.com/angular/zone.js/pull/866
The purpose is to `advance Jasmine.mockDate`, and in current newest `zone.js`, `jasmine.mockDate` has been advanced automatically, so we don't need to do that already.

So I think you may have another use case to use `doTick`, could you share your use case?

And I think the requirement should be, 
`doTick should be called when fake timer advanced.`

So 

1. multiple scheduled macroTasks with the same end time should not trigger multiple `doTick`.
2. `0` should be triggered, but still only once. 

The reason that we should only `doTick` once, because consider the following case.

```ts
setTimeout(function() {
  setTimeout(function() {
  }, 0);
}, 0);
```

if we call `tick(0, (millis) => console.log)`

the result should be
```
0 // triggered by the outer timeout
0 // triggered by the inner timeout
```

What do you think?

JiaLiPassion:
@vikerman , could you help to review this one (especially about when to trigger `doTick` and whether we should trigger `doTick` multiple times when multiple tasks have the same end time), Thank you!

lib/zone-spec/fake-async-test.ts
mitchellwills: My use case is also advancing the clock when integrating with some tooling we have.

My view on this was that either there are very clear semantics on when doTick fires or it should be considered an implementation detail in which case it doesn't matter if doTick is called as long as the durations add up to the correct amount and it's best to just keep the code as simple as possible. Are there any other use cases aside from advancing the clock.

A few options:
 * Fires after each macro task
 * Only called with a non-zero value
 * Once for each macro task time advance and then the duration from the last macro task to the end time, possibly 0 (I think this is what you suggested)
 * It's an implementation detail and the only guarantee is that it will be called to increment time the correct amount

Some other edge cases
```
setTimeout(function() {}, 10);
setTimeout(function() {}, 10);
tick(10, millis => console.log);
// should this print 10 or 10, 0

setTimeout(function() {}, 0);
setTimeout(function() {}, 0);
tick(10, millis => console.log);
// should this print 0, 10 or 10
```

My personal opinion would be whatever leads to the simplest code unless there are specific semantics that would be advantageous.

JiaLiPassion: > Only called with a non-zero value

I am not sure about this one, I think `doTick` should fire when `setTimeout(() => {}, 0); tick()`.
But if you mean `Fires after each macro task` is a stronger rule, I think it will be ok.
About the sample code you provided,

```ts
setTimeout(function() {}, 10);
setTimeout(function() {}, 10);
tick(10, millis => console.log);
// should this print 10 or 10, 0
// I prefer 10 here.

setTimeout(function() {}, 0);
setTimeout(function() {}, 0);
tick(10, millis => console.log);
// should this print 0, 10 or 10
// I prefer 0, 10 here
```

So the rule will be 
1. macroTask will fire `doTick` regardless of the duration is zero or not.
2. If not match the rule 1, no macroTasks triggered, then we should also trigger `doTick` regardless of the duration is zero or not.
3. If rule 1 is matched, some macroTask trigger `doTick` already, and after the trigger, the last triggered macroTasks' endTime equals finalTime, we should not `doTick` again with a zero.

```ts
// rule 1
setTimeout(function() {}, 0);
tick(0, (millis) => console.log);
// should print 0

// rule 2 and rule3
setTimeout(function(){}, 10);
tick(0, (millis) => console.log);
// should print 0

tick(5, (millis) => console.log);
// should print 5

tick(10, (millis) => console.log);
// should print 10 (not 10, 0)

```

what do you think?

mitchellwills: I think that would be fine; however, I haven't seen any justification for the additional complexity.
If there comes a time where having a specific behavior makes sense then it could be added, but until then is there any reason to overconstrain the behavior (more complex to maintain, harder to change in the future, etc)? As long at the function fulfills advancing a clock the correct amount of time it doesn't seem like there is any need to define the behavior further than that.

mitchellwills: SGTM
@vikerman I did run a global presubmit with the google code base. There were only one failure, but I don't think they were because of the number of calls to doTick.

JiaLiPassion:
@mitchellwills, some other PR was merged, could you rebase this one? thanks.

mitchellwills:
Rebased.

PR #1109, opened 2018-07-03
fix(core): fix #1108, window.onerror should have (message, source, ....) signiture

### LINKED ISSUE #1108: window.onerror signature is incorrect
The `window.onerror` handler has [a different method signature](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror#window.onerror) compared to all other `*.on*` event methods. When ZoneJS wraps `window.onerror` it changes the signature to be the standard handler signature instead of the `window.onerror` specific signature.

http://embed.plnkr.co/jiclV3raUVIhlz3uQ3IM/

JiaLiPassion:
Fix #1108,
Based on the document here, https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror#window.onerror

The signature  of `window.onerror`  should be 

```ts
window.onerror = function(message, source, lineno, colno, error) { ... }
```

And `window.addEventListener('error', listener)` should be 

```ts
window.addEventListener('error', function(event) { ... })
```

So we must handle the `onerror` special case.

### DIFF
--- a/file-size-limit.json
+++ b/file-size-limit.json
@@ -3,7 +3,7 @@
     {
       "path": "dist/zone.min.js",
       "checkTarget": true,
-      "limit": 41600
+      "limit": 42000
     }
   ]
 }
--- a/lib/browser/property-descriptor.ts
+++ b/lib/browser/property-descriptor.ts
@@ -10,7 +10,7 @@
  * @suppress {globalThis}
  */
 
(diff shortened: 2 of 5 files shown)


pamo:
Re-iterating what others have already asked: any ideas around when this will be released?

PR #1114, opened 2018-07-18
fix(bluebird): fix #1112, bluebird chained callback should return a Bluebird Promise

### LINKED ISSUE #1112: Patched bluebird does not provide data to promise chain
Just started investigating the upgrade of a large application based on AngularJS to Angular. This application relies on Bluebird as its global promise object and uses webpack as its module loader. The prescribed solutions mentioned in #455 were unsuccessful, as the promises from webpack kept resolving to `undefined` instead of the relevant view and controller content.

```typescript
...
import './polyfills';

import * as Bluebird from 'bluebird';

import 'zone.js';
import 'zone.js/dist/zone-bluebird';

... 

// Patch bluebird to be zone aware. This also sets the global Promise object
// to be bluebird.
// Hackishly cast Zone so we can index it by the symbol for "bluebird" as
// patched by zone-bluebird.
(Zone as any)[Zone.__symbol__('bluebird')](Bluebird);

// Disable promise checking in zone.js because it doesn't recognize Bluebird as 
// zone aware, even when patched.
Zone.assertZonePatched = () => {};

...
```

After investigation the reason for these failures seem to be the way the promises are resolved in the bluebird patch function. The callbacks for `.then`, `.spread`, and `.finally` are patched to schedule the invocation as a zone.js micro task. However, the result from the callbacks is never made available to bluebird breaking the promise chain. Specifically, the call to `zone.scheduleMicroTask` returns a task, but that task is never saved, and the result of the task invocation is not passed back to Bluebird.

To solve this I have patched the Bluebird patch to return promises from the callbacks, ensuring Bluebird waits for the values to resolve when the micro task is invoked, and allowing the values to continue into the promise chain.

```diff
diff --git a/web/webpack/node_modules/zone.js/lib/extra/bluebird.ts b/web/webpack/node_modules/zone.js/lib/extra/bluebird.ts
index 13a0496eaa..0e8ef3b358 100644
--- a/web/webpack/node_modules/zone.js/lib/extra/bluebird.ts
+++ b/web/webpack/node_modules/zone.js/lib/extra/bluebird.ts
@@ -22,11 +22,21 @@ Zone.__load_patch('bluebird', (global: any, Zone: ZoneType, api: _ZonePrivate) =
           const func = args[i];
           if (typeof func === 'function') {
             args[i] = function() {
+              let callbackResolve, callbackReject;
+              const promise = new Bluebird((resolve, reject) => {
+                callbackResolve = resolve;
+                callbackReject = reject;
+              });
               const argSelf: any = this;
               const argArgs: any = arguments;
               zone.scheduleMicroTask('Promise.then', () => {
-                return func.apply(argSelf, argArgs);
+                try {
+                  callbackResolve(func.apply(argSelf, argArgs));
+                } catch (e) {
+                  callbackReject(e);
+                }
               });
+              return promise;
             };
           }
         }
```

JiaLiPassion:
fix #1112.
The issue can be described as the following case.

```ts
new BluebirdPromise((resolve: any, reject: any) => {
      expect(Zone.current.name).toEqual('zone_A');
      resolve(1);
}).then((r: any) => {
      expect(r).toBe(1);
      expect(Zone.current.name).toEqual('zone_A');
      return Promise.resolve(2);
}).then((r: any) => {
      expect(r).toBe(2);
      expect(Zone.current.name).toEqual('zone_A');
});
```

So we should return a `new bluebird promise` in `then callback`.

### DIFF
--- a/file-size-limit.json
+++ b/file-size-limit.json
@@ -3,7 +3,7 @@
     {
       "path": "dist/zone.min.js",
       "checkTarget": true,
-      "limit": 42000
+      "limit": 42050
     }
   ]
 }
--- a/lib/common/promise.ts
+++ b/lib/common/promise.ts
@@ -476,11 +476,6 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
 
   if (NativePromise) {
     patchThen(NativePromise);
-
-    /*let fetch = global['fetch'];
-    if (typeof fetch == 'function') {
-      global['fetch'] = zoneify(fetch);
-    }*/
   }
 
   // This is not part of public API, but it is useful for tests, so we expose it.
--- a/lib/extra/bluebird.ts
+++ b/lib/extra/bluebird.ts
@@ -25,14 +25,30 @@ Zone.__load_patch('bluebird', (global: any, Zone: ZoneType, api: _ZonePrivate) =
                 args[i] = function() {
                   const argSelf: any = this;
                   const argArgs: any = arguments;
-                  zone.scheduleMicroTask('Promise.then', () => {
-                    return func.apply(argSelf, argArgs);
+                  return new Bluebird((res: any, rej: any) => {
+                    zone.scheduleMicroTask('Promise.then', () => {
+                      try {
+                        res(func.apply(argSelf, argArgs));
+                      } catch (error) {
+                        rej(error);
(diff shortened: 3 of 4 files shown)


jrabbe:
Changes LGTM, would be great to have this merged and released as soon as convenient, @mhevery. Thank you.

PR #1133, opened 2018-09-09
feat(custom-element): patch customElement v1 APIs

JiaLiPassion:
Patch the `CustomElement V1` APIs.

```
connectedCallback
disconnectedCallback
adoptedCallback
attributeChangedCallback
```
Make them run in the `zone` when they were `defined by customElements.define`.

So the issue like https://github.com/angular/angular/pull/24185 may be resolved by this one.

@robwormald, @mhevery, please review, thank you!


remackgeek:
This is cool!  Does it play nice with the document-register-element polyfill (specifically 1.8.1)?

JiaLiPassion:
@remackgeek, I believe before this PR, `zone.js` already support `legacy document.registerElement`. So `document-register-element` polyfill will also be supported.

krishna217:
Which version will this issue be fixed ?

antch:
Just FYI for anyone following the thread, this fix is available as of 0.8.27.

harveyelsisura:
@antch where?

PR #1140, opened 2018-09-24
feat(async): support native async/await

JiaLiPassion:
I think I may find a way to resolve `native async/await` issue.
The test code below works well.
```ts
  async function test() {
            console.log('test');
            return 1;
        }

        async function test1() {
            console.log('before await test', Zone.current.name);
            const result = await test();
            // should keep zone context test
            console.log('after await test', result, Zone.current.name);
        }

        async function test2() {
            console.log('before await test1', Zone.current.name);
            const result = await test1();
            // should keep zone context test
            console.log('after await test1', result, Zone.current.name);
        }
        console.log('begin test');
        Zone.current.fork({name: 'test'}).run(test1);
        // here should be <root> zone.
        console.log('outside test', Zone.current.name);
        Zone.current.fork({name: 'test1'}).run(() => {
            // should be test1 zone.
            console.log('another test', Zone.current.name);
        });
```

@mhevery, I will continue to add test cases, hope it can work.

### DIFF
--- a/lib/common/promise.ts
+++ b/lib/common/promise.ts
@@ -207,6 +207,9 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
             api.scheduleMicroTask();  // to make sure that it is running
           }
         }
+        if ((promise as any)['__zone_symbol__isAsync'] === true) {
+          Zone.setAsyncFrame();
+        }
       }
     }
     // Resolving an already resolved promise is a noop.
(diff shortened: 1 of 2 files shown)


KrzysztofKarol:
Hi @JiaLiPassion, do you have any update on this?

jguyenot:
Hi @JiaLiPassion have you time to correct integrations fails?

JiaLiPassion:
@KrzysztofKarol , @jguyenot , there are some cases this solution will not work, I will continue to find another solution, thanks!

jguyenot:
Thanks for your time. It's actually impossible to implement lazy loading module with angular 8 and zone.js so it's very important.
Good luck.

PR #1150, opened 2018-10-23
Fix ZoneAwarePromise.all to resolve at the correct time

### LINKED ISSUE #1077: Promise.all indexing issue when combining non-zone.js promises
When you use [Promise.all](https://github.com/angular/zone.js/blob/c8c599073b87025d44a4b7d6e2efea9cedfbd3e2/lib/common/promise.ts#L305-L331) for combining not only zone.js promises (e.g. WinJSPromise) you can fall into issue that Promise.all results into single element result.

Example of issue https://plnkr.co/edit/UO2QhIa5eeLm6tPAyAjt?p=preview. 
Important part of the code(skipping WinJSPromise import) is this one:
```
  let p1 = WinJSPromise.as(3);
  let p2 = Promise.resolve(45);
  Promise.all([p1, p2]).then(res => {
    console.log(`Promise.all: `, res);
  })
  WinJSPromise.join([p1, p2]).then(res => {
    console.log(`WinJSPromise.join: `, res);
  });
```
The output from Promise.all is [45], where expected is [3, 45]. **This works fine with native Promise.all**.

Problem is in how index is counted in this implementation:
https://github.com/angular/zone.js/blob/c8c599073b87025d44a4b7d6e2efea9cedfbd3e2/lib/common/promise.ts#L305-L331

Following update implementation fixes the issue (note the index variable missing in original version):
```
Promise.all = function (values: any): Promise<any> {
    let resolve: (v: any) => void;
    let reject: (v: any) => void;
    let promise = new this((res, rej) => {
      resolve = res;
      reject = rej;
    });
    let count = 0;
    let index = 0;
    const resolvedValues: any[] = [];
    for (let value of values) {
      if (!(value && value.then)) {
        value = this.resolve(value);
      }
      value.then(
          ((index) => (value: any) => {
            resolvedValues[index] = value;
            count--;
            if (!count) {
                resolve(resolvedValues);
            }
          })(index),
          reject);
      count++;
      index++;
    }
    if (!count) resolve(resolvedValues);
    return promise;
}
```

This issue was found and discussed a bit [here](https://github.com/Microsoft/monaco-editor/issues/790)

trevorade:
For ZoneAwarePromise.all, as [implemented](https://github.com/angular/zone.js/blob/7201d44451f6c67eac2b86eca3cbfafde14035d6/lib/common/promise.ts), the `count` variable serves three purposes:

1.  Count the number of values passed-in
2.  Specify the index of a resolved value in `resolvedValues`
3.  Track when all the promises have been resolved.

In the event that `value.then` is immediately called, `count` will be decremented at the incorrect time resulting in a potentially negative value or reaching 0 when not all values have actually been resolved. 

The fix is to be more meticulous about using the correct indices at the correct time and to not overload the count and number of resolved promises.  This updated version is more explicit about the purpose of the `unresolvedCount` and `valueIndex` variables.  `unresolvedCount` needs to start at 2 to prevent `resolve` from being called too soon in the event that `value.then` calls the callback immediately. The scoping of the index for use in `resolvedValues` is made constant to guarantee the correct index is used.

This buggy behavior specifically manifested as an issue in the Monaco editor but most likely occurs elsewhere as well in cases where promises may immediately call into `.then`.

Related issue:
https://github.com/Microsoft/monaco-editor/issues/790#issuecomment-378452532

This fixes #1077

### DIFF
--- a/lib/common/promise.ts
+++ b/lib/common/promise.ts
@@ -315,24 +315,37 @@ Zone.__load_patch('ZoneAwarePromise', (global: any, Zone: ZoneType, api: _ZonePr
         resolve = res;
         reject = rej;
       });
-      let count = 0;
+
+      // Start at 2 to prevent prematurely resolving if .then is called immediately.
+      let unresolvedCount = 2;
+      let valueIndex = 0;
+
       const resolvedValues: any[] = [];
       for (let value of values) {
         if (!isThenable(value)) {
           value = this.resolve(value);
         }
-        value.then(
-            ((index) => (value: any) => {
-              resolvedValues[index] = value;
-              count--;
-              if (!count) {
-                resolve(resolvedValues);
-              }
-            })(count),
-            reject!);
-        count++;
+
+        const curValueIndex = valueIndex;
+        value.then((value: any) => {
+          resolvedValues[curValueIndex] = value;
+          unresolvedCount--;
+          if (unresolvedCount === 0) {
+            resolve!(resolvedValues);
+          }
+        }, reject!);
+
+        unresolvedCount++;
+        valueIndex++;
       }
-      if (!count) resolve!(resolvedValues);
+
+      // Make the unresolvedCount zero-based again.
+      unresolvedCount -= 2;
+
+      if (unresolvedCount === 0) {
+        resolve!(resolvedValues);
+      }
+
       return promise;
     }
 


evmar:
cc @rkirov

evmar:
Maybe mention this fixes https://github.com/angular/zone.js/issues/1077

trevorade:
The continuous-integration/travis-ci/pr failure appears to be a disconnect.  I didn't see any failing tests.  I don't know how to re-run those tests.

JiaLiPassion:
I remembered some other guy get  the same issue before, I think this is the bug of `Monaco editor`.
https://github.com/angular/zone.js/issues/1077

evmar:
Igor says that @mhevery is the right person to review this

evmar:
Also at a higher level, why do we need to reimplement these?  Why can't we have our monkeypatches reuse the builtins?

JiaLiPassion:
should we add some test cases about this? I can do that in another PR.

JiaLiPassion:
@evmar, yeah, I think if we only need to keep the `zone` information, we can just `monkey-patch` the `constructor` and `then`. I believe the main reason is we want to have our own `microTaskQueue`, so we can run the `microTask` inside this `queue` after other `macroTask(setTimeout)` is invoked, is that correct, @mhevery?

trevorade:
@mhevery, this appears ready to go.  Not sure what the next step is for this.

PR #1159, opened 2018-11-04
WIP(core): Proposal: can unload, reload patch modules

JiaLiPassion:
The same idea with #1073.

In 1073, there is some performance concern, I used a more simple logic to implement this in this PR, every time only need to do a `small loop` and just `reset reference`.. I did some basic benchmark.
It will only cost `0.000022ms` every time `unload/reload` all the patches, so I think basically `no cost`


JiaLiPassion:
The idea of #1073 is,
This is a proposal to support `unload`, `reload` patch.

### Use case: only run specified code into zone.

```javascript
const zone = Zone.current.fork({name: 'zone'});
zone.run(() => {setTimeout(() => {
  console.log('zone', Zone.current.name);  // will run into zone
})});

setTimeout(() => {
  console.log('zone', Zone.current); // will run into root zone in current version.
}, 100);
```

for example, in current version, the code above, the 1st setTimeout will run into `zone`, the 2nd will run into `root Zone`. But even `root` Zone still have performance impact. So we may want the 2nd totally run in `native setTimeout`. 

in this PR proposal, only `patch` all `modules` such as `setTimeout, XHR, promise` in `zone.run`.
and `unPatch` them when `in root zone` or by calling some method like 'runOutsideOfZone` explicitly.

### Use case: such as angular elements to implement WebComponent. Only run logic with `zone.js` patched inside `angular elements without impact outside world. 

### Proposal. Auto patch when `zone.run`, Auto unpatch after `zone.run`. Auto repatch when `zone.run` again.

```javascript
const zone = Zone.current.fork({name: 'zone'});
zone.run(() => {setTimeout(() => {
  console.log('zone', Zone.current.name);  // will run into zone
})});

setTimeout(() => {
  console.log('zone', Zone.current); // will not in root zone, this is native setTimeout
}, 100);

zone.run(() => {setTimeout(() => {
  console.log('zone', Zone.current.name);  // will run into zone again by rePatch
})});
```

To make `unPatch` and `rePatch` faster, keep `patched delegate` and `native delegate` reference when call `Zone.__load_patch`, so we only reset `delegate` when `unPatch` and `rePatch`.

mleibman:
Can this behavior be configurable?
Just like in https://github.com/angular/zone.js/pull/1138#issuecomment-428585494, this would break some of our existing latency instrumentation that relies on monkey-patching Zone.runTask/Zone.scheduleTask/Zone.cancelTask to be able to track all tasks including the ones run in the root zone.

JiaLiPassion:
@mleibman, yes, this `new behavior` is totally configurable, by default, `zone.js` will work just like before. This new behavior supposes to just work for `Angular Elements`.

Issue #341, opened 2016-05-01
ZoneTask.toString() method missing


char0n:
I have reported this issue to AngularJs team: https://github.com/angular/angular.js/issues/14548

Basically their are right in their assumption that using zone.js should not change setTimeout. Imagine this situation when using zone.js.

```javascript
var deferredTracker = {}

var timerId1 = setTimeout(function() {});
var timerId2 = setTimeout(function() {});

deferredTracker[timerId1] = new Promise(....);
defferedTracker[timerId2] = new Promise(....);
```

The result of defferedTracker is:

```javascript
{"[object Object]": promiseInstance}
```

This is the way angular1 tracks timeouts and interval. Using zone.js allows to cancel only the last timeout or interval.

To fix this issue I would suggest for ZoneTask to have toString method which for setTimeout and setInterval should look like this:

```javascript
toString: function() {
  return this.data.handleId;
} 
```

Without fixing this, using zone.js in angular1 application is just not possible.

Thank you

char0n:
I have created modified distribution file here: http://filebin.ca/2hRxo13INo5U/zone.js-0.6.12.tgz

Affected files are:
- dist/zone.js
- dist/zone-node.js

Both file have toString method implemented. Using this file is the only way to use zone.js along with angular 1.5 side by side.

mhevery:
@char0n Nice catch. Could you create PR with a fix?

char0n:
@mhevery I'll try

char0n:
I am having problem installing the project. I run on Windows 8.1. In order to install the project I had to modify the package.json:

```javascript
"prepublish": "./node_modules/.bin/typings install && ./node_modules/.bin/tsc && gulp build"
```
to 
```javascript
"prepublish": "./node_modules/.bin/typings install && call ./node_modules/.bin/tsc && gulp build"
```

Then when I run npm install I get this error:

```
> zone.js@0.6.12 prepublish C:\Users\char0n\Documents\GitHub\zone.js
> typings install && call ./node_modules/.bin/tsc && gulp build


├── es6-promise (ambient)
├── node (ambient)
└── jasmine (ambient dev)

[10:30:27] Using gulpfile ~\Documents\GitHub\zone.js\gulpfile.js
[10:30:27] Starting 'build/zone.js'...
[10:30:28] Starting 'compile'...
[10:30:28] Starting 'build/zone.min.js'...
[10:30:28] Starting 'build/zone-node.js'...
[...]

PR #1166, opened 2018-11-21
fix(core): fix #1153, ZoneTask.toString should always be a string

### LINKED ISSUE #1153: ZoneTask.toString can return a number, which causes problems in Jasmine
If a karma/jasmine test fails due to an unhandled promise rejection, Jasmine's `PrettyPrinter` class throws an exception due to the `ZoneTask.toString` function not returning a string.

See: https://github.com/jasmine/jasmine/issues/1575

JiaLiPassion:
fix #1153.

### DIFF
--- a/lib/zone.ts
+++ b/lib/zone.ts
@@ -1240,7 +1240,7 @@ const Zone: ZoneType = (function(global: any) {
 
     public toString() {
       if (this.data && typeof this.data.handleId !== 'undefined') {
-        return this.data.handleId;
+        return this.data.handleId.toString();
       } else {
         return Object.prototype.toString.call(this);
       }
--- a/test/common/toString.spec.ts
+++ b/test/common/toString.spec.ts
@@ -47,3 +47,43 @@ describe('global function patch', () => {
        }));
   });
 });
+
+describe('ZoneTask', () => {
+  it('should return handleId.toString if handleId is available', () => {
+    let macroTask1: any = undefined;
+    let macroTask2: any = undefined;
+    let microTask: any = undefined;
+    const zone = Zone.current.fork({
+      name: 'timer',
+      onScheduleTask: (delegate: ZoneDelegate, curr: Zone, target: Zone, task: Task) => {
+        if (task.type === 'macroTask') {
+          if (!macroTask1) {
+            macroTask1 = task;
+          } else {
+            macroTask2 = task;
+          }
+        } else if (task.type === 'microTask') {
+          microTask = task;
+        }
+        return task;
+      }
+    });
+    zone.run(() => {
+      const id1 = setTimeout(() => {});
+      clearTimeout(id1);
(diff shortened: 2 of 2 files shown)


lib/zone.ts
```diff
@@ -1240,7 +1240,7 @@ const Zone: ZoneType = (function(global: any) {
 
     public toString() {
       if (this.data && typeof this.data.handleId !== 'undefined') {
-        return this.data.handleId;
+        return Object.prototype.toString.call(this.data.handleId);
```
gkalpak: I don't think this will have the expected behavior, as it will always return `[object Number]`  (which would cause #341 again).
I think it should be `this.data.handleId.toString()` here.

test/common/toString.spec.ts
```diff
@@ -47,3 +47,29 @@ describe('global function patch', () => {
...
+      }
+    });
+    zone.run(() => {
+      const id = setTimeout(() => {});
+      clearTimeout(id);
+      Promise.resolve().then(() => {});
+      expect(typeof macroTask.toString()).toEqual('string');
+      expect(macroTask.toString()).toEqual(id.toString());
+      expect(typeof microTask.toString()).toEqual('string');
+    });
```
gkalpak: I think it would be valueable to also verify that two different macrotasks have different string representations (just in case).

lib/zone.ts
JiaLiPassion: Yeah, you are right, thanks!

test/common/toString.spec.ts
JiaLiPassion: Got it, case added.

JiaLiPassion:
@gkalpak, thanks for the review, I didn't implement the `valueOf` because besides `setTimeout/interval`, other types of  task (microTask, eventTask) may not have some `valid primitive representations`.

test/common/toString.spec.ts
```diff
@@ -47,3 +47,43 @@ describe('global function patch', () => {
...
+      const id2 = setTimeout(() => {});
+      clearTimeout(id2);
+      Promise.resolve().then(() => {});
+      const macroTask1Str = macroTask1.toString();
+      const macroTask2Str = macroTask2.toString();
+      expect(typeof macroTask1Str).toEqual('string');
+      expect(macroTask1Str).toEqual(id1.toString());
+      expect(typeof macroTask2Str).toEqual('string');
+      expect(macroTask2Str).toEqual(id2.toString());
+      if (macroTask1.data && typeof macroTask1.data.handleId === 'number') {
```
gkalpak: Why is this check necessary?

JiaLiPassion: Yeah, the reason is this case will also run in `nodejs` environment, in `nodejs` the `handleId` will be a `TimerObject`, in that case, we will not check their equality.

PR #1179, opened 2019-01-19
fix(element): Fix native_shim.js not implement adoptedCallback error.

JiaLiPassion:
To pass test in `angular`, we need to handle some of `callbacks` are not implemented by some `polyfilles`. `@webcomponent/native_shim.js` doesn't implement `adopdatedCallback`, `zone.js` will throw error, so we need to check this case.

### DIFF
--- a/lib/browser/register-element.ts
+++ b/lib/browser/register-element.ts
@@ -26,7 +26,7 @@ function patchCallbacks(target: any, targetName: string, method: string, callbac
           if (descriptor && descriptor.value) {
             descriptor.value = wrapWithCurrentZone(descriptor.value, source);
             _redefineProperty(opts.prototype, callback, descriptor);
-          } else {
+          } else if (prototype[callback]) {
             prototype[callback] = wrapWithCurrentZone(prototype[callback], source);
           }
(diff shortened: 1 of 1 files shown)


lib/browser/register-element.ts
```diff
@@ -26,7 +26,7 @@ function patchCallbacks(target: any, targetName: string, method: string, callbac
           if (descriptor && descriptor.value) {
             descriptor.value = wrapWithCurrentZone(descriptor.value, source);
             _redefineProperty(opts.prototype, callback, descriptor);
-          } else {
+          } else if (prototype[callback]) {
```
gkalpak: There is already a check for `prototype.hasOwnProperty(callback)` a few lines above.
Does it mean that the properties exists, but are falsy? (That would be unexpected.)

JiaLiPassion:
Yes, hasOwnProperty return true, but prototype[call back] is undefined.

lib/browser/register-element.ts
JiaLiPassion: Yes, hasOwnProperty return true, but prototype[call back] is undefined.

PR #1191, opened 2019-02-24
build: build zone-evergreen.js in es2015, add terser minify support

JiaLiPassion:
1. build `zone-evergreen.js` in `es2015` format.
2. use `terser` to minify `zone-evergreen.js`.
Here is the `terser options` I used.
```
 compress: {
          module: true
        },
        ecma: 5, // specify one of: 5, 6, 7 or 8
        keep_classnames: false,
        keep_fnames: false,
        ie8: false,
        module: false,
        nameCache: null, // or specify a name cache object
        safari10: false,
        toplevel: false,
        warnings: false,
```


.travis.yml
```diff
@@ -55,5 +55,8 @@ script:
   - node_modules/.bin/gulp test/node
   - node_modules/.bin/gulp test/node -no-patch-clock
   - cp ./test/browser/custom-element.spec.js ./build/test/browser
+  - yarn tsc:es6
```
IgorMinar: please rename because es6 is not a thing, es2015 is the official name so let's be consistent :)

gulpfile.js
```diff
@@ -62,7 +66,24 @@ function generateScript(inFile, outFile, minify, callback, format) {
         .pipe(rename(outFile)),
   ];
   if (minify) {
-    parts.push(uglify());
+    if (format === 'es') {
+      parts.push(terser({
```
IgorMinar: @filipesilva is this the right configuration for terser esm2015 build? can you please double check. thanks

karma-build-jasmine.es6.conf.js
```diff
@@ -1,5 +1,11 @@
 
 module.exports = function (config) {
```
IgorMinar: can you rename this file to es2015?

karma-build-jasmine.es6.conf.js
```diff
@@ -1,5 +1,11 @@
...
   require('./karma-build-jasmine.conf.js')(config);
+  for (let i = 0; i < config.files.length; i ++) {
+    if (config.files[i] === 'node_modules/core-js-bundle/index.js') {
+      config.files.splice(i, 1);
+      break;
+    }
+  }
   config.client.entrypoint = 'browser_es6_entry_point';
```
IgorMinar: es2015 or esm2015

karma-build.conf.js
```diff
@@ -8,6 +8,7 @@
 
 module.exports = function(config) {
   require('./karma-base.conf.js')(config);
+  config.files.push('node_modules/core-js-bundle/index.js');
```
IgorMinar: why do we need this polyfill?

package.json
```diff
@@ -37,6 +37,7 @@
     "ws-server": "node ./test/ws-server.js",
     "tsc": "tsc -p .",
     "tsc:w": "tsc -w -p .",
+    "tsc:es6": "tsc -p tsconfig-esm-2015.json",
```
IgorMinar: esm2015

gulpfile.js
```diff
@@ -62,7 +66,24 @@ function generateScript(inFile, outFile, minify, callback, format) {
...
+        compress: {
+          module: true
+        },
+        ecma: 5, // specify one of: 5, 6, 7 or 8
+        keep_classnames: false,
+        keep_fnames: false,
+        ie8: false,
+        module: false,
```
IgorMinar: false? doesn't that mean that this will produce es2015 bundle rather than esm2015? in which case all the "esm2015" references should really be "es2015" referenfes. no?

.travis.yml
JiaLiPassion: Sure, got it!

gulpfile.js
JiaLiPassion: Yeah, I am not sure about this `configuration part`, thanks.

karma-build-jasmine.es6.conf.js
JiaLiPassion: Sure, got it.

karma-build-jasmine.es6.conf.js
JiaLiPassion: Got it!

karma-build.conf.js
JiaLiPassion: Because to compile in `es2015`, `ZoneAwarePromise` must implement
```
    get[Symbol.toStringTag]() {
      return 'Promise' as any;
    }
```

So we need the `core-js` to pass test in `IE`, and in `es2015/evergreen` test, I removed this `polyfill`.

gulpfile.js
JiaLiPassion: I will check it again!

gulpfile.js
filipesilva: @JiaLiPassion, as a reference, the Terser settings we use in Angular CLI for applications can be found [here](https://github.com/angular/angular-cli/blob/58bc5b8f2a2a285f8e09ec05ef407fe32be3ef05/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts#L228-L266).

I advise you to use only `terser` instead of `uglify` for some bundles and `terser` for other. At certain points in time either of these tools suffered from different hard to identify bugs. Ideally your bundles function the same. But if different formats use different minifiers, you may have different behaviour between them.

Under this mindset it is also preferable to keep as much of the minification settings identical between bundles.

Given that your existing es5 bundle uses default `uglify` settings, I recommend your es2015 bundle to use the same defaults and only pass in `{ ecma: 6 }`. This option will propagate to other Terser options automatically. I don't think you need any other options.

gulpfile.js
filipesilva: This option in Terser reads:
```
module (default false) — Use when minifying an ES6 module. "use strict" is implied and names can be mangled on the top scope. If compress or mangle is enabled then the toplevel option will be enabled.
```

Setting this option to `true` is helpful when minifying application code because you are not concerned with exposed public APIs and want the smallest possible self-contained application. But for libraries, you do not want to either mangle or remove code top level code.

So this option should be set to false, or left to default.

Please bear in mind that Terser is not capable of altering the bundle format. Terser options such as `ecma`, `toplevel` and `module` are meant only to enable/disable available optimizations that are applicable only in these cases. But the module format is always left unchanged.

gulpfile.js
JiaLiPassion: @filipesilva, thank you very much for the help, currently, zone.js use `uglify` for `es5 bundle` and use `terser` for `es2015 bundle`, I will use `terser` to handle both `es5` and `es6`.
And I will also delete all other options.

gulpfile.js
JiaLiPassion: @filipesilva, got it, I will remove those options and test again.

JiaLiPassion:
@IgorMinar, @filipesilva, I have updated the `terser`  and CI, and `CI` is green now. please review, thanks!

PR #1192, opened 2019-02-24
env: change BLACK_LISTED_EVENTS to DISABLE_EVENTS

JiaLiPassion:
based on this change in Angular repo, https://github.com/angular/angular/pull/28529
change env name from `__zone_symbol__BLACK_LISTED_EVENTS` to `__zone_symbol__DISABLE_EVENTS`.
And for `backward compatibility`, we still support `__zone_symbol__BLACK_LISTED_EVENTS`

### DIFF
--- a/lib/browser/api-util.ts
+++ b/lib/browser/api-util.ts
@@ -13,4 +13,19 @@ Zone.__load_patch('util', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
   api.patchMethod = patchMethod;
   api.bindArguments = bindArguments;
   api.patchMacroTask = patchMacroTask;
+  // In earlier version of zone.js (<0.9.0), we use env name `__zone_symbol__BLACK_LISTED_EVENTS` to
+  // define which events will not be patched by `Zone.js`.
+  // In newer version (>=0.9.0), we change the env name to `__zone_symbol__UNPATCHED_EVENTS` to keep
+  // the name consistent with angular repo.
+  // The  `__zone_symbol__BLACK_LISTED_EVENTS` is deprecated, but it is still be supported for
+  // backwards compatibility.
+  const SYMBOL_BLACK_LISTED_EVENTS = Zone.__symbol__('BLACK_LISTED_EVENTS');
+  const SYMBOL_UNPATCHED_EVENTS = Zone.__symbol__('UNPATCHED_EVENTS');
+  if (global[SYMBOL_UNPATCHED_EVENTS]) {
+    global[SYMBOL_BLACK_LISTED_EVENTS] = global[SYMBOL_UNPATCHED_EVENTS];
+  }
+  if (global[SYMBOL_BLACK_LISTED_EVENTS]) {
+    (Zone as any)[SYMBOL_BLACK_LISTED_EVENTS] = (Zone as any)[SYMBOL_UNPATCHED_EVENTS] =
+        global[SYMBOL_BLACK_LISTED_EVENTS];
+  }
 });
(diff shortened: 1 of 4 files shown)


JiaLiPassion:
@IgorMinar, this one related to https://github.com/angular/angular/pull/28529, please review, thanks.

lib/browser/api-util.ts
```diff
@@ -13,4 +13,14 @@ Zone.__load_patch('util', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
   api.patchMethod = patchMethod;
   api.bindArguments = bindArguments;
   api.patchMacroTask = patchMacroTask;
+  // load blackListEvents from global
+  const SYMBOL_BLACK_LISTED_EVENTS = Zone.__symbol__('BLACK_LISTED_EVENTS');
+  const SYMBOL_DISABLE_EVENTS = Zone.__symbol__('DISABLE_EVENTS');
```
IgorMinar: I'm not sure about this name.. I think it should be "DISABLED_EVENTS", or "UNPATCHED_EVENTS" or "UNINTEREPTED_EVENTS". 

because you'll use it declaratively rather than imperatively as in:

```
Zone.__symbol__('DISABLED_EVENTS') = ['scroll'];
```


"DISABLE_EVENTS" would be more suitable if the api looked like:

```
Zone.disableEvents("scroll", "mouseover", "somethingelse")
```

lib/browser/api-util.ts
```diff
@@ -13,4 +13,14 @@ Zone.__load_patch('util', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
   api.patchMethod = patchMethod;
   api.bindArguments = bindArguments;
   api.patchMacroTask = patchMacroTask;
+  // load blackListEvents from global
```
IgorMinar: you might want to clarify in the comment that the `BLACK_LISTED_EVENTS` name is deprecated, but still supported.

lib/browser/api-util.ts
JiaLiPassion: @IgorMinar, thanks, sure, I believe `DISABLED_EVENTS` is better, I just copied it from https://github.com/angular/angular/pull/28529, I will change it and also add comment in the `angular` PR.

lib/browser/api-util.ts
JiaLiPassion: Sure, got it!

JiaLiPassion:
@IgorMinar, I updated the env name to `__zone_symbol__DISABLED_EVENTS`, and also add comments, please review, thanks!

lib/browser/api-util.ts
kara: I think `DISABLED_EVENTS` is good, but `UNPATCHED_EVENTS` might be even better because it doesn't imply that things like scrolling just don't work.

JiaLiPassion: @kara, thanks, you are right, I will change it to `UNPATCHED_EVENTS`.

JiaLiPassion:
@IgorMinar, this one is also green now, and I changed the name based on @kara 's suggestion to `__zone__symbol__UNPATCHED_EVENTS`

lib/browser/api-util.ts
IgorMinar: Yes! Unlatched is the best

lib/browser/api-util.ts
IgorMinar: Can you update this plz?

JiaLiPassion: Sure, I just updated the comments, please review, thanks!

IgorMinar:
Thank you!

PR #1193, opened 2019-02-28
build: bump version to 0.9.0

JiaLiPassion:
1. Update package.json and CHANGELOG
2. because we have some new files under `dist`, I also `add dist` into this commit, if this not needed, I can remove the changes under `dist`.

@IgorMinar, @mhevery, please review, thanks!


IgorMinar:
We should either cut a beta first or create a PR in the cli repo that
verifies that this unreleased build will work well in a cli app. JiaLi
could you do that please? This might require differential loading in cli
(feature that we are currently working on) to be somewhat usable in CLI.

@manfredsteyer is leading that effort. Manfred could you please help JiaLi
verify that we've got all we need for differential loading before we cut
the 0.9 zone.js release? Thank you.

JiaLiPassion:
@IgorMinar, Sure, will do, thanks!

PR #1199, opened 2019-03-09
build: upgrade to typescript 3.2.2

JiaLiPassion:


lib/rxjs/rxjs.ts
```diff
@@ -145,26 +145,28 @@ import {Observable, Subscriber, Subscription} from 'rxjs';
     Subscriber.prototype.next = function() {
       const currentZone = Zone.current;
       const subscriptionZone = this._zone;
+      const args: [any?] = Array.prototype.slice.call(arguments) as [any?];
```
mhevery: This seems like extra work which we have to do. Why is the `slice` needed? Can't we just cast it somehow to make TS happy?

lib/rxjs/rxjs.ts
```diff
@@ -145,26 +145,28 @@ import {Observable, Subscriber, Subscription} from 'rxjs';
...
       } else {
-        return next.apply(this, arguments);
+        return next.apply(this, args);
       }
     };
 
     Subscriber.prototype.error = function() {
       const currentZone = Zone.current;
       const subscriptionZone = this._zone;
+      const args: [any?] = Array.prototype.slice.call(arguments) as [any?];
```
mhevery: same...

lib/rxjs/rxjs.ts
JiaLiPassion: Sure,  without making a new `slice`, just `cast arguments to any`.

lib/rxjs/rxjs.ts
JiaLiPassion: Got it.

PR #1201, opened 2019-03-15
build: make legacy standalone bundle

JiaLiPassion:
In `angular cli`, there is a file named `es2015-polyfill.js`, inside this file, all `polyfill` for `es5 legacy browsers` will be there, so with `zone evergreen bundle`, the most idle way to handle this is.

1. put a `zone-legacy.js` inside `es2015-polyfill.js`.
2. put `zone-evergreen.js` inside `polyfill.ts`.

So 
1. in `es5 legacy browser`, it will load `zone-legacy.js` + `zone-evergreen.js`.
2. in `modern browser`, it will only load `zone-evergreen.js`.

Before this PR, `zone-legacy.js` need to be loaded after `zone-evergreen.js`, because  the `zone-legacy` use a lot of `common logic` from `zone-evergreen.js`. So in this PR,

1. `zone-legacy.js` is a `lazy load module` which will only define a `factory function`.
2. When `zone-evergreen.js` loading, it will check whether this `lazy load module exist`, if exists, it will execute the `zone-legacy patch logic`.

### DIFF
--- a/lib/browser/browser-legacy.ts
+++ b/lib/browser/browser-legacy.ts
@@ -14,11 +14,16 @@ import {eventTargetLegacyPatch} from './event-target-legacy';
 import {propertyDescriptorLegacyPatch} from './property-descriptor-legacy';
 import {registerElementPatch} from './register-element';
 
-Zone.__load_patch('registerElement', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
-  registerElementPatch(global);
-});
+(function(_global: any) {
+_global['__zone_symbol__legacyPatch'] = function() {
+  const Zone = _global['Zone'];
+  Zone.__load_patch('registerElement', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
+    registerElementPatch(global, api);
+  });
 
-Zone.__load_patch('EventTargetLegacy', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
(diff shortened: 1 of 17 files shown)


lib/browser/browser-legacy.ts
```diff
@@ -14,11 +14,16 @@ import {eventTargetLegacyPatch} from './event-target-legacy';
 import {propertyDescriptorLegacyPatch} from './property-descriptor-legacy';
 import {registerElementPatch} from './register-element';
 
-Zone.__load_patch('registerElement', (global: any, Zone: ZoneType, api: _ZonePrivate) => {
-  registerElementPatch(global);
-});
+(function(_global: any) {
+_global['__zone_symbol__legacyPatch'] = function() {
```
mhevery: Why is it necessary to add `__zone_symbol__legacyPatch` to the global? Who invokes that function?

JiaLiPassion: Yeah, here is the flow.
1. loading `browser-legacy.js`, and register this factory function in global.
2. loading `zone-evergreen.js`, and call this `__zone_symbol__legacyPatch` here, https://github.com/angular/zone.js/pull/1201/files#diff-75082c639aefcc47b3b8df57c8870867R23, so the legacy patch can be loaded after the `common logic of zone` is loaded.

PR #1213, opened 2019-03-30
xhr: fix #1212, add XMLHttpRequest availability check to avoid error in service worker.

### LINKED ISSUE #1212: ServiseWorker compatiblitity issues in 0.9.0
The most recent update to 0.9.0 (in google3 cl/240639288 already rolled-back) broke one of the google3 apps. They have an "unusual" setup where they run Angular with zone.js in a service worker context. Zone.js 0.9.0 breaks in the serviceworker because property-descriptor-legacy.ts:56 directly references XmlHttpRequest[1], which is not available in service workers.

They use "self['__Zone_disable_XHR'] = true" to disable zone's other usages of XHR's and actually use the noop stub for Zone once bootstrapped:

```
 createPlatformFactory(platformCore, 'xxx_service_worker', [
    {provide: PLATFORM_ID, useValue: 'xxx_service_worker'},
  ])().bootstrapModuleFactory(XXXServiceWorkerModuleNgFactory, {ngZone: 'noop'});
```

Before re-landing 0.9.0 in google3, let's ensure that we avoid executing that code in service workers.

I'll also follow up with them about why they even load zone.js when then stub it out with `noop` during bootstrap, but we should still try to avoid breaking them.

[1] https://github.com/angular/zone.js/blob/31d31ceead4a8dfd0b747f1912f98f61c89dceff/lib/browser/property-descriptor-legacy.ts#L45

JiaLiPassion:
fix #1212.

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -94,6 +94,11 @@ Zone.__load_patch('XHR', (global: any, Zone: ZoneType) => {
   }
 
   function patchXHR(window: any) {
+    const XMLHttpRequest = window['XMLHttpRequest'];
+    if (!XMLHttpRequest) {
+      // XMLHttpRequest is not available in service worker
+      return;
+    }
     const XMLHttpRequestPrototype: any = XMLHttpRequest.prototype;
 
     function findPendingTask(target: any) {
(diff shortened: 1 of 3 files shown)


gkalpak:
Is Zone.js supposed to be running in Service Workers?

JiaLiPassion:
@gkalpak, `zone.js` support `WebWorker` from the very beginning, and it seems there is some application in `g3` use `zone.js` in `ServiceWorker` too.

gkalpak:
👍 (I would really be interested to know what they are doing with Zone.js in a ServiceWorker :smiley:)

JiaLiPassion:
@gkalpak, yeah, me too, it seems @IgorMinar will confirm the reason too.

PR #1214, opened 2019-04-03
Ensure that EventTarget is patched prior to legacy property descriptor patch

jwKimo:
If the event that EventTarget is defined on a device that also does not support patch via property descriptor, the result is that the XMLHttpRequest class would be patched prior to the event target being patched (due to returning in event-target-legacy if EventTarget is defined). This results in the wrapper that replaces XMLHttpRequest missing the patched listener fields. E.g. __zone_symbol__addEventListener and __zone_symbol__removeEventListener
event-target-legacy.

The result is that for any attempted http requests, a type error occurs in scheduleTask in browse.ts -> patchXhr, due to attempting call to non-existent __zone_symbol__addEventListener and  __zone_symbol__removeEventListener, causing the requests to fail.

Ensuring that patching of event target for legacy path takes place before legacy property descriptor patch eliminates this problem.

### DIFF
--- a/lib/browser/event-target-legacy.ts
+++ b/lib/browser/event-target-legacy.ts
@@ -101,7 +101,7 @@ export function eventTargetLegacyPatch(_global: any, api: _ZonePrivate) {
   // vh is validateHandler to check event handler
   // is valid or not(for security check)
   api.patchEventTarget(_global, apiTypes, {vh: checkIEAndCrossContext});
-  (Zone as any)[api.symbol('patchEventTarget')] = true;
+  (Zone as any)[api.symbol('patchEventTarget')] = !!_global[EVENT_TARGET];
   return true;
 }
 
--- a/lib/browser/event-target-legacy.ts
+++ b/lib/browser/event-target-legacy.ts
@@ -101,7 +101,7 @@ export function eventTargetLegacyPatch(_global: any, api: _ZonePrivate) {
   // vh is validateHandler to check event handler
   // is valid or not(for security check)
   api.patchEventTarget(_global, apiTypes, {vh: checkIEAndCrossContext});
-
+  (Zone as any)[api.symbol('patchEventTarget')] = true;
   return true;
 }
 
--- a/lib/browser/event-target-legacy.ts
+++ b/lib/browser/event-target-legacy.ts
@@ -24,7 +24,7 @@ export function eventTargetLegacyPatch(_global: any, api: _ZonePrivate) {
     // Workaround for: https://github.com/google/tracing-framework/issues/555
     apis = WTF_ISSUE_555_ARRAY.map((v) => 'HTML' + v + 'Element').concat(NO_EVENT_TARGET);
   } else if (_global[EVENT_TARGET]) {
-    // EventTarget is already patched in browser.ts
+    apis.push(EVENT_TARGET);
   } else {
     // Note: EventTarget is not available in all browsers,
     // if it's not available, we instead patch the APIs in the IDL that inherit from EventTarget
--- a/lib/browser/event-target.ts
+++ b/lib/browser/event-target.ts
@@ -7,6 +7,10 @@
  */
 
 export function eventTargetPatch(_global: any, api: _ZonePrivate) {
+  if ((Zone as any)[api.symbol('patchEventTarget')]) {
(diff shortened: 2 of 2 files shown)


jwKimo:
I signed it

JiaLiPassion:
@jwKimo, could you describe which `browser` will have this issue? And what is the error of this issue? Thanks

jwKimo:
This was discovered on internal Google application running on an LG WebOS 3.0 SmartTV device. The error is that XMLHttpRequests fail to process due to a type error occurring due to attempting to make an call on undefined for "__zone_symbol__addEventListener" at https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts#L166

This is due to patchClass of "XMLHttpRequest" (https://github.com/angular/zone.js/blob/master/lib/browser/property-descriptor-legacy.ts#L25) taking place before patching of event target(This was not the case in v0.8.21). Which does does take place in event-target-legacy as it returns if EventTarget is defined: https://github.com/angular/zone.js/blob/master/lib/browser/event-target-legacy.ts#L26

JiaLiPassion:
@jwKimo, thanks, so this error occurs before https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts#L78 is called, is that right?

jwKimo:
@JiaLiPassion No. The error occurs when attempting to create and send an XMLHttpRequest. The send triggers scheduleTask(https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts#L117) which will attempt to call undefined variable (https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts#L166)

JiaLiPassion:
@jwKimo, Sorry for the late reply, I know the error is because https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts#L166.
what I don't understand is I don't think this issue has relationship with   https://github.com/angular/zone.js/blob/master/lib/browser/property-descriptor-legacy.ts#L25

So the current situation is,

- `EventTarget` is available
- `canPatchViaPropertyDescriptor` return false.

And the load order is 
1. `browser-legacy.ts` -> `event-target-legacy.ts` (https://github.com/angular/zone.js/blob/master/lib/browser/event-target-legacy.ts#L26)
So `EventTarget` is not patched at this time.
2. `browser.ts` -> `event-target.ts` (https://github.com/angular/zone.js/blob/master/lib/browser/event-target.ts#L28), 
So `EventTarget` will be patched here.

And the error is `XMLHttpRequest.prototype.addEventListener` is not patched, so I am not sure the reason is https://github.com/angular/zone.js/blob/master/lib/browser/event-target.ts#L28. 

After your fix, the problem is gone, is that right? 

Sorry for the long text, I just want to make sure I understand the whole situation, Thank you!

jwKimo:
Ah yes should have elaborated on that part.
 1. The patching of event target yields the fields __zone_symbol__addEventListener and __zone_symbol__removeEventListener
 2.  It is expected that these fields exist on the XMLHttpRequest prototype when creating and sending requests: https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts#L166 
 3. property-descriptor-legacy makes a call to `patchClass('XMLHttpRequest')` which replaces the original XMLHttpRequest` with a wrapper with copied fields: https://github.com/angular/zone.js/blob/master/lib/browser/property-descriptor-legacy.ts#L25
 4. Thus if `patchClass('XMLHttpRequest')` runs before patching of event target, the patched event target fields mentioned in item 1 above, will be missing from the wrapper that replaces XMLHttpRequest.
5. Result is that XMLHttpRequests will all fail to send due to type error as it will attempt to retrieve and then call undefined function on XMLHttpRequest, "__zone_symbol__addEventListener": https://github.com/angular/zone.js/blob/master/lib/browser/browser.ts#L103

JiaLiPassion:
@jwKimo, thank you for the detail information, I understand the reason now, so the underlying reason is `patchClass` only copy the `fields`, not set the `prototype`. I think the current fix is ok, I will try to add the `prototype copy logic` later,  thanks!

jwKimo:
Thanks! I don't have write access so someone else would need to merge the pull request.

JiaLiPassion:
@jwKimo, yeah, I am asking @vikerman to check this PR, thanks!

PR #1218, opened 2019-04-07
feat(core): Add an option '__Zone_symbol_prefix' to set symbol prefix used in Zone.__symbol__().

mleibman:
Add a global environment option `__Zone_symbol_prefix` to allow configuring the symbol prefix Zone.js uses to store its state data on objects.

This is needed in situations where multiple instances of Zone.js may 'touch' the same object or DOM node, accessing and clobbering the other instance's data stored there, and causing unpredictable behavior. Ultimately, it would be nice if Zone.js used proper ES6 symbols or some unique identifier as a prefix, but many products rely on being able to access the data stored by it.

This change adds the env option and cleans up all hard-coded references to '__zone_symbol_XXX' to go through `Zone.__symbol__()`. In order to test the changes and protect against regressions, the tests are run with the symbol prefix changed to `__zone_symbol_test__`.

While the primary purpose of this change is to be able to isolate data stored on the objects & DOM nodes, some global environment options are also specified with the `__zone_symbol__` prefix and not the usual `__Zone_` one. The current API for providing env options isn't very consistent. For example, disabling 'on<X>' property patching is done via `__Zone_ignore_on_properties`, w/o using the symbol prefix, while disabling event patching is done via `__zone_symbol__UNPATCHED_EVENTS`, using the symbol prefix. This change only affects the options that are prefixed with `__zone_symbol__`.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -466,6 +466,8 @@ gulp.task('build', [
 ]);
 
 function nodeTest(specFiles, cb) {
+  require('./build/test/node-env-setup');
+
   // load zone-node here to let jasmine be able to use jasmine.clock().install()
   // without throw error
   require('./build/lib/node/rollup-main');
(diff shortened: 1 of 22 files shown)


mleibman:
Git commits got messed up, will create a new PR...

PR #1221, opened 2019-04-09
fix: patch jasmine clock only if `enableClockPatch` is set

vikerman:


JiaLiPassion:
this one has been fixed in other pr

PR #1225, opened 2019-04-18
fix: move property patch to legacy

### LINKED ISSUE #1224: Overwritten method Object.defineProperty breaking functionality of jsPDF
The rewriteDescriptor function being called by the zone.js Object.defineProperty override is breaking some functionality of the popular jsPDF library: https://github.com/MrRio/jsPDF/issues/2429

I don't really understand why the rewriteDescriptor method changes the desc.configurable to true but it breaks several places in the jsPDF code where jsPDF tries to define non-configurable properties like this:
```
var _BBox = [];
Object.defineProperty(this, 'BBox', {
   configurable: false,
   writeable: true,
   get: function () {
       return _BBox;
   },
   set: function (value) {
       _BBox = value;
   }
});
```
It could be fixed by freezing the desc object in jsPDF like this:
```
var _BBox = [];
Object.defineProperty(this, 'BBox', Object.freeze({
   configurable: false,
   writeable: true,
   get: function () {
       return _BBox;
   },
   set: function (value) {
       _BBox = value;
   }
}));
```
However, the jsPDF maintainers make a good point when they ask why they should change their code when it's adhering to standard functionality. I really can't answer that as I don't know enough about zone.js but I hope someone here can help us get this figured out.

JiaLiPassion:
Close #1224

Should not try to override `Object.defineProperty` in `evergreen browser`, because this property patch seems for some old browser (chrome mobile for customElement createdCallback).

### DIFF
--- a/file-size-limit.json
+++ b/file-size-limit.json
@@ -8,7 +8,7 @@
     {
       "path": "dist/zone.min.js",
       "checkTarget": true,
-      "limit": 44000
+      "limit": 44500
     }
   ]
 }
--- a/lib/browser/browser-legacy.ts
+++ b/lib/browser/browser-legacy.ts
@@ -10,6 +10,7 @@
  * @suppress {missingRequire}
  */
 
+import {propertyPatch} from './define-property';
 import {eventTargetLegacyPatch} from './event-target-legacy';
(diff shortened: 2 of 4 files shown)


yearnaudov:
Any idea when will this PR be merged ?

PR #1232, opened 2019-05-07
Zone bazel

JiaLiPassion:
Bazel rule for zone.js


WORKSPACE
```diff
@@ -0,0 +1,36 @@
...
+load("@build_bazel_rules_nodejs//:defs.bzl", "yarn_install")
+
+yarn_install(
+    name = "npm",
+    exclude_packages = [
+        # Don't need to run Bazel under Bazel
+        "@bazel/bazel",
+        "@bazel/ibazel",
+        "@bazel/buildifier",
+        # TODO: fails because it tries to install itself elsewhere on disk:
```
alexeagle: remove unneeded comments

dist/webapis-notification.min.js
```diff
@@ -1 +1,15 @@
...
\ No newline at end of file
+/**
+* @license
+* Copyright Google Inc. All Rights Reserved.
+*
+* Use of this source code is governed by an MIT-style license that can be
+* found in the LICENSE file at https://angular.io/license
+*/
+!function(t){"function"==typeof define&&define.amd?define(t):t()}(function(){"use strict";
+/**
```
alexeagle: seems like we should figure out how to avoid having the license duplicated? I suppose it compresses well...

lib/BUILD.bazel
```diff
@@ -0,0 +1,38 @@
+load("@npm_bazel_typescript//:defs.bzl", "ts_library")
+package(default_visibility = ["//:__pkg__"])
+
+exports_files([
+    "zone.ts",
```
alexeagle: exposing sources outside the package is a smell, usually any rules using them should be declared here too. At least add a comment explaining why another package needs to be able to use these sources?
(the `.ts` files in particular seem wrong here, they should be compiled by a rule in this package)

rollup.bzl
```diff
@@ -0,0 +1,12 @@
+load("//lib:rollup.bzl", "rollup_bundle")
+def rollup(bundles):
```
alexeagle: I don't think the extra indirection to this file is useful, just inline this into the BUILD.bazel ?

test/BUILD.bazel
```diff
@@ -0,0 +1,50 @@
...
+ts_library(
+  name = "test_lib",
+  deps = [
+    "//lib",
+    "@npm//@types",
+    "@npm//rxjs",
+  ],
+  srcs = glob([
+    "**/*.spec.ts",
+    "browser_entry_point.ts",
```
alexeagle: Once we move this to angular/angular, buildifier will warn that entries in a glob with no wildcard should be outside the glob
`glob(["**/*.spec.ts"]) + ["browser-zone-setup.ts"]`

test/test.bzl
```diff
@@ -0,0 +1,24 @@
...
+
+test_deps = [
+    "test/wtf_mock",
+    "test/test_fake_polyfill",
+    "lib/zone",
+    "lib/common/promise",
+    "test/main",
+]
+
+def get_web_test_deps():
```
alexeagle: this isn't called anywhere yet?

We generally prefer to just expand this out into the BUILD file. It's a bit repetitive to have 5 ts_library targets with similar setup, however tooling like ts_auto_deps makes it manageable. By having this in a macro, tooling like buildozer will no longer be able to machine-edit the rules

BUILD.bazel
```diff
@@ -0,0 +1,151 @@
...
+]
+
+genrule(
+    name = "zone_externs",
+    srcs = ["//lib:zone_externs"],
+    outs = ["dist/zone_externs.js"],
+    cmd = "cp $< $(@D)/zone_externs.js",
+)
+
+npm_package(
```
alexeagle: we should add an e2e test for the structure and content of the package so that the `genrule` spaghetti above can be safely refactored out.
something like https://github.com/angular/angular/blob/master/packages/bazel/test/ng_package/core_package.spec.ts maybe?

BUILD.bazel
```diff
@@ -0,0 +1,151 @@
...
+[
+    filegroup(
+        name = b + ".umd.min",
+        srcs = [":" + b],
+        output_group = "umd_min",
+    )
+    for b in ES2015_BUNDLES
+]
+
+# Extract and rename each es5 bundle to a .js and .min.js in the dist/ dir
```
alexeagle: this seems wrong. If we want the artifacts to appear in the `dist/` subfolder in the output tree (and therefore in the npm package) then we should probably have a BUILD file in `dist/` in the source tree which declares them. then we don't have to move things around. I can do some refactoring to help with that.

WORKSPACE
JiaLiPassion: done, thanks!

dist/webapis-notification.min.js
JiaLiPassion: The current `zone.js bundles` have the same issues, because inside the source code, we also have the `license header`, so I think we can resolve this issue later.

lib/BUILD.bazel
JiaLiPassion: This for test, I have removed it and will re-design when we figure out how to make `karma` for zone.js work.

rollup.bzl
JiaLiPassion: Sure, you are right, I have removed all separated `rollup.bzl` and move the definition into `BUILD.bazel`.

test/BUILD.bazel
JiaLiPassion: Got it, I will modify this later, because currently this settings not work...

test/test.bzl
JiaLiPassion: Got it, I will modify this later, because currently this settings not work... I will delete all `bazel` settings of test for now. So this PR is only about `npm_package`.

BUILD.bazel
JiaLiPassion: Got it, I made it work, and testing it now.

BUILD.bazel
JiaLiPassion: Sure, got it, I have added test cases.

lib/BUILD.bazel
```diff
@@ -0,0 +1,26 @@
+load("@npm_bazel_typescript//:defs.bzl", "ts_library")
+
+package(default_visibility = ["//:__pkg__"])
+
+exports_files([
+    "closure/zone_externs.js",
+])
+
+filegroup(
```
alexeagle: I don't think we need both the `exports_files` and the `filegroup` to expose the zone_externs.js file

dist/zone.js.d.ts
```diff
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+/// <amd-module name="__main__/lib/zone" />
```
alexeagle: this is a bit gross, users don't want to see __main__ - in angular we do something to trim these lines from published .d.ts

BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+    name = "npm_package",
+    srcs = [
+        "CHANGELOG.md",
+        "LICENSE",
+        "LICENSE.wrapped",
+        "README.md",
+        "package.json",
+    ],
+    replacements = {
+        # Since the bazel package is experimental, patch the version number
```
alexeagle: this seems like copy-paste - this isn't the bazel package

dist/BUILD.bazel
```diff
@@ -0,0 +1,121 @@
...
+genrule(
+    name = "zone_externs",
+    srcs = ["//lib:closure/zone_externs.js"],
+    outs = ["zone_externs.js"],
+    cmd = "cp $< $@",
+)
+
+genrule(
+    name = "zone_d_ts",
+    srcs = ["//lib"],
```
alexeagle: is this concatenating multiple .d.ts inputs into a single output? maybe we should use the api-extractor like we do in the angular repo.

dist/BUILD.bazel
```diff
@@ -0,0 +1,121 @@
...
+        name = b + ".umd.min",
+        srcs = [":" + b + "-rollup"],
+        output_group = "umd_min",
+    )
+    for b in ES2015_BUNDLES
+]
+
+# Extract and rename each es5 bundle to a .js and .min.js in the dist/ dir
+[
+    genrule(
```
alexeagle: this is pretty gross... maybe I can find a way to refactor it later

BUILD.bazel
```diff
@@ -0,0 +1,49 @@
...
+        "@npm//@types",
+    ],
+)
+
+jasmine_node_test(
+    name = "test_npm_package",
+    srcs = [":npm_package_spec_lib"],
+    data = [
+        ":npm_package",
+        "@npm//@types/shelljs",
```
alexeagle: I don't think the typings are needed by the test, these would only be useful for a rule that does type-checking

dist/zone.js.d.ts
JiaLiPassion: Got it, could you show me the source that how `angular` trim those lines? Thanks

BUILD.bazel
JiaLiPassion: You are right, I removed this `replacements`.

dist/BUILD.bazel
JiaLiPassion: This `rule` just copy the `zone.d.ts` from `lib` to `dist/zone.js.d.ts`, we only have one `d.ts` file here.

dist/BUILD.bazel
JiaLiPassion: You are right, thanks!

BUILD.bazel
JiaLiPassion: Got it, `typings` removed. thanks!

JiaLiPassion:
https://github.com/angular/angular/pull/30615 passed, so this PR should work fine at least in `angular` repo. Need to very it in `G3` too.

PR #1239, opened 2019-06-07
build: add __load_patch and __symbol__ in zone_extern for closure

JiaLiPassion:
add `__load_patch` and `__symbol__` in `zone_extern.js` for `closure` compiler. 
In PR https://github.com/angular/angular/pull/30533, we use `Zone.__symbol__` function to get the native delegate, so we need those API to be reserved by closure compiler.

### DIFF
--- a/dist/zone_externs.js
+++ b/dist/zone_externs.js
@@ -1,10 +1,10 @@
 /**
-* @license
-* Copyright Google Inc. All Rights Reserved.
-*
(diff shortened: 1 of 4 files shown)


JiaLiPassion:
@mhevery, please review, after this is merged, we need to cut a new release to make https://github.com/angular/angular/pull/30533 work! Thanks!

Issue #1227, opened 2019-04-19
zone-patch-rxjs - Unsubscription error - Cannot read property 'apply' of null


AleG94:
I'm trying to implement a retry http interceptor for Angular 7. 

Without `import 'zone.js/dist/zone-patch-rxjs';` it works fine but when I add it, I get:

```
ERROR {"message":"1 errors occurred during unsubscription:\n1) TypeError: Cannot read property 'apply' of null","name":"UnsubscriptionError","errors":[{}]}
```

This is an example repo to reproduce the issue: 

[https://stackblitz.com/edit/angular-s3iv5u?file=polyfills.ts](url)

Comment and uncomment the line `import 'zone.js/dist/zone-patch-rxjs';` in `polyfills.ts` to see the error appear and disappear.

tkem:
I can confirm this occurs when using `retryWhen` with `zone-patch-rxjs`.

PR #1248, opened 2019-07-09
fix(zone-patch-rxjs): make retryWhen work

### LINKED ISSUE #1194: Patch-rxjs: _unsubscribe function can be  can be null
here https://github.com/angular/zone.js/blob/9c65d25f583548d8366d94a633a8fceaecc6d1ca/lib/rxjs/rxjs.ts#L120
_unsbscribe is applied before null check
for retryWhen Operator it is set to be null for brief time see here 

https://github.com/ReactiveX/rxjs/blob/d229235f4812dd5140209300e7cddf55553d5f53/src/internal/operators/retryWhen.ts#L108

NicBright:
This PR is intended to fix #1194  and #1227 

I've not included a test yet because I want to get feedback first whether this change looks good or not. @JiaLiPassion what do you think?

### DIFF
--- a/lib/rxjs/rxjs.ts
+++ b/lib/rxjs/rxjs.ts
@@ -113,12 +113,12 @@ import {Observable, Subscriber, Subscription} from 'rxjs';
         },
         set: function(this: Subscription, unsubscribe: any) {
           (this as any)._zone = Zone.current;
-          (this as any)._zoneUnsubscribe = function() {
+          (this as any)._zoneUnsubscribe = unsubscribe ? function() {
             if (this._zone && this._zone !== Zone.current) {
               return this._zone.run(unsubscribe, this, arguments);
             }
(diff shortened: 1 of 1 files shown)


NicBright:
@JiaLiPassion: the tests failed because of a timeout ... could you maybe restart the test run? Or shall I push another dummy commit?

JiaLiPassion:
@NicBright , thanks for the PR, I will restart CI, and could you explain why `unsubscribe` will be null with a simple reproduce repo? And if possible, please add a test case in `zone.js` also, thanks!.

JiaLiPassion:
@NicBright , could you explain why the `unsubscribe` will be null and also create a new PR in `angular repo`, because now the `zone.js repo` will be archived.

alexeagle:
sorry for the churn @NicBright - we hope you'll follow us to the Angular monorepo and re-submit the PR.

